PR #11, opened 2026-05-12
Noise reduction: sox noisered + tighter compand gate

### LINKED ISSUE #10: alpha.6: noise reduction in capture preprocessing (ES8388 ADC hiss + room noise during speech)
## Symptom

On hardware testing of PR #9 (sox compand for quiet-speech lift), the operator reports that even after compand and the band-pass filters there is broadband Gaussian-like noise mixed in with the actual speech in `/tmp/geniepod-last-rec.wav`. The noise is audible during playback alongside the user's voice.

Sources contributing to the noise are at least:
1. **ES8388 ADC self-noise** — measurable across the speech band even with the mic input silent.
2. **Room ambient noise** — HVAC, computer fans, ambient electronics.
3. **Acoustic bleed from USB headphone** — TTS output leaking back into the LyraT mic (separate issue if it becomes severe; not all noise is bleed).
4. **Mains-frequency hum and harmonics** — usually killed by `highpass 100`, but residual harmonics around 120/180 Hz can survive.

Compand alone can compress + lift dynamic range; it cannot subtract spectrally-correlated noise from the signal.

### A. Tighten the compand floor (zero new code, 4-number change)

Current curve: `compand 0.02,0.20 -50,-50,-25,-12,-5,-5 -2`
Proposed:     `compand 0.02,0.20 -65,-90,-50,-50,-25,-12,-5,-5 -2`

Adds an aggressive gate below -65 dBFS (push to -90 dBFS = effectively silence). Helps the *silent-between-words* case. Does NOT help noise that is masked by speech.

### B. sox `noisered` with a noise profile (modest scope, needs profile)

Generate a noise profile once at install time from a silent LyraT capture, store at `/opt/geniepod/data/mic-noise.prof`, then sox uses it for spectral subtraction on every capture.

Sox chain becomes:
```
channels 1 -> highpass 100 -> lowpass 7000 -> noisered <profile> 0.21 -> compand ... -> gain -n -3
```

Noise profile is host-specific (mic + room). Generated by:
```bash
arecord -D plughw:APE,0 -c 2 -r 24000 -f S16_LE -d 3 /tmp/silence.wav
sox /tmp/silence.wav -n channels 1 highpass 100 lowpass 7000 noiseprof /opt/geniepod/data/mic-noise.prof
```

`setup-jetson.sh` would prompt the operator to be silent for 3 seconds during a fresh install.

### C. RNNoise integration (high quality, more invasive)

[RNNoise](https://github.com/xiph/rnnoise) is a small RNN-based noise suppressor (~85 KB model) used in many real-time voice applications. C library, real-time on a CPU thread.

Best perceptual quality; requires building/linking a new dependency. Probably 200-300 LoC of glue + a build-time integration. Significantly better than spectral subtraction but a real engineering lift.

### D. WebRTC NS via libwebrtc-audio-processing

Comparable to RNNoise; widely deployed. Similar integration cost.

## Recommendation

Land A first (it's free, ~5 minutes), then **B** as the primary noise-reduction strategy for alpha.6, with `setup-jetson.sh` capturing the noise profile at install time. C/D are alpha.7+ if B is insufficient.

## Acceptance criteria

- [ ] `aplay /tmp/geniepod-last-rec.wav` after a typical capture sounds noticeably less noisy than alpha.5/PR #9 baseline.
- [ ] No clearly-audible artifacts on speech (musical noise, swirling, dropouts). Spectral subtraction can introduce these at high aggressiveness; tune the `noisered` amount conservatively (0.15-0.25).
- [ ] STT accuracy improves on quiet/marginal captures (the same kind that compand helped with), measurable against a small fixed test set.
- [ ] If the noise profile is missing, the pipeline falls back to the existing compand chain (no fatal failure).

## Related

- #6 — compand (the prior preprocessing PR; landed quiet-speech lift)
- #5 — whisper-medium fallback (parallel track for marginal-audio accuracy)

ai-hpc:
## Summary

Capture-side noise reduction in three pieces:

1. **`record_audio` sox chain gains a conditional `noisered <profile> 0.21`** step (spectral noise subtraction) when `/opt/geniepod/data/mic-noise.prof` exists.
2. **`setup-jetson.sh [5f/6]`** captures 3 seconds of ambient silence at install time, runs sox `noiseprof` against it, and stores the per-host noise profile at the canonical path.
3. **compand transfer function** tightened with an aggressive gate below -65 dBFS to clean up the silent-between-words noise band.

Closes #10.

## Why

Hardware testing of PR #9 (compand) confirmed compand works for the *dynamic-range* problem (quiet speech lifted) but the operator could still hear broadband Gaussian-like noise mixed *in with* their voice in `/tmp/geniepod-last-rec.wav` playback. That's not a dynamic-range issue — it's spectral noise (ES8388 ADC self-noise + room ambient + electronics) that compand and band-pass alone cannot remove.

The fix uses two complementary mechanisms:

| Noise class | Mechanism |
|---|---|
| Silent-between-words noise | Tighter compand gate: input -65 dBFS maps to output -90 dBFS (effectively silenced). Existing quiet-speech lift at -25/-12 untouched. |
| Noise mixed with speech | `sox noisered` spectral subtraction using a host-specific noise profile generated at install time. |

## Sox chain — before vs after

Before (PR #9):
```
channels 1 -> highpass 100 -> lowpass 7000 -> compand (lift only) -> gain -n -3
```

After (this PR):
```
channels 1 -> highpass 100 -> lowpass 7000
  -> noisered <profile> 0.21         [if profile exists; skipped otherwise]
  -> compand (gate + lift)
  -> gain -n -3
```

## The noise profile

Host-specific (mic + room + ambient electronics) so cannot be shipped in the repo. `setup-jetson.sh` generates it at install time via:

```bash
arecord -D plughw:APE,0 -c 2 -r 24000 -f S16_LE -d 3 /tmp/silence.wav
sox /tmp/silence.wav -n channels 1 highpass 100 lowpass 7000 \
    noiseprof /opt/geniepod/data/mic-noise.prof
```

Operator is told to stay silent for the 3-second capture. If they speak during capture, the profile gets contaminated with speech and subsequent noisered will reduce that speech band — fix is to delete the profile and re-run setup.

If sox is missing or arecord fails, the step prints a SKIP/WARN and continues. Voice pipeline falls back cleanly to compand-only.

## `noisered` amount: 0.21

Sox `noisered` takes an amount in `[0.0, 1.0]`. Higher = more aggressive subtraction. Aggressive values (>0.4) produce the famous "musical noise" artifact — swirling, watery, robotic remnants of the speech band. 0.15-0.25 is the conservative range where the noise reduction is audible but the artifacts aren't. 0.21 was picked as a starting point; tune per-host if needed.

## Compand transfer — before vs after

Before: `compand 0.02,0.20 -50,-50,-25,-12,-5,-5 -2`

After:  `compand 0.02,0.20 -65,-90,-50,-50,-25,-12,-5,-5 -2`

New first knee adds a gate. Everything else unchanged. Speech is unaffected because its dynamic range sits above -50 dBFS.

## Test plan

- [ ] `make deploy` from build VM
- [ ] On Jetson: re-run `setup-jetson.sh`, watch for the new `[5f/6] Mic noise profile for sox noisered...` section and the `>>> STAY SILENT <<<` prompt. Be silent for 3 s, confirm `mic-noise.prof` lands in `/opt/geniepod/data/`.
- [ ] Voice loop log line now shows `noisered = true` (assuming profile was captured): `INFO preprocessed audio ... noisered = true`
- [ ] `aplay /tmp/geniepod-last-rec.wav` after a typical voice cycle — should be noticeably less hissy than alpha.5-baseline playback.
- [ ] No "swirling" / "watery" / "underwater" artifacts on speech (if there are, lower the 0.21 amount).
- [ ] STT accuracy on quiet/mumbled prompts improves vs PR #9-alone baseline.

## Acceptance criteria (from #10)

- [ ] `aplay /tmp/geniepod-last-rec.wav` sounds less noisy than alpha.5/PR #9 baseline
- [ ] No clearly-audible artifacts on speech (musical noise, swirling, dropouts)
- [ ] STT accuracy improves on quiet/marginal captures
- [x] Missing profile is a no-op (sox chain falls back to compand-only)

## Notes

- Re-run `setup-jetson.sh` after moving the device to a new room. Profiles are spectrally local.
- If the operator can't stay silent during install (e.g. noisy office), delete `/opt/geniepod/data/mic-noise.prof` and rerun setup from a quieter moment; sox falls back gracefully if the profile is absent.
- This is purely capture-side. TTS bleed into the mic via the USB headphone (separate issue if persistent) is not addressed here — that would need AEC or hardware isolation.


ai-hpc:
Closing in favor of the alpha.7 DeepFilterNet direction (#12, PR #13).

DFN denoise was verified subjectively cleaner than sox `noisered` on-device, and handles non-stationary noise (typing, fan cycling, background voices) that `noisered` cannot follow even with a captured profile. The noise-profile capture infrastructure added here (`setup-jetson.sh` [5f/6]) is also not needed by DFN, which uses no profile.

PR #13 is currently closed too, pending #14 (second-cycle STT regression). The plan is to fix #14 first, then re-open / open a fresh alpha.7 DFN PR.

Branch `fix/issue-10-noise-reduction` left intact in case we ever need a non-DFN fallback path (e.g. if DFN turns out to be the cause of #14).

Issue #33, opened 2026-05-15
Default to genie-ai-runtime (v0.1.0-alpha.8) once #32 lands


ai-hpc:
## Background

`genie-ai-runtime` is explicitly the long-term replacement for `llama-server` in this stack:

- README.md L85 — *"`genie-ai-runtime`: Jetson-only C++ LLM runtime customized from `llama.cpp`"*
- ARCHITECTURE.md L70 — *"llama.cpp fork, CUDA kernels, model memory planner: `genie-ai-runtime`"*
- ARCHITECTURE.md L81 — *"`llama.cpp` OpenAI-compatible client → `genie-ai-runtime` client"*

The runtime already ships a drop-in `jetson-llm-server` binary with the same `/v1/chat/completions` shape that `LlmClient` (`crates/genie-core/src/llm/client.rs:6`) speaks. Per the runtime's own README, it is positioned as *"drop-in replacement target for `llama-server`"* with API shape *"closely enough"* to allow GenieClaw to swap backends via configuration.

Performance — reported on identical Jetson Orin Nano Super hardware:

| Metric | llama.cpp | genie-ai-runtime alpha.8 |
|---|---|---|
| Prefill | 17.97 tok/s | **38.68 tok/s** (+115%) |
| Time-to-first-token | baseline | **−61% vs runtime's own alpha.2** |

This is the parity-flip ticket.

## Dependency

**Blocked by #32** (LLM backend abstraction). When #32 lands, default backend selection becomes a single TOML key, and this issue is the one that flips that key.

## Concrete proposal

When #32 lands:

1. Change the default value of `[services.llm].backend` from `\"llama_cpp\"` to `\"genie_ai_runtime\"` in `deploy/config/geniepod.toml`.
2. Pin the runtime version: target release `v0.1.0-alpha.8` (or latest at flip time).
3. Default model: `Qwen3-4B-Q4_K_M.gguf` (the runtime's validated model).
4. Hardware target unchanged — Jetson Orin Nano Super 8 GB. The runtime explicitly does not support x86, discrete GPUs, Windows, or macOS. Dev-machine builds must continue to default to `llama_cpp`.

## Parity gate — verify, don't trust

The runtime's drop-in claim is *"closely enough"*. We should validate before flipping default. Acceptance gate:

- [ ] **API parity** — identical `chat` and `chat_stream` responses for a fixed prompt set (10 prompts × 3 temperatures), comparing token-by-token streaming and final string equality.
[...]

Issue #5, opened 2026-05-12
alpha.6: whisper-medium fallback for marginal-audio captures


ai-hpc:
## Context

Alpha.5 ships with `ggml-small.bin` (487 MB) as the default whisper model. With server mode keeping it GPU-resident, per-call STT latency is ~270 ms on clean audio.

When the input is **clean and loud**, accuracy is excellent. When the input is **quiet, distorted, or has background noise** (a common condition with LyraT onboard mics at room distance), whisper-small has a known tendency to fall back on hallucinated assistant-stock phrases ("I'm here to help", "feel free to ask", "thank you very much"), or to mis-hear specific words ("testing" → "tasting", "room" → "loom") because those alternatives have high decoder prior in its training data.

Even with the alpha.5 fixes for level (`gain -n -3`), spectrum (`highpass 100 lowpass 7000`), language (English-only decoder), channel handling (stereo → sox downmix to mono), and alpha.7's DFN denoise + half-duplex gate, whisper-small still occasionally hallucinates or mis-hears on marginal input.

## Strategy (revised — context-aware confidence escalation)

Run whisper-small as primary, then **gate the transcript through a context-aware confidence check before accepting it**. Only escalate to whisper-medium when the check says the transcript is suspect. This is cheaper on the common case (clean audio, small does fine) than always running medium, and more reliable than a static hallucination-phrase blocklist.

### Pipeline

```
arecord → DFN → sox → whisper-small (verbose_json) → confidence check ─┬─ pass → transcript
                                                                       └─ fail → whisper-medium (verbose_json) → transcript
```

### Confidence check (alpha version — "most recent context only")

Three signals, any one of which marks the transcript as low-confidence:

1. **Whisper-native scores.** Request `response_format=verbose_json` from whisper-server so the response includes per-segment `no_speech_prob` and `avg_logprob`. Low-confidence threshold:
   - `no_speech_prob > 0.6` (whisper itself isn't sure there was speech), OR
   - `avg_logprob < -1.0` (decoder ran on low-entropy ambiguous audio), OR
   - `compression_ratio > 2.4` (likely repetition / babble).

2. **Hallucination-phrase blocklist.** The existing `clean_hallucinations` ghost-phrase list (`"thank you"`, `"goodbye"`, `"thanks for watching"`, etc.) — if whisper-small's output reduces to one of these AND the audio had non-trivial energy (`peak dBFS > -40`), assume hallucination.

[...]

Issue #3, opened 2026-05-12
alpha.6: automate genie-llm.service model warm-up


ai-hpc:
## Problem

First `llama-server` inference after `systemctl start genie-llm.service` takes 30-60 s on Orin Nano while Phi-4-mini-Q4_K_M (2.4 GB) loads into iGPU memory. The first user-visible voice cycle either:
- Times out the LLM with `503: Loading model`, or
- Blocks for the full load time before responding.

Subsequent cycles are fast (~1 s LLM round-trip).

## Workaround (alpha.5)

A one-shot curl after starting `genie-llm.service` forces model load:

```bash
sleep 12
curl -s -X POST http://127.0.0.1:8080/completion \
  -H "Content-Type: application/json" \
  -d "{\"prompt\":\"hi\",\"n_predict\":4}" > /dev/null
```

This is documented in conversation transcripts and `doc/lyrat-jetson-audio.md` but is not automated. Operators have to remember to do it manually after each reboot/restart of the LLM service.

## Proposal

Add a systemd one-shot unit `genie-llm-warmup.service`:

```ini
[Unit]
Description=GeniePod LLM Warmup (force model into GPU memory)
After=genie-llm.service
Wants=genie-llm.service
Requires=genie-llm.service

[Service]
Type=oneshot
RemainAfterExit=yes
ExecStart=/bin/sh -c 'for i in $(seq 1 60); do curl -sf http://127.0.0.1:8080/health > /dev/null && break; sleep 1; done; curl -sf -X POST http://127.0.0.1:8080/completion -H "Content-Type: application/json" -d "{\"prompt\":\"hi\",\"n_predict\":1}" > /dev/null || true'
TimeoutStartSec=120

[Install]
WantedBy=geniepod.target
```

`setup-jetson.sh` enable loop should include `genie-llm-warmup` so a fresh `make deploy` + reboot ends with the LLM already hot before the first user interaction.

## Acceptance criteria

- [ ] New unit lives in `deploy/systemd/genie-llm-warmup.service`.
- [ ] `setup-jetson.sh` enables it.
- [ ] Tracked by `geniepod.target` so it's pulled in on every boot.
- [ ] First voice cycle after a cold boot completes within 12 s (vs ~60+ s today).
- [ ] CHANGELOG entry under alpha.6.

Issue #2, opened 2026-05-12
alpha.6: STT latency degrades from ~270 ms to ~3.6 s after LLM activity (iGPU contention)


ai-hpc:
## Symptom

On Jetson Orin Nano (8 GB iGPU), STT latency is fast and consistent before any LLM activity but degrades dramatically after `llama-server` (Phi-4-mini-Q4_K_M) generates a response.

Observed across three consecutive push-to-talk cycles on alpha.5:

| Cycle | STT latency |
|---|---:|
| 1 (cold, no prior LLM activity) | **270 ms** |
| 2 (after cycle 1's tool call + short TTS) | 379 ms |
| 3 (after cycle 2's verbose 3-sentence LLM response) | **3645 ms** |

## Hypothesized root cause

Both services share the Orin's 7.6 GB iGPU memory:
- `whisper-server` with ggml-small.bin: ~487 MB
- `llama-server` with Phi-4-mini-Q4_K_M: ~2.4 GB (model) + KV cache growth (~hundreds of MB per response)

When llama-server's KV cache grows during generation, it likely evicts whisper-server's model pages from active GPU memory. Next whisper inference has to page back in → 13× slowdown.

## Mitigation options to evaluate

- [ ] Pin whisper-server's GPU memory (force resident, no eviction).
- [ ] Force whisper to CPU (consistent ~3 s baseline, no GPU contention; trade variance for predictability).
- [ ] Add a small serializer in `genie-core` that pauses LLM streaming during STT calls (per-cycle, brief).
- [ ] Switch whisper to ggml-base.bin (~142 MB, smaller GPU footprint, less likely to be evicted).
- [ ] Run whisper-server on a higher-priority CUDA stream.
- [ ] Investigate Tegra-specific options like `cudaMallocManaged` vs explicit residency hints.

## Reproducer

```bash
ssh -t aihpc@<jetson> 'sudo -E RUST_LOG=info /opt/geniepod/bin/genie-core --config /etc/geniepod/geniepod.toml'
# Then in voice loop:
# Cycle 1: "what time is it"    -> note STT latency (~270 ms expected)
# Cycle 2: "tell me a story"    -> long LLM response
# Cycle 3: "what time is it"    -> STT latency now >1 s
```

## Acceptance criteria

- [ ] STT latency p95 within 2× of cold-cycle p50 for at least 5 consecutive cycles.
- [ ] No regression on cold-cycle performance.
[...]

PR #35, opened 2026-05-15
llm: introduce backend facade

### LINKED ISSUE #32: Make llama.cpp an optional LLM backend (introduce LlmBackend abstraction)
## Background

GenieClaw's LLM client is currently hard-wired to `llama-server`'s OpenAI-compatible API at `/v1/chat/completions`. See `crates/genie-core/src/llm/client.rs:6`:

```rust
/// Client for llama.cpp `--server` OpenAI-compatible API.
```

The repo already calls this out as a transitional adapter that should be peeled off when `genie-ai-runtime` is split out:

- README.md L68 — *"transitional `llama.cpp` support while `genie-ai-runtime` is not yet split out"*
- ARCHITECTURE.md L81 — *"`llama.cpp` OpenAI-compatible client → `genie-ai-runtime` client. Keep the client boundary narrow and model/runtime assumptions explicit."*

This issue promotes that boundary from "soft TODO" to a real backend abstraction, with `llama.cpp` becoming **one option among several** rather than the only path. The companion issue (flipping the default to `genie-ai-runtime`) depends on this one landing first.

## Proposed design

Introduce an `LlmClient` trait (or enum) in `crates/genie-core/src/llm/` that the rest of the codebase consumes:

```rust
#[async_trait]
pub trait LlmClient: Send + Sync {
    async fn health(&self) -> bool;
    async fn chat(&self, req: ChatRequest) -> Result<ChatResponse>;
    async fn chat_stream(
        &self,
        req: ChatRequest,
        on_token: impl FnMut(&str),
    ) -> Result<String>;
}

pub enum LlmBackend {
    LlamaCpp(LlamaCppClient),          // current llama-server adapter
    GenieAiRuntime(GenieRuntimeClient), // companion issue
    // future: Triton, vLLM-aarch64, etc.
}
```

The existing `LlmClient` struct in `client.rs:11` gets renamed to `LlamaCppClient` and moved to `crates/genie-core/src/llm/llama_cpp.rs`. The rest of the code consumes the trait, not a concrete type.

## Configuration shape

In `geniepod.toml`:

```toml
[services.llm]
backend = "llama_cpp"   # or "genie_ai_runtime"; default stays llama_cpp here
                        # (the default flip is the companion issue)
url     = "http://127.0.0.1:8080/v1/chat/completions"

# Backend-specific knobs go under a backend-named subtable:
[services.llm.llama_cpp]
systemd_unit = "genie-llm.service"

[services.llm.genie_ai_runtime]
systemd_unit = "genie-ai-runtime.service"
```

For backwards compatibility, when `backend` is absent we infer `llama_cpp` from the legacy keys (`services.llm.url`, `services.llm.systemd_unit`) for one release cycle.

## Files touched

- `crates/genie-core/src/llm/client.rs` — split into `mod.rs` (trait) + `llama_cpp.rs` (current impl)
- `crates/genie-core/src/llm/retry.rs` — generic over the trait
- `crates/genie-common/src/config.rs:148` — add `backend` enum field on `[services.llm]`
- `crates/genie-core/src/main.rs` — backend selection at startup
- `crates/genie-core/src/voice_loop.rs:152` — `llm.health()` already trait-shaped
- `crates/genie-core/src/lib.rs` — re-exports
- `deploy/config/geniepod.toml` and `geniepod.dev.toml` — add commented-out `backend = "..."` examples
- `deploy/systemd/genie-llm.service` — unchanged, but now optional in the deploy story

## Feature flags

Cargo features gate which backend code is compiled:

```toml
[features]
default      = ["llama-cpp"]
llama-cpp    = []
genie-ai-runtime = []
```

`make jetson` enables both; `make release` for dev machines defaults to `llama-cpp` only.

## Acceptance criteria

- [ ] `LlmClient` is a trait; `client.rs:6` doc comment no longer says "Client for llama.cpp"
- [ ] `geniepod.toml` accepts `[services.llm].backend` and the right client is constructed
- [ ] `genie-ctl`, `genie-health`, `genie-governor` continue to work against both backends
- [ ] CI matrix exercises both backends (mock for the runtime if cross-compile in CI is impractical)
- [ ] Legacy config keys still work for one release cycle (deprecation warning logged)
- [ ] ARCHITECTURE.md L81 entry resolved
- [ ] README.md L68 wording updated from "transitional" to "default with `genie-ai-runtime` opt-in"

## Why this is one of the cheapest moves we can make

The boundary is already narrow — `LlmClient` is constructed once in `main.rs` and threaded through. The work is mostly trait extraction + config plumbing, no behavioral change. Risk is low; payoff is that the companion issue (defaulting to `genie-ai-runtime`) becomes a one-line config change.

## Related

- Companion issue: defaulting `genie-ai-runtime` once parity is confirmed.
- Upstream: https://github.com/GeniePod/genie-ai-runtime

carlos4s:
## Summary
- First slice for #32, intentionally limited to `crates/genie-core/src/llm/`.
- Preserves the public `LlmClient` API and keeps `llama.cpp` as the default path.
- Extracts the raw OpenAI-compatible HTTP/SSE client behind a backend facade.
- Adds `llama_cpp` and `genie_ai_runtime` backend adapters without wiring config yet.
- Widens the `chat_stream` stream error object with `Send` so backend implementations satisfy the async trait boundary.

## Split plan
This PR is a preparatory slice and does not close #32 by itself. I opened it from `carlos4s:issue-32-llm-backend-abstraction` so the maintainer can review/merge the LLM-module boundary first.

Follow-up PRs can layer on:
- `services.llm.backend` config parsing and legacy default behavior.
- Startup/governor/service-unit plumbing for backend selection.
- Docs and CI/backend matrix updates.

## Tests
- `cargo check -p genie-core`
- `cargo test -p genie-core llm::`
- `cargo test -p genie-core`

Note: one unrelated memory-promotion test failed on the first full run, then passed when rerun directly and the second full `cargo test -p genie-core` run passed cleanly.

### DIFF
--- a/crates/genie-core/src/llm/genie_ai_runtime.rs
+++ b/crates/genie-core/src/llm/genie_ai_runtime.rs
@@ -1,0 +1,55 @@
+use anyhow::Result;
+use async_trait::async_trait;
+
+use super::openai_compat::OpenAiCompatClient;
+use super::{LlmBackendClient, Message, ResponseFormat};
+
+/// Adapter for the `genie-ai-runtime` OpenAI-compatible chat API surface.
+pub struct GenieAiRuntimeBackend {
+    inner: OpenAiCompatClient,
+}
+
+impl GenieAiRuntimeBackend {
+    pub fn new(host: &str, port: u16) -> Self {
+        Self {
+            inner: OpenAiCompatClient::new("genie-ai-runtime", host, port),
+        }
+    }
+
+    pub fn from_url(url: &str) -> Self {
+        Self {
+            inner: OpenAiCompatClient::from_url("genie-ai-runtime", url),
+        }
+    }
+}
+
+#[async_trait]
+impl LlmBackendClient for GenieAiRuntimeBackend {
+    fn backend_name(&self) -> &str {
+        self.inner.backend_name()
+    }
+
+    async fn health(&self) -> bool {
+        self.inner.health().await
+    }
+
+    async fn chat_with_format(
+        &self,
+        messages: &[Message],
+        max_tokens: Option<u32>,
+        response_format: Option<ResponseFormat>,
+    ) -> Result<String> {
+        self.inner
+            .chat_with_format(messages, max_tokens, response_format)
+            .await
+    }
+
+    async fn chat_stream(
+        &self,
+        messages: &[Message],
+        max_tokens: Option<u32>,
+        on_token: &mut (dyn for<'a> FnMut(&'a str) + Send),
+    ) -> Result<String> {
+        self.inner.chat_stream(messages, max_tokens, on_token).await
+    }
+}
--- a/crates/genie-core/src/llm/llama_cpp.rs
+++ b/crates/genie-core/src/llm/llama_cpp.rs
@@ -1,0 +1,55 @@
+use anyhow::Result;
+use async_trait::async_trait;
+
+use super::openai_compat::OpenAiCompatClient;
+use super::{LlmBackendClient, Message, ResponseFormat};
+
+/// llama.cpp `--server` adapter for the OpenAI-compatible chat API.
+pub struct LlamaCppBackend {
+    inner: OpenAiCompatClient,
+}
+
+impl LlamaCppBackend {
+    pub fn new(host: &str, port: u16) -> Self {
+        Self {
+            inner: OpenAiCompatClient::new("llama.cpp", host, port),
+        }
+    }
+
+    pub fn from_url(url: &str) -> Self {
+        Self {
+            inner: OpenAiCompatClient::from_url("llama.cpp", url),
+        }
+    }
+}
+
+#[async_trait]
+impl LlmBackendClient for LlamaCppBackend {
+    fn backend_name(&self) -> &str {
+        self.inner.backend_name()
+    }
+
+    async fn health(&self) -> bool {
+        self.inner.health().await
+    }
+
+    async fn chat_with_format(
+        &self,
+        messages: &[Message],
+        max_tokens: Option<u32>,
+        response_format: Option<ResponseFormat>,
+    ) -> Result<String> {
+        self.inner
+            .chat_with_format(messages, max_tokens, response_format)
+            .await
+    }
+
+    async fn chat_stream(
+        &self,
+        messages: &[Message],
+        max_tokens: Option<u32>,
+        on_token: &mut (dyn for<'a> FnMut(&'a str) + Send),
+    ) -> Result<String> {
+        self.inner.chat_stream(messages, max_tokens, on_token).await
+    }
+}
--- a/crates/genie-core/src/llm/mod.rs
+++ b/crates/genie-core/src/llm/mod.rs
@@ -1,6 +1,136 @@
-mod client;
+mod genie_ai_runtime;
+mod llama_cpp;
+mod openai_compat;
 mod retry;
 
-pub use client::{LlmClient, Message};
+use anyhow::Result;
+use async_trait::async_trait;
+
+pub use genie_ai_runtime::GenieAiRuntimeBackend;
+pub use llama_cpp::LlamaCppBackend;
+pub use openai_compat::{Message, ResponseFormat};
 #[allow(unused_imports)]
 pub use retry::RetryLlmClient;
+
+#[async_trait]
+pub trait LlmBackendClient: Send + Sync {
+    fn backend_name(&self) -> &str;
+
+    async fn health(&self) -> bool;
+
+    async fn chat_with_format(
+        &self,
+        messages: &[Message],
+        max_tokens: Option<u32>,
+        response_format: Option<ResponseFormat>,
+    ) -> Result<String>;
+
+    async fn chat_stream(
+        &self,
+        messages: &[Message],
+        max_tokens: Option<u32>,
+        on_token: &mut (dyn for<'a> FnMut(&'a str) + Send),
+    ) -> Result<String>;
+}
+
+/// LLM client facade used by agent orchestration.
+///
+/// The public constructors preserve the legacy llama.cpp default. Backend
+/// selection and config plumbing will be layered on top of this facade in a
+/// follow-up PR.
+pub struct LlmClient {
+    backend: Box<dyn LlmBackendClient>,
+}
+
+impl LlmClient {
+    pub fn new(host: &str, port: u16) -> Self {
+        Self::llama_cpp(host, port)
+    }
+
+    pub fn from_url(url: &str) -> Self {
+        Self::from_llama_cpp_url(url)
+    }
+
+    pub fn llama_cpp(host: &str, port: u16) -> Self {
+        Self {
+            backend: Box::new(LlamaCppBackend::new(host, port)),
+        }
+    }
+
+    pub fn from_llama_cpp_url(url: &str) -> Self {
+        Self {
+            backend: Box::new(LlamaCppBackend::from_url(url)),
+        }
+    }
+
+    pub fn genie_ai_runtime(host: &str, port: u16) -> Self {
+        Self {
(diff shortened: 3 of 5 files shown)


ai-hpc (commented):
Read through end-to-end. The boundary you've drawn is the right one and the slice is correctly scoped — happy to see it land as-is, with five small things worth tightening either in this PR or in immediate follow-ups.

## What's good

- **Faithful rename.** `client.rs → openai_compat.rs` with `LlmClient → OpenAiCompatClient(backend_name)` is exactly the right boundary. Error messages now read `"llama.cpp 400: …"` or `"genie-ai-runtime 400: …"` and that small operational win is load-bearing once both backends coexist.
- **Backward compat preserved.** `LlmClient::new` and `LlmClient::from_url` both default to llama.cpp. Existing callers (`main.rs:47`, `voice/mod.rs:45`) compile unchanged.
- **Slice discipline.** PR body explicitly defers `services.llm.backend` config, governor / systemd plumbing, and docs/CI to follow-ups. That is the right way to land a refactor toward #32.
- **The PR doesn't try to be #33 too.** Good. Keep the parity-flip work separate.

## Concrete asks (none are blockers)

### 1. Narrow the `mod.rs` re-exports

```rust
pub use openai_compat::{ChatRequest, ChatResponse, Choice, Delta, Message, ResponseFormat};
```

`ChatRequest`, `ChatResponse`, `Choice`, `Delta` are wire-shape implementation details. Exporting them at crate scope risks accidentally codifying llama.cpp's response shape into the public API — exactly the thing #33's parity work needs the option to diverge on. Suggest:

```rust
pub use openai_compat::{Message, ResponseFormat};
```

and demote the rest to `pub(crate)` (or `pub(super)`) inside `openai_compat.rs`.

### 2. Make the two new unit tests actually verify backend identity

```rust
#[test]
fn legacy_constructor_keeps_llama_cpp_default() {
    let _client = LlmClient::from_url("http://127.0.0.1:8080/health");
}
```

This passes as long as `from_url` doesn't panic — it doesn't verify the inner backend is `LlamaCppBackend`. With `Box<dyn LlmBackendClient>` and no introspection there's nothing to assert.

Cheap fix that pays for itself elsewhere too: add a `backend_name(&self) -> &str` method on the trait (you already plumb `backend_name: &'static str` through `OpenAiCompatClient`), then:

```rust
assert_eq!(client.backend_name(), "llama.cpp");
assert_eq!(genie_client.backend_name(), "genie-ai-runtime");
```

Bonus: that method becomes useful for `genie-ctl status` / `/api/health` / log filtering once #33 lands.

### 3. Call out the `Send` bound widening on `chat_stream` in the PR body

The trait method signature forces the public facade to add `Send`:

```rust
// before
where F: FnMut(&str)
// after
where F: FnMut(&str) + Send
```

I grep'd all in-tree callers (`repl.rs:138,201`, `server.rs:475`, `voice/mod.rs:178,219`, `voice/streaming.rs:32`, `voice/pipeline.rs:122`) — every closure captures only `Send` data, so this is **not breaking in-tree**. But it IS a tightened public contract. Worth one sentence in the PR body so a future caller wrapping an `Rc<_>` knows what changed and why. Same expansion applies in `retry.rs:87`.

### 4. Mark the llama.cpp-specific quirks in `openai_compat.rs` as backend-specific

`flatten_system_into_first_user` exists because of llama.cpp's complaint about consecutive system messages. It's correct for llama.cpp, may or may not be correct for `genie-ai-runtime`. Right now it runs unconditionally for every backend that uses `OpenAiCompatClient`. A `// TODO(#33): verify parity with genie-ai-runtime; may need per-backend behavior` near that function would set the right expectation for the next person.

Same applies to the `chat_json` doc string which now says "Uses backend response-format support when available" — true and accurate, but worth a parity note since `genie-ai-runtime`'s response-format conformance is one of the explicit items in the #33 parity gate.

### 5. The flaky memory-promotion test is worth its own issue

Your PR note says it failed once on a full run then passed. Don't let that sit as an unwritten footnote — file a one-line "flaky test: memory promotion" issue and link it. Otherwise the next contributor hits the same surprise and assumes their change caused it.

## Verdict

The trait + facade + per-backend adapter shape is right. The renames are faithful. The default behavior is preserved. The split is honest. Items #1 and #2 are the only ones I'd actually want before merge; #3-#5 can land as the immediate follow-up.

Approve from my side once #1 and #2 are in.

ai-hpc:
@carlos4s — two small follow-up asks from [my review above](https://github.com/GeniePod/genie-claw/pull/35#pullrequestreview-3556168), pulled out as direct requests so they're easy to action. Both are quick:

### Ask 1 — Narrow the `mod.rs` re-exports

In `crates/genie-core/src/llm/mod.rs`:

```rust
// current
pub use openai_compat::{ChatRequest, ChatResponse, Choice, Delta, Message, ResponseFormat};

// suggested
pub use openai_compat::{Message, ResponseFormat};
```

`ChatRequest`, `ChatResponse`, `Choice`, `Delta` are llama.cpp-shaped wire types and shouldn't be part of the public crate API — exporting them now risks codifying llama.cpp's response shape, which is exactly the surface `genie-ai-runtime` may need to diverge on (per the #33 parity gate). Demoting them to `pub(crate)` (or `pub(super)`) inside `openai_compat.rs` keeps the facade's contract narrow.

### Ask 2 — Add `backend_name()` to the trait so the new tests verify identity

You already plumb `backend_name: &'static str` through `OpenAiCompatClient`. Surfacing it on the trait makes both new tests verify *which* backend was constructed, not just "didn't panic":

```rust
#[async_trait]
pub trait LlmBackendClient: Send + Sync {
    fn backend_name(&self) -> &str;
    async fn health(&self) -> bool;
    // ... existing methods
}
```

Then in `LlmClient`:

```rust
pub fn backend_name(&self) -> &str { self.backend.backend_name() }
```

And the tests become:

```rust
#[test]
fn legacy_constructor_keeps_llama_cpp_default() {
    let client = LlmClient::from_url("http://127.0.0.1:8080/health");
    assert_eq!(client.backend_name(), "llama.cpp");
}

#[test]
fn can_construct_genie_ai_runtime_client() {
    let client = LlmClient::from_genie_ai_runtime_url("http://127.0.0.1:8080/health");
    assert_eq!(client.backend_name(), "genie-ai-runtime");
}
```

Bonus: that accessor lands useful elsewhere later — `genie-ctl status`, `/api/health`, log filtering once the config-side selection lands.

Both should be ~15 line changes. Happy to approve once those are in. The four other items in the review (Send-bound note, parity TODO, flaky test) are fine as follow-ups.

carlos4s:
Addressed the requested review fixes in `973ef98`:

- Narrowed the `llm` public re-exports to `Message` and `ResponseFormat`, and kept the OpenAI-compatible wire shapes crate-private.
- Added `backend_name()` on the backend trait/facade and asserted the two constructor backend names in tests.
- Updated the PR body to mention the `chat_stream` `Send` bound widening.

Verification:
- `cargo check -p genie-core`
- `cargo test -p genie-core llm::`
- `cargo test -p genie-core`

ai-hpc (approved):
Both asks addressed cleanly:

- Re-exports in `mod.rs` are now `pub use openai_compat::{Message, ResponseFormat};` and the wire types (`ChatRequest`/`ChatResponse`/`Choice`/`Delta`) are `pub(crate)`. Public surface stays narrow, leaves #33 free to diverge response shapes if needed.
- `LlmBackendClient::backend_name(&self) -> &str` lands on the trait, surfaced on `LlmClient`, both backends implement it, and the two tests now assert backend identity instead of just construction.

Squash-merging. The four follow-up items from the original review (Send-bound note, `flatten_system_into_first_user` TODO, flaky test, deferred config/governor work) belong in their own follow-up PRs/issues toward #32.

ai-hpc:
Merged at `60a1ec8f48c4a04fddc5de15e295040b0133b355` — thanks @carlos4s!

Issue #34, opened 2026-05-15
Add basic CI pipeline (fmt, clippy, test, aarch64 cross-compile, audit)


ai-hpc:
## Background

There is currently no CI in this repository (`.github/` is absent). Every check that exists — `cargo fmt`, `cargo test`, `make jetson` — runs only when a contributor remembers to run it locally. That's tolerable for one or two committers; it does not scale, and it means regressions in the Jetson cross-compile target only surface at deploy time on the device.

The project structure is well-shaped for CI:

- Rust workspace, **9 crates** (`crates/genie-common`, `genie-core`, `genie-ctl`, `genie-governor`, `genie-health`, `genie-api`, `genie-skill-sdk`, plus `skills/hello-world` and the root)
- Rust edition **2024** (requires rustc ≥ 1.85)
- Cross-compile target `aarch64-unknown-linux-gnu` already wired in `.cargo/config.toml` and `make jetson`
- `bundled` SQLite — no external system library dependencies, CI is straightforward
- AGPL-3.0-only license — a `cargo-deny` license scan is worth having

This issue proposes a minimal but useful first pipeline that covers the daily checks and the one cross-arch invariant. Bigger items (Jetson hardware-in-the-loop test, signed release artifacts, container builds, Python typing for `deploy/scripts/*.py`) belong in follow-ups.

## Goals

1. Catch formatting and lint regressions on every PR.
2. Catch build / test breakage on every PR across the **workspace**.
3. Catch breakage in the **Jetson cross-compile** target on every PR (this is the one most likely to silently rot).
4. Catch dependency advisories before they reach a release.
5. Cache aggressively — first build green; subsequent should be < 3 min.
6. Stay free-tier on `ubuntu-latest` runners; no self-hosted runner requirement yet.

## Non-goals (intentionally deferred)

- Running on real Jetson hardware (needs self-hosted runner; track in a separate issue).
- Releasing signed binaries (track in a separate issue once #32 / #33 land and the backend story is settled).
- Coverage reporting (low ROI on a Rust workspace at this stage).
- Container image builds (the docker-compose path is dev-only).

## Proposed workflows

Land these as three separate files under `.github/workflows/` so they can be enabled, debugged, and tuned independently.

### 1. `.github/workflows/ci.yml` — fmt + clippy + test (the daily loop)

```yaml
name: ci

on:
  pull_request:
  push:
    branches: [main]

env:
  CARGO_TERM_COLOR: always
[...]

PR #36, opened 2026-05-15
chore: fix fmt + clippy drift in genie-core/voice (issue #34 prep)

statxc:
Preparation work for issue #34 (CI pipeline). On `rustc 1.95` the current `main` has small fmt and clippy drift that the new `cargo fmt --check` + `cargo clippy -- -D warnings` jobs would block on. This PR is the minimal, mechanical cleanup to make the tree green so the CI workflows in the follow-up PRs pass on day 1.

## Changes

- **`voice/stt.rs`, `voice/tts.rs`** — collapse three `if let Ok(_) { if … }` chains into the `if let Ok(_) && …` let-chain form (`clippy::collapsible_if`).
- **`voice/stt.rs`** — fix doc-list indentation on the `Denoiser` enum docs (`clippy::doc_list_item_overindented`).
- **`voice/aec.rs`** — annotate `process_aec_skips_stale_reference` with `#[allow(clippy::await_holding_lock)]` plus a comment explaining the test-serialization role. The std `TEST_LOCK` Mutex serializes ECHO_REF tests; holding it across awaits is safe because it gates only sibling tests in this module, never production code.
- **`voice_loop.rs`** — apply rustfmt's preferred line-wrapping for the `audio_output_device` if/else (cosmetic, no behavior change).

## Verification

- `cargo fmt --all -- --check` — clean
- `cargo clippy --workspace --all-targets --locked -- -D warnings` — clean
- `cargo test --workspace --locked --all-targets` — 457 tests pass

No public API changes. No runtime behavior changes.

Refs #34.

### DIFF
--- a/crates/genie-core/src/voice/aec.rs
+++ b/crates/genie-core/src/voice/aec.rs
@@ -338,6 +338,10 @@ mod tests {
         assert!(!has_ref, "Reference should be cleared");
     }
 
+    // TEST_LOCK serializes tests that touch the global ECHO_REF; the guard is
+    // intentionally held across awaits because it only gates other tests in
+    // this module, never production code.
+    #[allow(clippy::await_holding_lock)]
     #[tokio::test]
     async fn process_aec_skips_stale_reference() {
         let _guard = TEST_LOCK.lock().unwrap();
--- a/crates/genie-core/src/voice/stt.rs
+++ b/crates/genie-core/src/voice/stt.rs
@@ -22,10 +22,10 @@ pub fn audio_captured_at() -> Option<std::time::Instant> {
 }
 
 fn mark_audio_captured() {
-    if let Ok(mut g) = AUDIO_CAPTURED_AT.lock() {
-        if g.is_none() {
-            *g = Some(std::time::Instant::now());
-        }
+    if let Ok(mut g) = AUDIO_CAPTURED_AT.lock()
+        && g.is_none()
+    {
+        *g = Some(std::time::Instant::now());
     }
 }
 
@@ -213,8 +213,7 @@ impl SttEngine {
         );
         let body_end = format!("\r\n--{boundary}--\r\n");
 
-        let content_length =
-            text_parts.len() + file_part.len() + wav_data.len() + body_end.len();
+        let content_length = text_parts.len() + file_part.len() + wav_data.len() + body_end.len();
 
         let stream = tokio::net::TcpStream::connect(&addr).await?;
         let (reader, mut writer) = stream.into_split();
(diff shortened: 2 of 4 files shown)


statxc:
@ai-hpc Please review this PR.

ai-hpc (approved):
Clean, mechanical, well-scoped. Read every hunk, traced semantics, verified the one substantive change. Approving.

## What the diff actually does

**Three let-chain collapses** (`stt.rs:22 mark_audio_captured`, `tts.rs:39 mark_first_audio`, `tts.rs:46 mark_first_speak_called`) — `if let Ok(mut g) = X.lock() { if g.is_none() { … } }` → `if let Ok(mut g) = X.lock() && g.is_none() { … }`. Semantically identical: the let-chain `&&` short-circuits the same way, and the guard scope is identical (from acquisition through end-of-if-body in both forms). The mutex is held for the same duration.

**Doc-list indentation** on `Denoiser` enum — `clippy::doc_list_item_overindented`. Renders identically in rustdoc, just satisfies the lint.

**Pure rustfmt reflow** across `stt.rs` (`content_length`, `Denoiser::from_config`, `preprocess_capture` match arms, `sox_cmd.args`, three `String::from_utf8_lossy(...).lines().next().unwrap_or("")` chains) and `voice_loop.rs` (`audio_output_device` if/else). No semantic change.

**One `#[allow(clippy::await_holding_lock)]`** on `process_aec_skips_stale_reference`. Verified the justification by reading the rest of the file:

```rust
// aec.rs:241
static TEST_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
```

Four tests use it. Three are `#[test]` (sync), one is `#[tokio::test]` (async — this one). The std::sync::Mutex serializes access to the `static ECHO_REF` global across the whole test module. The async test holds `_guard` through several `tokio::fs::*` and `process_aec()` awaits, but:

- The other TEST_LOCK consumers are sync `#[test]`s running on threads from the test runner — not on tokio's executor.
- No other task on this tokio runtime tries to acquire `TEST_LOCK`.
- `_guard` is only ever held inside tests; never in production code paths.

So the lint warning is correct in general but doesn't apply here. The `#[allow]` is the right surgical fix, and the three-line comment explaining the scope (`test serialization, not production`) is exactly the form an `#[allow]` should take.

## Other observations

- **No public API change** confirmed. `Denoiser::from_config`'s signature is syntactically reformatted but type-identical.
- **No new dependencies.**
- **Tests still pass** per the author's `cargo test --workspace --locked --all-targets` — 457 tests.
- **Sequencing is right.** Landing this *before* the CI workflows in #34 means the first CI run won't be a sea of red. Exactly the right ordering.

## Minor FYI (not a blocker)

The let-chain `if let X && Y` syntax stabilized in **rustc 1.88**. `Cargo.toml` says `edition = "2024"` which only mandates rustc 1.85+, so this PR implicitly raises the effective MSRV to 1.88. Worth either pinning `rust-version = "1.88"` in `[workspace.package]` or noting in the CI workflows from #34 that `dtolnay/rust-toolchain@stable` is the required toolchain channel. Neither needs to land in this PR.

## Possible refactor (also a non-blocker)

The three collapsed `mark_*` patterns are all "lazily set an `Option<Instant>` behind a Mutex if currently None." Could be a `mark_once(cell: &Mutex<Option<Instant>>)` helper in one shared location. Out of scope for a lint-cleanup PR; mentioning so it's on someone's radar.

Approving. Squash-merge whenever convenient.

ai-hpc:
Merged at `09b5d6fdc8f309f046b72620947fb3f59e2b0569` — thanks @statxc!

PR #38, opened 2026-05-16
voice: report configured llm backend

carlos4s:
## Summary
- Voice-only follow-up after #35 landed the LLM backend facade.
- Reports the configured backend name in voice health and LLM error paths.
- Removes the remaining voice-mode operator hint that hard-coded `llama-server`.
- Adds regression coverage for backend-neutral voice LLM status messages.

## Scope
This intentionally only touches `crates/genie-core/src/voice*` paths. It does not add `[services.llm].backend` config plumbing or change the default backend; those can stay as separate follow-up PRs if the maintainer wants smaller slices.

## Tests
- `cargo check -p genie-core`
- `cargo test -p genie-core voice_loop::tests`
- `cargo test -p genie-core voice::`
- `git diff --check`

PR #39, opened 2026-05-16
core: expose configured llm backend

carlos4s:
## Summary
- Follow-up backend-abstraction slice after #35 and #38.
- Keeps scope inside `crates/genie-core/src` only.
- Logs the configured `LlmClient` backend at core startup.
- Adds `llm_backend` to `/api/health` while preserving the existing `llm` connected/offline field.
- Adds regression coverage with a `genie-ai-runtime` facade client.

## Scope
This intentionally does not add config parsing or change the default backend. It only surfaces the backend identity that the `LlmClient` facade already knows.

## Tests
- `cargo check -p genie-core`
- `cargo test -p genie-core server::tests::health_endpoint_reports_llm_backend`
- `cargo test -p genie-core server::tests`
- `git diff --check`

### DIFF
--- a/crates/genie-core/src/main.rs
+++ b/crates/genie-core/src/main.rs
@@ -45,6 +45,7 @@ async fn main() -> Result<()> {
     // Build shared components.
     let llm_url = &config.services.llm.url;
     let llm = llm::LlmClient::from_url(llm_url);
+    tracing::info!(backend = %llm.backend_name(), "LLM backend configured");
 
     let ha = ha::provider_from_config(&config);
 
--- a/crates/genie-core/src/server.rs
+++ b/crates/genie-core/src/server.rs
@@ -972,6 +972,7 @@ async fn handle_health(
     let resp = serde_json::json!({
         "status": status,
         "llm": if llm_ok { "connected" } else { "offline" },
+        "llm_backend": llm.backend_name(),
         "memories": mem_count,
         "conversations": conv_count,
         "mem_available_mb": mem_avail,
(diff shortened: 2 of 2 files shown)


ai-hpc (approved):
Third slice toward #32 after #35 (facade + `backend_name()`) and #38 (voice log surfaces). Same author, same scope discipline, body matches diff.

## What lands

Two additive surfaces, both useful operationally:

- **`main.rs:48`** — one `tracing::info!(backend = %llm.backend_name(), "LLM backend configured")` at startup. The first thing in journalctl now names which backend the daemon picked up. Costs one log line, saves one round-trip every time an operator wonders "which one is this Jetson running."

- **`server.rs:975`** — new `llm_backend` field in the `/api/health` JSON body, alongside the preserved legacy `llm: connected|offline`. Critical for `genie-api` dashboard, Prometheus exporters, and any external monitoring — they can now distinguish llama.cpp from genie-ai-runtime per-instance, machine-readably. Field is additive; nothing existing breaks.

The new regression test (`server::tests::health_endpoint_reports_llm_backend`) calls the real `handle_health` against an unreachable `from_genie_ai_runtime_url("http://127.0.0.1:1/health")`, parses the JSON, and asserts both the preserved `llm = "offline"` and the new `llm_backend = "genie-ai-runtime"`. Right shape — exercises the actual handler, verifies identity flow from constructor → `backend_name()` → JSON output, uses unique PID+nanosecond temp paths so parallel runs can't collide, and cleans up after itself.

1. Startup uses `tracing::info!` here while #38's voice startup uses `eprintln!` — two valid surfaces (structured logs vs operator console). Worth picking one convention long-term, but not in this PR's scope.
2. Test asserts only the two fields it cares about (`llm`, `llm_backend`), not a full health-endpoint schema. Right call — targeted regression coverage, not a contract test.
3. Hard-coded `127.0.0.1:1` for the unreachable URL is fine; if it ever flakes on a runner where port 1 is bound (very unlikely), `127.0.0.1:0` is deterministically unreachable for HTTP.

## Sequence after this lands

- #35 — facade + accessor
- #38 — voice log surfaces
- **#39 — startup log + health-endpoint JSON (this PR)**
- Still ahead: `[services.llm].backend` config parsing, governor/health awareness of backend selection, default flip (#33).

Approving.

ai-hpc:
Merged at `1435ee4c6de8a605591cc16329c9204d145c27f2`.
Thanks to @carlos4s!

PR #40, opened 2026-05-16
runtime: resolve llm service from config

carlos4s:
## Summary
- Follow-up for the runtime/service-tooling part of the LLM backend abstraction work.
- Replaces governor mode hard-codes of `genie-llm` with a generic `llm` service alias.
- Resolves `llm` through `services.llm.systemd_unit`, so governor starts/stops/restarts the configured LLM service.
- Updates `genie-ctl health`, diagnostics, and support bundles to use core `/api/health` for backend-neutral LLM status instead of probing llama.cpp directly.
- Keeps `genie-health` on the existing config-driven `services.llm.url` path and verifies it still passes.

## Scope
This PR intentionally does not add `[services.llm].backend` config parsing or change the default backend. It only makes runtime/service tooling honor the existing configured LLM service unit and backend-neutral health reporting.

## Tests
- `cargo test -p genie-common config::tests::service_unit_aliases_use_configured_units`
- `cargo test -p genie-governor`
- `cargo test -p genie-ctl`
- `cargo test -p genie-health`
- `cargo check -p genie-common -p genie-governor -p genie-ctl -p genie-health`
- `git diff --check`

### DIFF
--- a/crates/genie-common/src/config.rs
+++ b/crates/genie-common/src/config.rs
@@ -159,7 +159,7 @@ pub struct CoreConfig {
     #[serde(default = "defaults::voice_continuous_secs")]
     pub voice_continuous_secs: u32,
 
-    /// LLM model path (for GPU time-sharing — voice loop restarts llama-server).
+    /// LLM model path used when runtime tooling swaps the configured LLM service.
     #[serde(default = "defaults::llm_model_path")]
     pub llm_model_path: PathBuf,
 
@@ -594,10 +594,38 @@ impl Config {
     /// Whether the current deployment should manage a given service alias.
     pub fn manages_service_alias(&self, alias: &str) -> bool {
         match alias {
+            "core" | "genie-core" | "llm" | "genie-llm" => true,
             "homeassistant" => self.services.homeassistant.is_some(),
             "nextcloud" => self.services.nextcloud.is_some(),
             "jellyfin" => self.services.jellyfin.is_some(),
             _ => true,
+        }
+    }
+
+    /// Resolve a service alias used by runtime tooling to its configured
+    /// systemd unit. Optional services return `None` when they are not
+    /// configured for this deployment.
+    pub fn service_unit_for_alias(&self, alias: &str) -> Option<String> {
+        match alias {
+            "core" | "genie-core" => Some(self.services.core.systemd_unit.clone()),
+            "llm" | "genie-llm" => Some(self.services.llm.systemd_unit.clone()),
+            "homeassistant" => self
+                .services
+                .homeassistant
+                .as_ref()
+                .map(|service| service.systemd_unit.clone()),
+            "nextcloud" => self
+                .services
+                .nextcloud
+                .as_ref()
+                .map(|service| service.systemd_unit.clone()),
+            "jellyfin" => self
+                .services
+                .jellyfin
+                .as_ref()
+                .map(|service| service.systemd_unit.clone()),
+            _ if self.manages_service_alias(alias) => Some(alias.to_string()),
+            _ => None,
         }
     }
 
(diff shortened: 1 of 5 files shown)


ai-hpc (approved):
Fourth slice toward #32 after #35 / #38 / #39. Largest of the four so far at +235/-41 across 5 files, but the size is structurally justified — it replaces hard-coded `genie-llm` references in three different consumers (mode definitions, governor mode-transition logic, `genie-ctl` operator tools) with a single config-driven resolver. Body matches diff.

## What lands

**`genie-common/src/config.rs`** — single source of truth via new `Config::service_unit_for_alias(alias) -> Option<String>`, plus `"core" | "genie-core" | "llm" | "genie-llm"` added to `manages_service_alias`. Right abstraction layer: the alias→unit mapping lives once in Config, and every consumer (governor, ctl) routes through it.

**`genie-common/src/mode.rs`** — every `"genie-llm"` literal in `Mode::required_services()` and `stopped_services()` becomes `"llm"`. This is the critical decoupling. Modes now refer to a logical service; config decides which systemd unit serves that role. When #33 flips the default to `genie-ai-runtime.service`, `mode.rs` doesn't need a single line touched.

**`genie-governor/src/governor.rs`** — mode-transition loops resolve aliases via `service_unit_for_alias()`. Three LLM model-swap call sites now pass the resolved unit to the parametrized `swap_llm_model(unit, path)`. New `llm_service_unit()` fallback helper. New regression test `resolves_llm_alias_to_configured_service_unit` exercising the override.

**`genie-governor/src/service_ctl.rs`** — `swap_llm_model` signature changes from `(model_path)` to `(unit, model_path)`. **This is the one that would have silently broken `genie-ai-runtime` backend switching had it been missed** — the systemd override directory `/etc/systemd/system/genie-llm.service.d/` is now derived from the unit name, so an operator switching to `genie-ai-runtime.service` gets overrides in the matching `.d` directory instead of writing to the wrong one. New `normalize_systemd_unit()` + `llm_override_dir_for_unit()` helpers with tests.

**`genie-ctl/src/main.rs`** — `cmd_health` and `cmd_diag` consume `/api/health` JSON for backend identity instead of probing `127.0.0.1:8080/health` for llama.cpp directly. Output label is now dynamic: `[OK] LLM (genie-ai-runtime)` vs `[OK] LLM (llama.cpp)`. `binary_inventory` adds `genie-ai-runtime` alongside `llama-server`. `config_presence` resolves systemd unit paths from Config with a clean fallback if `Config::load` fails.

## Tests

Eight new tests across four files, each targeted and meaningful:

- `config::tests::service_unit_aliases_use_configured_units` — verifies the resolver returns configured unit names including a non-default `genie-ai-runtime.service` override
- `governor::tests::resolves_llm_alias_to_configured_service_unit` — exercises the new helper from a Governor instance
- `service_ctl::tests::normalizes_systemd_unit_names` and `llm_override_dir_uses_configured_unit`
- `genie-ctl::tests::llm_service_label_includes_backend_when_known` and `systemd_unit_file_path_normalizes_unit_name`
- Plus updated existing tests (`mode_required_services`, `mode_stopped_services`, `should_manage_service`) to assert the new `"llm"` alias

## Observations (non-blockers)

1. **Runtime dependency on #39.** The new `cmd_health` reads `llm` and `llm_backend` from core's `/api/health` JSON. Those fields require #39 to have shipped on the running core. #39 just merged at `1435ee4c`, so this is fine — but a newer `genie-ctl` against an older core would print `[DOWN] LLM` as a false negative. Worth noting in the changelog so operators upgrade ctl + core together.

2. **`stopped_services()` loop doesn't `is_active` check.** The required-services loop guards `start` with `if !is_active`, but the stopped-services loop unconditionally calls `stop`. This matches prior behavior (systemctl stop on an inactive unit is harmless), just noting the asymmetry.

3. **Doc comment on `CoreConfig::llm_model_path`** lost the "GPU time-sharing" context. Was `voice loop restarts llama-server`; now `runtime tooling swaps the configured LLM service`. Backend-neutral is correct, but a reader gets less operational context. Tiny tweak like "used during GPU time-sharing mode transitions, where runtime tooling swaps the configured LLM service" would carry both. Not worth holding the PR for.

4. **The `_ if self.manages_service_alias(alias) => Some(alias.to_string())` fallback** in `service_unit_for_alias` is currently unreachable (the explicit arms cover every alias `manages_service_alias` returns true for). Fine as defensive plumbing in case the alias list grows.

## Sequence after this lands

- #35 — facade + `backend_name()`
- #38 — voice log surfaces
- #39 — startup log + `/api/health` field
- **#40 — mode-transition unit resolution + ctl backend-neutral status (this PR)**
- Still ahead: `[services.llm].backend` config parsing (the actual feature toggle), default flip (#33).

After this, the only thing left to make `genie-ai-runtime` selectable end-to-end is the config key parsing + the constructor branching in `main.rs`. The runtime plumbing is now complete.

Approving.

ai-hpc:
Merged at `311c948c2635459247fede19acdd7df810eb91f9`.
Thanks to @carlos4s!

PR #43, opened 2026-05-16
llm: select backend from config

carlos4s:
## Summary
- Adds `backend` parsing on service endpoints, defaulting to `llama_cpp` for legacy configs.
- Supports `backend = "genie_ai_runtime"` on `[services.llm]`.
- Builds the core `LlmClient` from `services.llm.backend` instead of always using the llama.cpp constructor.
- Applies the same config-based construction in the voice orchestrator path.
- Adds focused tests for legacy default parsing and `genie_ai_runtime` selection.

## Scope
This is the config-selection slice only. It does not change the default backend, add deploy TOML examples, update docs, or introduce Cargo feature gates.

## Tests
- `cargo test -p genie-common llm_service_backend`
- `cargo test -p genie-core llm::tests`
- `cargo test -p genie-common`
- `cargo check -p genie-common -p genie-core -p genie-health -p genie-governor -p genie-ctl`
- `git diff --check`

### DIFF
--- a/crates/genie-common/src/config.rs
+++ b/crates/genie-common/src/config.rs
@@ -556,6 +556,20 @@ pub struct Esp32C6UartConfig {
 pub struct ServiceEndpoint {
     pub url: String,
     pub systemd_unit: String,
+    /// LLM backend selector. Only meaningful for `services.llm`; defaults
+    /// preserve the legacy llama.cpp path for existing configs.
+    #[serde(default)]
+    pub backend: LlmBackendKind,
+}
+
+#[derive(Debug, Deserialize, Clone, Copy, PartialEq, Eq, Default)]
+#[serde(rename_all = "snake_case")]
+pub enum LlmBackendKind {
+    #[default]
+    #[serde(alias = "llama-cpp")]
+    LlamaCpp,
+    #[serde(alias = "genie-ai-runtime")]
+    GenieAiRuntime,
 }
 
 impl Config {
@@ -774,10 +788,12 @@ impl Default for ServicesConfig {
             core: ServiceEndpoint {
                 url: "http://127.0.0.1:3000/api/health".into(),
                 systemd_unit: "genie-core.service".into(),
+                backend: LlmBackendKind::default(),
             },
             llm: ServiceEndpoint {
                 url: "http://127.0.0.1:8080/health".into(),
                 systemd_unit: "genie-llm.service".into(),
+                backend: LlmBackendKind::LlamaCpp,
             },
             homeassistant: None,
             nextcloud: None,
@@ -900,6 +916,7 @@ bind_host = "0.0.0.0"
         config.services.nextcloud = Some(ServiceEndpoint {
             url: "http://127.0.0.1:8180/status.php".into(),
             systemd_unit: "nextcloud.service".into(),
+            backend: LlmBackendKind::default(),
         });
 
         assert!(config.manages_service_alias("genie-core"));
@@ -916,6 +933,7 @@ bind_host = "0.0.0.0"
         config.services.nextcloud = Some(ServiceEndpoint {
             url: "http://127.0.0.1:8180/status.php".into(),
             systemd_unit: "nextcloud.service".into(),
+            backend: LlmBackendKind::default(),
         });
 
         assert_eq!(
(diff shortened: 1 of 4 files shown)


ai-hpc (approved):
Fifth slice toward #32 after #35 / #38 / #39 / #40. This is the one that makes the backend selectable from config — and after it lands, #33 (default flip) becomes a one-line change in `deploy/config/geniepod.toml`. Body matches diff, scope is honest, defaults preserved.

## What lands

**`genie-common/src/config.rs`** — new `LlmBackendKind` enum (`LlamaCpp` default, `GenieAiRuntime`) with `#[serde(rename_all = "snake_case")]` plus alias support for both hyphenated (`llama-cpp`, `genie-ai-runtime`) and snake_case (`llama_cpp`, `genie_ai_runtime`) spellings. Added as `backend: LlmBackendKind` on `ServiceEndpoint` with `#[serde(default)]`. The `ServicesConfig::Default` impl explicitly sets `LlamaCpp` on the `llm` endpoint, preserving today's behavior for any deployment that doesn't yet add the `backend = "..."` line.

**`genie-core/src/llm/mod.rs`** — new `LlmClient::from_service_config(&ServiceEndpoint)` factory that matches on `service.backend` and dispatches to either `from_llama_cpp_url` or `from_genie_ai_runtime_url`. Doc comment on `LlmClient` updated to point to this as the canonical construction path.

**`genie-core/src/main.rs` and `voice/mod.rs`** — both call sites flip from `from_url(&config.services.llm.url)` to `from_service_config(&config.services.llm)`. Single-line behavioral change at each site; no other call sites left over.

**Tests** — three targeted additions:
- `llm_service_backend_defaults_to_llama_cpp` — `ServiceEndpoint` without `backend` parses as `LlamaCpp`. This is the regression-gate against accidentally breaking existing configs.
- `llm_service_backend_accepts_genie_ai_runtime` — explicit `backend = "genie_ai_runtime"` parses as the new variant.
- `service_config_selects_genie_ai_runtime_backend` — exercises `from_service_config` end-to-end and asserts `backend_name() == "genie-ai-runtime"`, proving identity flow from TOML through the factory through the trait.

Both `llama-cpp` / `llama_cpp` and `genie-ai-runtime` / `genie_ai_runtime` are accepted via `#[serde(alias = ...)]`. Small but real ergonomics win — operators reading config docs will write one form, GeniePod-internal naming uses the other, both work.

## Observations (non-blockers)

1. **`ServiceEndpoint::backend` lives on the generic struct.** It's only meaningful for `services.llm` — the doc comment correctly says so — but a `[services.homeassistant] backend = "genie_ai_runtime"` would parse cleanly and silently do nothing. A future validation pass could warn on non-default `backend` on non-LLM endpoints. Optional defensive polish; not in scope here.

2. **No `unknown_variant` test.** A `backend = "tensorrt_llm"` value will produce a Serde parse error (the right behavior), but there's no test pinning that contract. Cheap to add as a follow-up; not worth holding the PR for.

3. **`RetryLlmClient` (per #35) doesn't get a `from_service_config` constructor.** Not a live problem — neither `main.rs` nor `voice/mod.rs` uses `RetryLlmClient` — but if a future caller wraps an `LlmClient` they'll need to do `RetryLlmClient::new(LlmClient::from_service_config(...))` by hand. A `RetryLlmClient::from_service_config` helper could land in the same follow-up that updates the deploy TOML examples.

4. **`deploy/config/geniepod.toml` not touched.** PR body defers this — fine for the slice, but operators reading the prod config can't see how to flip the backend until that lands. Should be the next follow-up PR (commented-out `backend = "..."` example under `[services.llm]`, plus a docs section).

5. **`LlmBackendKind` derives `Debug, Deserialize, Clone, Copy, PartialEq, Eq, Default`** but not `Serialize` or `Display`. The backend-name string already flows through `backend_name()` for logging, so this isn't urgent, but a round-trip-clean `Serialize` would be useful for `/api/health` if it ever needs to emit the configured-backend value (today it emits the runtime backend identity via `backend_name()`, which is the right answer).

## Sequence

- #35 — facade + `backend_name()`
- #38 — voice operator logs surface backend identity
- #39 — startup log + `/api/health` JSON field
- #40 — runtime/service tooling (mode-transition unit resolution + ctl backend-neutral status)
- **#43 — config selection (this PR)**

After this lands, `genie-ai-runtime` is selectable end-to-end via a one-line change in `geniepod.toml`. #33 (default flip) becomes a one-line TOML change plus the systemd unit + parity-gate validation. Everything else in the abstraction has been built.

Approving.

ai-hpc:
Merged at `6bec4d1e6f0b1c39d08bec0dfdf7e6be33708631`.
Thanks to @carlos4s!

PR #47, opened 2026-05-17
deploy: opt-in genie-ai-runtime backend in setup + systemd (#27)

enjoyandlove:
## Summary

Closes #27

- Wire `[services.llm].backend` (#43, #40, #39) into the deploy layer so opt-in installs validate the right binary and enable the right systemd unit instead of silently falling back to `llama-server` / `genie-llm.service`.
- Add `deploy/systemd/genie-ai-runtime.service` for `jllm-server`, mirroring `genie-llm.service` flag-for-flag so the OpenAI-compatible client in `genie-core` works against either backend unchanged.
- Document the selector, accepted values, and migration steps in [doc/configuration.md](doc/configuration.md); list the new unit in [doc/deployment-and-ops.md](doc/deployment-and-ops.md).

## Changes

- [deploy/setup-jetson.sh](deploy/setup-jetson.sh)
  - Parse `[services.llm].backend` and `[services.llm].systemd_unit` from `geniepod.toml` with a tolerant awk reader (strips quotes, trailing comments, whitespace; accepts hyphenated aliases).
  - Step 5 now branches on the configured backend: checks `llama-server` for `llama_cpp` (default), `jllm-server` for `genie_ai_runtime`, and warns on unrecognized values without aborting.
  - The systemd enable loop and the final "Start services" hint resolve the LLM unit from config instead of hardcoding `genie-llm`.
- [deploy/systemd/genie-ai-runtime.service](deploy/systemd/genie-ai-runtime.service) — new unit for `/opt/geniepod/bin/jllm-server`. Same flag surface as `genie-llm.service`, gated by `ConditionPathExists` so it no-ops when the binary isn't installed.
- [doc/configuration.md](doc/configuration.md) — `[services.llm].backend` table, hyphenated-alias note, and the three-step flip checklist (set `systemd_unit`, install binary, re-run setup).
- [doc/deployment-and-ops.md](doc/deployment-and-ops.md) — add `genie-ai-runtime.service` to the systemd unit inventory.

## Test plan

- [ ] On a config with no `[services.llm].backend`, `setup-jetson.sh` reports `Checking llama.cpp backend (configured: llama_cpp)` and enables `genie-llm.service` (regression — default path unchanged).
- [ ] Set `[services.llm].backend = "genie_ai_runtime"` and `systemd_unit = "genie-ai-runtime.service"`; re-run setup — script reports the genie-ai-runtime check, enables `genie-ai-runtime.service`, and the final hint prints `sudo systemctl start genie-ai-runtime`.
[...]

PR #37, opened 2026-05-15
deps: bump rustls-webpki to 0.103.13 for RUSTSEC-2026-0104 (issue #34 prep)

statxc:
> 🔒 **One-line Cargo.lock bump.** Patches a reachable panic in `rustls-webpki`'s CRL parser ahead of the new `audit.yml` workflow in #34.

## Why this exists

The follow-up `audit.yml` workflow for issue #34 fails on day one against the current lockfile because `rustls-webpki 0.103.12` is affected by [**RUSTSEC-2026-0104**](https://rustsec.org/advisories/RUSTSEC-2026-0104). This PR lands the upstream fix as a standalone, single-file change so the audit pipeline starts clean.

## The advisory

> Reachable panic in `BorrowedCertRevocationList::from_der` / `OwnedCertRevocationList::from_der` when parsing a syntactically valid empty `BIT STRING` in the `onlySomeReasons` element of an `IssuingDistributionPoint` CRL extension.
>
> **This panic is reachable prior to CRL signature verification.**

Applications that don't validate CRLs are not directly exploitable today, but the affected crate is on the transitive HTTPS path:

genie-core ─► reqwest ─► hyper-rustls ─► rustls ─► rustls-webpki

## Fix

| Field | Value |
| :--- | :--- |
| Affected | `rustls-webpki 0.103.12` |
| Fixed in | `rustls-webpki >=0.103.13, <0.104.0-alpha.1` |
| Command | `cargo update -p rustls-webpki` |
| Diff size | 2 lines (version + checksum), `Cargo.lock` only |

No API surface change. No feature-flag change. No source files touched.

## Verification

- [x] `cargo build --release --locked -p genie-core` — clean
- [x] `cargo build --release --locked -p genie-ctl -p genie-governor -p genie-health -p genie-api` — clean
- [x] `cargo test --workspace --locked --all-targets` — **457 / 457 pass**
- [x] aarch64 cross-compile (`make jetson`-equivalent) — all 5 Jetson binaries produce valid ELFs

---

Part of the issue #34 CI work. Refs #34.

PR #48, opened 2026-05-17
deploy: pin genie-* services to dedicated CPU cores for voice latency…

### LINKED ISSUE #25: CPU pinning / core isolation for voice subsystem latency stability
## Motivation

STT latency on the Jetson Orin Nano shows high variance across cycles even after #17 (whisper warmup) loaded the model into iGPU memory:

- Best-case warm: ~245-285 ms
- Mid-session: 3.5 s (observed during #14 chase)
- First-call cold: 60-90 s before #17 (now < 1 s)

Variance after warmup hints at **CPU/GPU contention** — `genie-core`, `llama-server`, `whisper-server`, `piper`, `aplay`, `arecord`, `sox`, and `deep-filter` all schedule freely on the same six cores. When TTS spawns `piper` + `aplay` between cycles, those processes can deschedule `whisper-server`'s HTTP thread, costing milliseconds-to-seconds on the next STT.

For stable latency we want to pin each subsystem to dedicated cores so they don't fight each other on a shared core scheduler.

## Proposal

Jetson Orin Nano has 6 Cortex-A78AE cores (CPUs 0-5). Suggested layout:

| Cores | Pinned to | Why |
| --- | --- | --- |
| 0-1   | system / kernel softirqs / ALSA / `arecord` | Audio I/O is latency-sensitive; keep on standard cores so `irqbalance` works normally. |
| 2-3   | `whisper-server` | Owns STT. Dedicated cores → no schedule-out under LLM/TTS load. |
| 4     | `llama-server` (CPU side; main work is GPU anyway) | Phi-4-mini is iGPU-heavy but the CUDA host driver and HTTP loop still want a core. |
| 5     | `genie-core` + `piper` + `sox` + `deep-filter` | The orchestration shell; bursty CPU but never blocks STT/LLM. |

Implementation options, in increasing complexity:

1. **`CPUAffinity=` in systemd units.** One-line change per service. `genie-whisper.service` → `CPUAffinity=2 3`, `genie-llm.service` → `CPUAffinity=4`, `genie-core.service` → `CPUAffinity=5`. Done.
2. **`isolcpus=2,3,4` on the kernel command line.** Reserves cores from the general scheduler. Combined with `CPUAffinity=`, gives `whisper-server` and `llama-server` exclusive use. Requires editing `/boot/extlinux/extlinux.conf` and rebooting.
3. **Real-time priority for `whisper-server`.** `SCHED_FIFO` + `nice=-10` so it preempts everything else on its cores when an inference request comes in. Worth measuring before assuming the win.

Start with option 1, measure with the #19 latency banner (and ideally an extended version that captures per-cycle distributions), then escalate to options 2-3 only if option 1 leaves residual variance.

### In scope
- Add `[Service] CPUAffinity=...` lines to `genie-whisper.service`, `genie-llm.service`, `genie-core.service`.
- Document the layout in `deploy/setup-jetson.sh`'s `[5b/6]` performance-mode block (operator can see the assignment).
- Provide a `genie-ctl audio-perf` (or similar) command that dumps current core assignments + per-process CPU time so we can verify pinning took effect.
- Measure on hardware: 10-cycle latency variance before vs after pinning, with the #19 banner extended to log per-cycle.

### Out of scope (for this issue)
- `isolcpus=` kernel boot change (option 2). Separate issue if option 1 isn't enough.
- `SCHED_FIFO` real-time (option 3). Separate issue.
- Userspace per-thread affinity inside `genie-core` (tokio worker pinning). Separate issue if `genie-core`-side jitter shows up.

## Acceptance criteria

- 10 consecutive voice cycles' STT latency (after warmup) stays within ±100 ms of median. Today's variance is several hundred ms to several seconds.
- `taskset -cp $(pgrep whisper-server)` reports cores 2-3, `llama-server` reports 4, `genie-core` reports 5.
- No regression on first-reply latency (#19 banner) — pinning should not make cold-start worse.
- `tegrastats --interval 5000` shows whisper / llama / genie-core distributed across distinct cores during a voice cycle, not all crowded on the same one.

## Related

- #2 — earlier iGPU contention work. Same root concern at a different layer (iGPU memory eviction); CPU pinning is the host-side complement.
- #17 — whisper warmup. Pinning is the follow-on lever for reducing STT *variance* now that cold-start is solved.
- #19 — latency banner. Becomes a measurement tool for pinning effectiveness.

enjoyandlove:
## Summary
 Closes #25
- Pins each `genie-*` systemd unit to a dedicated CPU bucket on the Jetson Orin Nano (6 cores) so the voice path (wake → STT → LLM → TTS → playback) no longer competes with itself for scheduler time. Closes #25.
- Bucket layout: **0–1** system + ALSA + MQTT + `genie-api`/`genie-governor`/`genie-health`/`genie-wakeword`, **2–3** `whisper-server` (STT decode, 2 threads), **4** `llama-server` (GPU-bound, hosts CUDA dispatch + sampler), **5** `genie-core` and every audio child it spawns (`piper`, `sox`, `deep-filter`, `arecord`, `aplay`).
- Moves `genie-wakeword` off cores 4–5 (now reserved for llama / core) onto the system bucket 0–1; keeps `SCHED_FIFO` priority 50 so the continuous audio loop is not preempted by best-effort work sharing those cores.
- Documents the bucket map, verification commands (`taskset -pc`, `ps -L -o psr`, `tegrastats`), and the acceptance signal in [doc/deployment-and-ops.md](doc/deployment-and-ops.md) — including why we stopped short of kernel-level `isolcpus=` (requires a Jetson reflash / extlinux edit, out of scope for this PR).

### Real Behavior Proof

- [X]    I have built and run the affected code locally.
- [X]    I have verified the equivalent path with the targeted CI-size-budget test.

## Test plan

- [ ] `sudo systemctl daemon-reload && sudo systemctl restart genie-*.service` on the Orin Nano and confirm every unit reaches `active (running)`.
- [ ] For each service, run `taskset -pc "$(systemctl show -p MainPID --value <svc>.service)"` and confirm the affinity matches the bucket map above.
- [ ] `ps -L -o pid,tid,psr,comm -p "$(pidof whisper-server)"` shows decode threads spread across cores 2–3 only; same check for `llama-server` shows all threads on core 4.
- [ ] Verify `genie-core`'s spawned children (`piper`, `sox`, `deep-filter`, `arecord`, `aplay`) inherit core 5 via `for p in $(pgrep -P "$(systemctl show -p MainPID --value genie-core.service)"); do taskset -pc "$p"; done`.
- [ ] Run 10 consecutive voice cycles end-to-end and capture STT latency; median ± 100 ms is the acceptance signal called out in #25. Watch `sudo tegrastats --interval 250` during the run for unexpected core saturation.
- [ ] Sanity-check wake-word responsiveness on the new 0–1 affinity (SCHED_FIFO 50) — confirm no missed wakes under simultaneous LLM load.

### DIFF
--- a/deploy/systemd/genie-ai-runtime.service
+++ b/deploy/systemd/genie-ai-runtime.service
@@ -31,6 +31,13 @@ Restart=on-failure
 RestartSec=5
 TimeoutStartSec=120
 
+# CPU pinning for voice latency stability (issue #25). Inference is GPU-bound
+# (jetson-llm-server offloads to Jetson iGPU), so one dedicated core is enough
+# to host CUDA dispatch + sampler + tokenizer without contending with whisper
+# (2-3) or genie-core (5). Mirrors genie-llm.service; only one of the two units
+# runs at a time (Conflicts=genie-llm.service above), so both can claim core 4.
+CPUAffinity=4
+
 # GPU needs full system access
 ProtectSystem=no
 SupplementaryGroups=video render
--- a/deploy/systemd/genie-ai-runtime.service
+++ b/deploy/systemd/genie-ai-runtime.service
@@ -1,0 +1,43 @@
+[Unit]
+Description=GeniePod LLM Server (genie-ai-runtime / jllm-server)
+Documentation=https://github.com/GeniePod/genie-ai-runtime
+Documentation=https://github.com/GeniePod/genie-claw/issues/27
+After=network.target
+ConditionPathExists=/opt/geniepod/bin/jllm-server
+
+[Service]
+Type=simple
+# Drop page cache before loading model — Jetson NvMap needs contiguous blocks.
+ExecStartPre=/bin/sh -c 'sync && echo 3 > /proc/sys/vm/drop_caches'
+ExecStart=/opt/geniepod/bin/jllm-server \
+    --model ${GENIEPOD_LLM_MODEL} \
+    --host 0.0.0.0 \
+    --port 8080 \
+    --ctx-size 2048 \
+    --n-gpu-layers 999 \
+    --threads 4 \
+    --parallel 1
+# The jllm-server flag surface intentionally mirrors llama-server so the
+# OpenAI-compatible client in genie-core works against either backend without
+# code changes. Issue #27 tracks the opt-in rollout; flip
+# [services.llm].backend = "genie_ai_runtime" in geniepod.toml after
+# installing /opt/geniepod/bin/jllm-server.
+
+Environment=GENIEPOD_LLM_MODEL=/opt/geniepod/models/phi-4-mini-instruct-q4_k_m.gguf
+Restart=on-failure
+RestartSec=5
+TimeoutStartSec=120
+
+# CPU pinning for voice latency stability (issue #25). Inference is GPU-bound
+# (--n-gpu-layers 999), so one dedicated core is enough to host CUDA dispatch
+# + sampler + tokenizer without contending with whisper (2-3) or genie-core
+# (5). --threads=4 above stays as-is; with a single-core affinity the four
+# worker threads time-share core 4 — fine for the GPU-bound critical path.
+CPUAffinity=4
+
+# GPU needs full system access
+ProtectSystem=no
+SupplementaryGroups=video render
+
+[Install]
+WantedBy=geniepod.target
--- a/deploy/systemd/genie-api.service
+++ b/deploy/systemd/genie-api.service
@@ -12,6 +12,10 @@ Restart=always
 RestartSec=5
 MemoryMax=32M
 
+# Pin to the system bucket (issue #25) so the dashboard's HTTP loop doesn't
+# steal cycles from whisper (2-3), llama-server (4), or genie-core (5).
+CPUAffinity=0 1
+
 ProtectSystem=strict
 ProtectHome=yes
 ReadWritePaths=/opt/geniepod/data
--- a/deploy/systemd/genie-core.service
+++ b/deploy/systemd/genie-core.service
@@ -15,6 +15,15 @@ StandardInput=null
 Restart=always
 RestartSec=3
 
+# CPU pinning for voice latency stability (issue #25). Core 5 hosts the
+# orchestrator and every audio child it spawns (piper, sox, deep-filter,
+# arecord, aplay) — they inherit this affinity. Keeping the TTS pipeline on
+# one core shields whisper (2-3) and llama-server (4) from preemption during
+# decode. arecord runs on core 5 by inheritance (not 0-1 as the issue's
+# system-bucket bullet suggests); revisit with per-spawn taskset only if
+# capture jitter shows up under load.
+CPUAffinity=5
+
 # Needs GPU (Whisper TRT, Piper TTS), audio (ALSA/I2S), and network.
 ProtectSystem=no
 SupplementaryGroups=video render audio
--- a/deploy/systemd/genie-governor.service
+++ b/deploy/systemd/genie-governor.service
@@ -15,6 +15,10 @@ WatchdogSec=30
 MemoryMax=32M
 CPUWeight=50
 
+# Pin to the system bucket (issue #25) — governor only polls memory and writes
+# drop-in configs, keep it off the inference cores (2-5).
+CPUAffinity=0 1
+
 # Security hardening
 ProtectSystem=strict
 ProtectHome=yes
--- a/deploy/systemd/genie-health.service
+++ b/deploy/systemd/genie-health.service
@@ -14,6 +14,10 @@ WatchdogSec=60
 MemoryMax=32M
 CPUWeight=20
 
+# Pin to the system bucket (issue #25) — health monitor is a periodic poller,
+# keep it off the inference cores (2-5).
+CPUAffinity=0 1
+
 # Security hardening
 ProtectSystem=strict
 ProtectHome=yes
--- a/deploy/systemd/genie-llm.service
+++ b/deploy/systemd/genie-llm.service
@@ -32,6 +32,14 @@ Restart=on-failure
 RestartSec=5
 TimeoutStartSec=120
 
+# CPU pinning for voice latency stability (issue #25). Inference is GPU-bound
+# (--n-gpu-layers 999), so one dedicated core is enough to host CUDA dispatch
+# + sampler + tokenizer without contending with whisper (2-3) or genie-core
+# (5). --threads=4 above stays as-is for now; with a single-core affinity the
+# four worker threads time-share core 4 — fine for the GPU-bound critical path
+# but worth revisiting (--threads 1 or 2) if prompt-eval shows scheduler churn.
+CPUAffinity=4
+
 # GPU needs full system access
 ProtectSystem=no
 SupplementaryGroups=video render
--- a/deploy/systemd/genie-mqtt.service
+++ b/deploy/systemd/genie-mqtt.service
@@ -9,6 +9,10 @@ Restart=always
 RestartSec=3
 MemoryMax=16M
 
+# Pin to the system bucket (issue #25) — broker traffic is light and event-
+# driven, keep it off the inference cores (2-5).
+CPUAffinity=0 1
+
 ProtectSystem=strict
 ProtectHome=yes
 ReadWritePaths=/opt/geniepod/data/mqtt
--- a/deploy/systemd/genie-wakeword.service
+++ b/deploy/systemd/genie-wakeword.service
@@ -10,8 +10,12 @@ ExecStart=/usr/bin/env python3 /opt/geniepod/bin/genie-wakeword.py
 Restart=always
 RestartSec=1
 
-# Pin to isolated CPU cores for real-time audio processing.
-CPUAffinity=4 5
+# Pin to audio/system cores for real-time wake-word detection (issue #25).
+# Cores 4 and 5 are now reserved for llama-server and genie-core respectively;
+# the wake-word listener belongs alongside ALSA + capture on the system bucket.
+# SCHED_FIFO at priority 50 is kept so the continuous audio loop isn't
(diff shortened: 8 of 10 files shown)


ai-hpc (commented):
_(duplicate review — see the full review just below.)_

ai-hpc (commented):
Design is sound and the doc page is exactly the right shape — bucket map, verification commands, acceptance signal, and an honest "we stopped short of `isolcpus=` because that needs a Jetson reflash" note. The keep-SCHED_FIFO-but-move-affinity-to-0-1 treatment of `genie-wakeword` is also the right call (the previous `4 5` clashes with this PR's own llama / core assignment). I'd be happy to land this in spirit.

Three things blocking merge as-is, in rough order of severity:

### 1. The pinning targets `genie-llm.service`, but the default LLM unit on `main` is now `genie-ai-runtime.service`

Between when this PR was opened (2026-05-17) and today, the default LLM backend got flipped:

- #55 — flipped `[services.llm].backend` default to `genie_ai_runtime`, made `genie-ai-runtime.service` the default unit.
- #56 — landed the `genie-ai-runtime` v1.0 build/install pipeline.
- #70 — removed `ExecStartPre=drop_caches` from `genie-ai-runtime.service` to preserve warm Qwen3-4B page-cache across restart.
- #76 — added `Before=genie-whisper.service genie-whisper-warmup.service homeassistant.service genie-core.service` to `genie-ai-runtime.service`, plus `--int8-kv -c 8192` and `Environment=GENIEPOD_AI_RUNTIME_CONTEXT=8192`.

This PR pins `genie-llm.service` to `CPUAffinity=4`, which is correct for an operator on the llama.cpp fallback path. But on the default alpha.9 stack, `genie-llm.service` is masked / not running — the active LLM unit is `genie-ai-runtime.service`, which this PR leaves unpinned. So the central acceptance criterion of #25 (LLM on dedicated core 4, no contention with whisper / core) doesn't actually hold on the default deploy.

Fix: add a `CPUAffinity=4` line to `deploy/systemd/genie-ai-runtime.service` with the same justification comment, parallel to the one this PR adds to `genie-llm.service`. The two units already `Conflicts=` each other so only one will ever be active at a time; the affinity setting on the inactive one is a no-op.

### 2. PR body contains the Claude Code AI-attribution trailer

The trailer at the bottom of the PR body — the robot-emoji line that names the tool — is exactly what #91 just added a `Contribution / PR body checklist` CI rule against. Same project norm that `CONTRIBUTING.md` "Commit hygiene" already calls out for `Co-Authored-By: Claude / Copilot / other AI-assistant trailers`. Using AI tooling to help draft the PR is welcome (and the writeup here is genuinely good); attribution in the PR description and `git log` stays with the human contributor so credit is unambiguous.

The check didn't run on this PR's existing rollup because the workflow uses `pull_request_target` and the PR hasn't been touched since the rule landed — but the next edit / push will fire it as a failure. Fix is one line: strip the footer from the PR body before re-pushing.

### 3. CI hasn't run on this PR

`statusCheckRollup` is empty. The PR pre-dates the CI suite landing on `main` (PR #37 / #49 / #50 / #51 / #62), and the standard `pull_request` workflows source their definitions from the PR's head ref — which doesn't carry them. After #89 switched the contribution-checklist workflow to `pull_request_target`, that one will fire on the next push, but `cargo fmt / clippy / test` and `aarch64 cross-compile` need first-time-contributor approval for fork PRs.

This isn't your problem to fix — but it means the PR can't be verified to compile / clippy-clean / cross-compile until either (a) a maintainer approves the CI run, which we can do once you push the rebased version, or (b) you `git rebase main` to bring the workflow files onto your head branch.

### Two smaller things, neither blocking

- **`genie-core.service` `CPUAffinity=5` interaction with PR #87.** PR #87 changed `genie-core` to use `tokio::task::LocalSet::spawn_local` so HTTP requests multiplex on a single OS thread. Pinning genie-core to one core is consistent with that single-thread design — the chat-vs-health-poll concurrency PR #87 unlocked stays intact. But it's worth a sentence in the comment acknowledging the interaction so a future reader doesn't reflexively bump the thread count without understanding the trade-off.
- **`--threads=4` on llama-server with `CPUAffinity=4` (one core).** Your comment acknowledges this and suggests revisiting `--threads 1 or 2` if prompt-eval shows scheduler churn. That's the right framing — leave the threads at 4 for now and put the measurement burden on whoever sees the issue.

### Suggested path forward

Pick one:

**(a) Refresh and re-push.** Rebase onto `main`, add `CPUAffinity=4` to `deploy/systemd/genie-ai-runtime.service`, strip the `🤖 Generated with [...]` trailer from the PR body. We approve CI, check the 7 jobs, and merge. The bulk of the work (doc page, layout, justification comments) is reusable as-is.

**(b) Close as superseded** if you don't have the bandwidth, and we either re-open with the rebase or someone else picks up #25. Pattern matches what we did with #47 a few days ago — the design here would feed into the follow-up cleanly.

Either is fine. Let me know which way you want to take it, and we'll move from there.

enjoyandlove:
Hi,@ai-hpc
I have just fix this PR. Could you please review it?
Thank you for your time.

ai-hpc (changes requested):
Thanks for the revision — PR body now has the proof checkboxes ticked and the AI-attribution trailer is gone (point #2 from the earlier review). The added pinning on `genie-api` / `genie-governor` / `genie-health` / `genie-mqtt` / `genie-wakeword` / `genie-whisper` / `genie-core` is all good and lands real value.

One blocker, plus a new issue this revision introduced. Both small.

### Blocker — `deploy/systemd/genie-ai-runtime.service` still has no `CPUAffinity=`

This was the substantive concern from the first review and it's still open. Confirmed against `43fa029` on this PR's head:

```bash
$ gh api 'repos/GeniePod/genie-claw/contents/deploy/systemd/genie-ai-runtime.service?ref=43fa0299' \
    --jq .content | base64 -d | grep -nE '^CPUAffinity'
# (no output)
```

On the alpha.9 default deploy, the active LLM unit is `genie-ai-runtime.service`, not `genie-llm.service`. As this PR stands, the central acceptance criterion from #25 ("LLM on dedicated core 4") only fires when an operator has explicitly flipped `[services.llm].backend = "llama_cpp"`. The default deploy gets no LLM pinning at all — Qwen3 ends up on whichever core the kernel picks, contending with whisper / core / system.

Fix is one stanza on `deploy/systemd/genie-ai-runtime.service`, mirroring what's already on `genie-llm.service`:

```ini
# CPU pinning for voice latency stability (issue #25). Same core 4 as the
# llama.cpp fallback (genie-llm.service), since genie-ai-runtime and llama-
# server are mutually exclusive via Conflicts= and only one is ever active.
# Inference is GPU-bound, so a single dedicated core hosts CUDA dispatch +
# sampler without contending with whisper (2-3) or genie-core (5).
CPUAffinity=4
```

### New issue this revision introduced — doc documents pinning that doesn't exist

The updated `doc/deployment-and-ops.md` now says core 4 is `llama-server / jllm-server (GPU-bound; one core hosts CUDA dispatch + sampler — whichever LLM backend is active)`. That's factually wrong on the merged state of this PR: an operator following the doc and running

```bash
pid=$(systemctl show -p MainPID --value genie-ai-runtime.service)
taskset -pc "$pid"
```

…would see the runtime is NOT on core 4, despite the doc claiming it. Same fix as the blocker — once the unit has `CPUAffinity=4`, the doc becomes accurate.

### CI status

I approved the two `action_required` runs (`CI` and `Cross-compile (aarch64 / Jetson)`) for this PR's head, so the next cycle should give us the full check matrix. Currently only the `PR body checklist` has run on the new commits; the cargo / cross-compile / no-default-features signal needs the queued runs to land before this can merge.

### Suggested next step

One follow-up commit: add the `CPUAffinity=4` stanza to `deploy/systemd/genie-ai-runtime.service`. The rest of this PR is good as-is. Once that lands and CI clears, I'll re-review and merge.

enjoyandlove:
Hi, @ai-hpc 
Could you please review this PR again?

ai-hpc (approved):
Both blockers from the previous review are resolved. Diffing `deploy/systemd/genie-ai-runtime.service` at the new head `7ccf14d`:

```diff
+# CPU pinning for voice latency stability (issue #25). Inference is GPU-bound
+# (jetson-llm-server offloads to Jetson iGPU), so one dedicated core is enough
+# to host CUDA dispatch + sampler + tokenizer without contending with whisper
+# (2-3) or genie-core (5). Mirrors genie-llm.service; only one of the two units
+# runs at a time (Conflicts=genie-llm.service above), so both can claim core 4.
+CPUAffinity=4
```

That's exactly the right shape — including the `Conflicts=genie-llm.service` callout that explains why both LLM units can safely declare `CPUAffinity=4` without stepping on each other. Future readers won't have to puzzle out why two units claim the same core.

With this commit, the doc page's "core 4 — llama-server / jllm-server (whichever LLM backend is active)" line is now accurate on both backend paths. The doc-vs-code mismatch I flagged in the v2 review is closed.

Summary of where the PR sits versus my earlier review points:

1. **PR body trailer / proof checkboxes** (v1 review): fixed in the earlier revision.
2. **`genie-ai-runtime.service` missing `CPUAffinity=`** (v1+v2 review blocker): fixed in commit `7ccf14d`. Doc and code now agree.
3. **CI** (v1+v2 review): I approved the new `CI` and `Cross-compile` runs that were queued on this push. Waiting on them to land green before final merge — the substantive review is done.

Once CI clears on `7ccf14d`, this is ready to land. Approving on the code; the maintainer can squash-merge as soon as the remaining checks turn green.

enjoyandlove:
Hi, @ai-hpc 
I have just addressed requested update.
could you please review this PR again?

ai-hpc:
Merged at `7268bbba418b6134b52c80a40b1263733d5e50ac`.
Thanks to @enjoyandlove!

enjoyandlove:
@ai-hpc 
Thank you for merging this PR.

PR #51, opened 2026-05-17
ci: add shellcheck + ruff workflow for shell / Python scripts (issue #34)

statxc:
> 🧪 **The stretch goal from #34.** Lints every tracked shell script and Python script in the repo. Won't run on Rust-only diffs.

Closes the stretch slice of issue #34. **Independent** — no dependency on the other CI PRs in the series.

## Why a separate workflow?

The shell scripts under `deploy/` are critical to Jetson deployment (`setup-jetson.sh`, `detect-audio-device.sh`, `genie-restart-all.sh`). The Python scripts (`genie-wake-listen.py`, `genie-wakeword.py`) drive the wake-word detection that fronts the entire voice loop. A subtle bug in either is hard to catch by code review alone, and the Rust CI pipeline doesn't cover them. This workflow closes that gap.

## The two jobs

| Job | What it runs |
| :--- | :--- |
| `shellcheck` | `apt-get install shellcheck`, then discover via `git ls-files '*.sh'`, then `shellcheck --severity=warning` over the list |
| `ruff` | `actions/setup-python@v5` (3.12), `pip install 'ruff==0.6.*'`, discover via `git ls-files '*.py'`, then `ruff check --output-format=github` |

Both jobs **exit cleanly when discovery returns no files**, so a future cleanup pass that removes every shell or Python script doesn't break CI.

Triggers are scoped to `**.sh` / `**.py` / the workflow file itself — Rust-only diffs never spin this up.

## Why `ruff.toml` and not stricter ruff defaults?

```toml
target-version = "py310"
line-length = 100

[lint]
ignore = ["E402"]
```

`E402` (*module-level import not at top of file*) is intentionally allowed because the wakeword scripts under `deploy/scripts/` **must** run code before importing PyAudio:

```python
import warnings
warnings.filterwarnings("ignore")     # Silence scipy/numpy version warnings

import os, sys
_stderr_fd = os.dup(2)
_devnull = os.open(os.devnull, os.O_WRONLY)
os.dup2(_devnull, 2)                  # Redirect stderr → /dev/null

import argparse
import numpy as np
import pyaudio                        # ← E402 fires here without the ignore
from scipy.signal import resample
from openwakeword.model import Model

os.dup2(_stderr_fd, 2)                # Restore stderr
```

Without the redirect, ALSA / PortAudio dump C-level diagnostics on stderr that pollute the protocol stdout the scripts use to talk to genie-core. The ordering is load-bearing. The `ruff.toml` comment documents this so a future contributor doesn't "fix" it.

`py310` matches the Python that ships with Jetson Ubuntu 22.04 (the deploy target).

## Local verification

Reproduced both CI jobs on this branch:

- [x] **shellcheck `--severity=warning`** — clean on:
  - `deploy/setup-jetson.sh`
  - `deploy/scripts/detect-audio-device.sh`
  - `deploy/scripts/genie-restart-all.sh`
- [x] **`ruff check`** with the new config — clean on:
  - `deploy/scripts/genie-wake-listen.py`
  - `deploy/scripts/genie-wakeword.py`
- [x] `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/scripts.yml'))"` — valid YAML

---

Refs #34.

### DIFF
--- a/.github/workflows/scripts.yml
+++ b/.github/workflows/scripts.yml
@@ -1,0 +1,68 @@
+name: Scripts
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - "**.sh"
+      - "**.py"
+      - ".github/workflows/scripts.yml"
+  pull_request:
+    branches: [main]
+    paths:
+      - "**.sh"
+      - "**.py"
+      - ".github/workflows/scripts.yml"
+
+concurrency:
+  group: scripts-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  shellcheck:
+    name: shellcheck
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install shellcheck
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends shellcheck
+      - name: Run shellcheck on all .sh files
+        run: |
+          set -euo pipefail
+          mapfile -t files < <(git ls-files '*.sh')
+          if [ "${#files[@]}" -eq 0 ]; then
+            echo "no shell scripts to check"
+            exit 0
+          fi
+          printf '%s\n' "${files[@]}"
+          shellcheck --severity=warning "${files[@]}"
+
+  ruff:
+    name: ruff
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - name: Install ruff
+        run: pip install --no-cache-dir 'ruff==0.6.*'
+      - name: Discover Python files
+        id: discover
+        run: |
+          set -euo pipefail
+          mapfile -t files < <(git ls-files '*.py')
+          if [ "${#files[@]}" -eq 0 ]; then
+            echo "no python files to check"
+            echo "found=false" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+          printf '%s\n' "${files[@]}"
+          echo "found=true" >> "$GITHUB_OUTPUT"
+      - name: ruff check
+        if: steps.discover.outputs.found == 'true'
+        run: ruff check --output-format=github .
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,18 @@
 
 ### Added
 
+- `.github/workflows/scripts.yml` and `ruff.toml` — shellcheck + ruff
+  workflow for the stretch slice of issue #34. Discovers all
+  tracked `*.sh` and `*.py` files via `git ls-files`, then runs
+  `shellcheck --severity=warning` and `ruff check
+  --output-format=github`. `ruff.toml` pins `target-version = "py310"`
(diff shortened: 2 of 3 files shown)


ai-hpc (approved):
Stretch slice of #34 done correctly. Two CI jobs (shellcheck + ruff), path-scoped triggers, exit-cleanly-on-empty-discovery, pinned `ruff==0.6.*` to dodge minor-version breakage, py310 target matching Jetson Ubuntu 22.04. Independent of the other #34 PRs in the series.

## What lands

- `.github/workflows/scripts.yml` — shellcheck + ruff, triggered only on `**.sh` / `**.py` / the workflow file itself so Rust-only diffs skip
- `ruff.toml` — `target-version = "py310"`, `line-length = 100`, single ignore `E402` with an inline comment explaining the wakeword-script ordering constraint
- CHANGELOG entry that carries the E402 rationale inline

## What I read closely

The E402 ignore is the only judgment call in this PR and it's correct. `deploy/scripts/genie-wake-listen.py` and `genie-wakeword.py` do:

```python
import warnings; warnings.filterwarnings("ignore")
import os, sys
_stderr_fd = os.dup(2)
_devnull = os.open(os.devnull, os.O_WRONLY)
os.dup2(_devnull, 2)                  # ALSA/PortAudio stderr → /dev/null
import pyaudio                        # ← E402 fires here without the ignore
...
os.dup2(_stderr_fd, 2)                # restore stderr
```

Without that ordering, ALSA / PortAudio dump C-level diagnostics on stderr and pollute the line-based `WAKE <score>` / `READY` protocol stdout the wakeword scripts use to talk to `voice_loop.rs`. A future contributor "fixing" the E402 to make ruff happy would break the voice loop's wake handshake at the worst possible moment (in the field, in audio, intermittently). The `ruff.toml` comment block is exactly the right inoculation against that.

## Things you got right that are worth naming

- **`git ls-files` discovery** rather than `find` — only walks tracked files, no false positives from build artifacts or `.gitignore`'d vendored code.
- **"No files to check" exit-0 path** in both jobs. A future cleanup that removes every script doesn't break CI.
- **`ruff` step's `if: steps.discover.outputs.found == 'true'` guard** composes correctly with the discover step's `GITHUB_OUTPUT`.
- **`ruff==0.6.*` version pin** — allows patch/minor of 0.6, refuses 0.7. Ruff has breaking lint changes between minor versions; this is the right pinning shape.
- **`shellcheck --severity=warning`** rather than `info`. Pragmatic — catches real bugs without the cosmetic noise.
- **5-minute timeouts** and `concurrency.cancel-in-progress: true` are appropriate.

1. `ruff check . --output-format=github` runs against the whole repo rather than the discovered file list. That's actually correct — ruff's own walker plus `ruff.toml` handle scoping, and `.` is the canonical CLI form. The "discover Python files" step is doing logging/early-exit, not driving the ruff invocation. The two-step shape is fine; just noting the asymmetry.
2. No scheduled run — only push/pull-request. Means dep rot is theoretical for `shellcheck` and `ruff` (both stable, no transitive churn). Acceptable.
3. No `ruff format --check` job, only lint. Right call to keep scope tight; format-check can land later if/when style drift becomes a real concern.

## Verdict

Approving and merging. None of the above is worth holding for.

ai-hpc:
Merged at `5375654f78e16773a33931ecc65c7d29b6446ea3`.
Thanks to @statxc!

Issue #42, opened 2026-05-16
Telegram voice-message support: alternative STT pipeline that bypasses the on-device mic array


ai-hpc:
## Background

The voice pipeline today is wired end-to-end around the on-device microphone:

```
wake word / PTT  →  arecord (ALSA mic)  →  AEC ref + half-duplex gate
                                          ↓
                            DSP / VAD / denoise (sox / DeepFilterNet)
                                          ↓
                            Whisper STT (CLI or :8178 server)
                                          ↓
                            intent gate + speaker identity + language
                                          ↓
                            memory inject + LLM stream
                                          ↓
                            Piper TTS  →  aplay
```

That pipeline is correct for the Jetson + LyraT / USB-headset deployment, but every stage upstream of Whisper assumes a physical mic on the box. For Telegram users — who already record audio on their phone before sending — most of that pipeline is irrelevant:

- No wake word: the act of sending a voice message *is* the wake.
- No AEC, no half-duplex gate: Telegram is a turn-based channel; the bot never hears itself.
- No on-device DSP: the user's phone has already done AGC, noise gate, and Opus encoding.
- No ALSA: there is no mic on the device the bot runs on (the bot can run on a chat-only x86 server or in Docker with no audio hardware at all).

This issue proposes a **second voice pipeline** — file-driven, Telegram-sourced — that reuses the STT and LLM-orchestration layers genie-claw already has, and bypasses the on-device-mic stages that don't apply.

## Goal

A Telegram user can send a voice message to the bot, and the bot:

1. downloads the audio from Telegram,
2. transcribes it via the existing Whisper engine,
3. runs the same memory / tool / LLM agent loop the text and on-device voice paths already use,
4. replies in Telegram (text in phase 1; optionally a voice message in phase 2).

Crucially: this works on a **chat-only build** (per #41 — no `voice` feature required if we land a smaller `voice-stt` sub-feature; see "Feature gating" below) and on a Jetson with no LyraT / USB mic attached.

## Proposed pipeline

```
Telegram voice msg
   │  voice: { file_id, duration, mime_type: "audio/ogg" }
   ▼
Telegram Bot API getFile  →  HTTPS GET on file_path
   │
   ▼
.ogg / .oga in /tmp
   │
   ▼  ffmpeg -i <ogg> -ar 16000 -ac 1 -f wav <wav>
   │  (or rust opus decoder + WAV writer; ffmpeg is simpler and matches existing
[...]

Issue #41, opened 2026-05-16
Make voice pipeline an optional Cargo feature (gated subsystem)


ai-hpc:
## Background

The voice subsystem is currently **unconditionally compiled into `genie-core`** — every binary built from this repo carries the wake-word loop, push-to-talk loop, Whisper/Piper subprocess machinery, AEC, VAD, DSP, identity, intent, language, streaming, format, and pipeline modules. That's roughly **~6,000 LoC across 12 submodules plus `voice_loop.rs`**:

```
crates/genie-core/src/voice/aec.rs        402 lines
crates/genie-core/src/voice/dsp.rs        238
crates/genie-core/src/voice/format.rs     301
crates/genie-core/src/voice/identity.rs   836
crates/genie-core/src/voice/intent.rs     221
crates/genie-core/src/voice/language.rs   190
crates/genie-core/src/voice/mod.rs        333
crates/genie-core/src/voice/noise.rs      355
crates/genie-core/src/voice/pipeline.rs   257
crates/genie-core/src/voice/streaming.rs  158
crates/genie-core/src/voice/stt.rs        959
crates/genie-core/src/voice/tts.rs        493
crates/genie-core/src/voice/vad.rs        120
crates/genie-core/src/voice_loop.rs      1149
                                         -----
                                        ~6,012
```

That code is paid for whether or not it's used. Runtime selection (`voice_enabled` config / `--voice` CLI flag / `GENIEPOD_VOICE=1` env) controls whether the loop *runs*, not whether it *exists in the binary*.

Three real deployment shapes that would benefit from a compile-time gate:

1. **Chat-only Telegram / Slack / Discord deployments.** No microphone, no speakers. Currently ship dead code.
2. **Cross-platform dev (macOS, Windows, non-Jetson Linux).** The voice loop shells out to `arecord` / `aplay` (ALSA, Linux-only) — on other platforms the code compiles but every voice path errors at runtime. A compile-time gate makes the platform-incompatibility explicit at build time.
3. **CI iteration speed.** Workflows from #34 (`ci.yml`, `cross.yml`) currently compile the voice subsystem on every PR even when nothing voice-adjacent changed. A `--no-default-features` build path in the CI matrix would also catch dead-code rot in the voice gate itself.

## Goal

Make the voice pipeline a **Cargo feature** in `genie-core`, **on by default**, that consumers can opt out of with `--no-default-features` or by re-enabling other features explicitly.

[...]

PR #53, opened 2026-05-17
feat(telegram): voice-message ingestion via Whisper (issue #42 phase 1)

bittoby:
Part of #42 (phase 1 only).

A Telegram user can send a voice message; the bot downloads it, transcodes
OGG/Opus → 16 kHz mono WAV with ffmpeg, transcribes via the same Whisper
server (`/inference`) or `whisper-cli` the on-device voice loop uses, then
routes the transcript through the existing `/api/chat` flow and replies as text.

## Design
- File-driven STT path; no `voice/*` runtime modules pulled in → works on
  chat-only deployments without ALSA.
- Hallucination filter + duration cap + RAII temp-file cleanup.
- Concurrent voice messages get unique `/tmp` paths (pid + nanos nonce).
- New config block `[telegram.voice]`; off by default.

## Test plan
- [x] `cargo check --workspace --all-targets` : clean.
- [x] `cargo test -p genie-core telegram::` : 6/6 passed.
- [x] `cargo test --workspace` — 388 passed; 2 pre-existing `memory::tests::promotion_*`
      flakes (parallel-mode SQLite/temp-path collision; pass with `--test-threads=1`).
- [ ] Manual end-to-end against a real Telegram bot + Jetson whisper-server.
- [ ] Integration test with mock Telegram API + ffmpeg fixture — deferred to follow-up.

## Deferred (phase 2 / 3 per issue)
- Piper TTS reply via `sendVoice`.
- Speaker identity from Telegram user_id.
- Group-chat voice handling.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -643,6 +643,22 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f8ca58f447f06ed17d5fc4043ce1b10dd205e060fb3ce5b979b8ed8e59ff3f79"
 
 [[package]]
+name = "mime"
+version = "0.3.17"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6877bb514081ee2a7ff5ef9de3281f14a4dd4bceac4c09388074a6b5df8a139a"
+
+[[package]]
+name = "mime_guess"
+version = "2.0.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f7c44f8e672c00fe5308fa235f821cb4198414e1c77935c1ab6948d3fd78550e"
+dependencies = [
+ "mime",
+ "unicase",
+]
+
+[[package]]
 name = "mio"
 version = "1.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -838,6 +854,7 @@ dependencies = [
  "base64",
  "bytes",
  "futures-core",
+ "futures-util",
  "http",
  "http-body",
  "http-body-util",
@@ -846,6 +863,7 @@ dependencies = [
  "hyper-util",
  "js-sys",
  "log",
+ "mime_guess",
  "percent-encoding",
  "pin-project-lite",
  "quinn",
@@ -1358,6 +1376,12 @@ name = "try-lock"
 version = "0.2.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e421abadd41a4225275504ea4d6566923418b7f05506fbc9c0fe86ba7396114b"
+
+[[package]]
+name = "unicase"
+version = "2.9.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dbc4bc3a9f746d862c45cb89d705aa10f187bb96c76001afab07a0d35ce60142"
 
 [[package]]
 name = "unicode-ident"
--- a/crates/genie-common/src/config.rs
+++ b/crates/genie-common/src/config.rs
@@ -448,6 +448,43 @@ pub struct TelegramConfig {
     /// Bypass the allowlist and accept messages from any chat.
     #[serde(default)]
     pub allow_all_chats: bool,
+
+    /// Voice-message handling for the Telegram channel (issue #42).
+    #[serde(default)]
+    pub voice: TelegramVoiceConfig,
+}
+
+#[derive(Debug, Deserialize, Clone)]
+pub struct TelegramVoiceConfig {
+    /// Enable voice-message ingestion. When false, voice messages get a polite
+    /// text reply explaining that voice is not enabled on this deployment.
+    #[serde(default)]
+    pub enabled: bool,
+
+    /// Hard cap on accepted voice duration. Telegram includes a `duration`
+    /// field; anything longer is rejected before download.
+    #[serde(default = "defaults::telegram_voice_max_duration_secs")]
+    pub max_voice_duration_secs: u32,
+
+    /// Delete the downloaded `.ogg` and transcoded `.wav` after handling.
+    #[serde(default = "defaults::telegram_voice_delete_temp_audio")]
+    pub delete_temp_audio: bool,
+
+    /// Path to the `ffmpeg` binary used to transcode Telegram OGG/Opus to the
+    /// 16 kHz mono WAV that Whisper consumes.
+    #[serde(default = "defaults::telegram_voice_ffmpeg_path")]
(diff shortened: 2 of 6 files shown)


bittoby:
Hi @ai-hpc Could you plz review my first PR for this repo? Welcome to any feedback. Thanks.

ai-hpc (commented):
Phase 1 of #42 delivered end-to-end. Read every hunk; the architecture matches what the issue asked for and the tradeoffs are correctly justified in code comments. One design question worth a one-sentence author response, otherwise approve-quality.

## What lands and what's good

**The right architectural call — and you flagged it explicitly:**

```rust
//! The Telegram adapter stays out of process boundaries with `voice/*`
//! modules — it speaks to Whisper directly via the same HTTP server or
//! `whisper-cli` subprocess the on-device voice loop uses, so a chat-only
//! deployment without ALSA still gets voice-in.
```

Issue #42 specifically asked for "works on a chat-only build (per #41 — no `voice` feature required if we land a smaller `voice-stt` sub-feature)." You chose to *re-implement* the minimal STT subset (whisper-server multipart, whisper-cli subprocess, hallucination filter, language normalization) inline rather than share code with `voice/stt.rs`. That's the right tradeoff for #41 compatibility — when voice becomes a Cargo feature, this adapter still works on `--no-default-features`. The duplication is documented in the comment:

```rust
// Lives parallel to `voice::stt::SttEngine::transcribe_via_server`
// rather than reusing it directly so the Telegram adapter stays
// callable from chat-only builds where the `voice` feature is off.
```

That comment alone is doing real maintenance-debt prevention work. Future contributors won't "refactor" the duplication out and accidentally couple Telegram to `voice/*`.

**RAII `TempCleanup`** is correct — `Drop` runs on success, error, and panic during unwind. Honors `delete_temp_audio = false` for live debugging. `_ = std::fs::remove_file(...)` correctly swallows the "file may not exist if download failed before write" case.

**Concurrency via `pid + subsec_nanos`** matches the pattern from `server.rs`'s `health_endpoint_reports_llm_backend` (PR #39). Race window is two voice messages arriving within the same nanosecond — astronomically unlikely on Telegram polling cadence.

**Server-vs-CLI selection logic** mirrors `voice_loop.rs:132-147`. Operator's existing whisper config carries over to Telegram without duplication or per-channel knobs.

**User-facing error messages are distinct per failure mode:**

- voice not enabled → "Voice messages aren't enabled on this deployment."
- too long → "Voice message is too long (Xs); the limit is Ys."
- download failed → "Sorry, I couldn't download that voice message from Telegram."
- ffmpeg failed → "Sorry, I couldn't decode that voice message (ffmpeg failed)."
- transcribe failed → "Sorry, I couldn't transcribe that voice message."
- empty / hallucination → "I couldn't make out any speech in that voice message."

Right shape — Telegram users get actionable feedback per failure class, debugging info stays in `tracing::warn!` logs.

**Config defaults are honest.** `#[serde(default)]` on the whole `voice` block, `enabled: false` by default. Existing Telegram deployments don't change behavior on upgrade.

**Tests cover the right surface:**

- `configured_language_normalizes_auto_and_blank` — boundary cases for the auto/empty hint
- `clean_transcript_drops_whisper_hallucinations` — four canonical hallucination strings plus a positive case
- `telegram_voice_message_deserializes` — wire-format contract that asserts `voice = Some, audio = None, text = None` simultaneously. That's the contract a future refactor of the `TelegramMessage` struct could silently break; the test pins it.

**Dep add is minimal:** one reqwest feature (`multipart`) → three transitive crates (`mime`, `mime_guess`, `unicase`). Cargo.lock diff is tight.

## The one design question worth a one-sentence response

```rust
if let Some(voice) = message.voice.as_ref().or(message.audio.as_ref()) {
    return self.handle_voice_message(chat_id, voice).await;
}
```

Telegram's `voice` and `audio` message types are genuinely different:

| Type | What it is | Size limit | Format |
|---|---|---|---|
| `voice` | Recorded with the in-app voice button | ~1 MB | OGG/Opus, mono, narrow channel |
| `audio` | A music / audio file the user attached | up to 50 MB | MP3, AAC, FLAC, etc., often stereo |

The PR treats them identically. ffmpeg handles arbitrary formats so the transcode works. `max_voice_duration_secs = 60` clips the abuse vector. But a user sending a 30-second song clip will get the lyrics transcribed and routed to the LLM as a chat message — probably not what either side intends.

Two reasonable answers, either fine:

- **Drop `audio` for phase 1.** Only handle `voice`. Simpler contract; voice button is the obvious UX. The issue spec actually only mentions "voice messages."
- **Keep both, but add a sub-knob** like `accept_audio_messages = false` (default off) for the rare case where someone really wants attached audio files transcribed.

Either is a one-line change. I'd lean toward the first (drop `audio`) because it matches the issue's intent and removes a category of "did the user really want this?" ambiguity.

## Smaller observations (not blockers)

1. **`clean_transcript` filter doesn't normalize internal whitespace.** `"Thank   you."` (double space) wouldn't match `"thank you."` after lowercase. Robustness gap; rare in Whisper output.
2. **`"you"` is in the hallucinations list.** A user actually saying "you" as a single-word utterance gets silently dropped. Mirrors the on-device behavior so consistency wins; just noting.
3. **No bot-mention strip in voice path.** Voice messages don't carry `@mentions` so this isn't a bug, just asymmetric with the text path.
4. **`from_user` is dropped on the floor.** Phase 3 (Telegram-user-id → speaker identity) is where it'd plug in. Correctly out of scope.

## The real validation gap

Your test plan flags it honestly:

> - [ ] Manual end-to-end against a real Telegram bot + Jetson whisper-server.
> - [ ] Integration test with mock Telegram API + ffmpeg fixture — deferred to follow-up.

Pure-function unit tests are excellent. The end-to-end on a real Jetson with real Telegram traffic is the meaningful validation that nobody's done yet. The two pre-existing `memory::tests::promotion_*` flakes you mentioned are a known issue (same flake came up in PR #35's test report) — not your problem.

## Verdict

Substantively approve-quality. If the maintainer wants to merge as-is, the only thing I'd ask first is a one-sentence call on the `voice`/`audio` conflation — either drop `audio` (one-line change) or document the intent (the comment in `handle_voice_message` could note "intentionally treats `audio` attachments identically to `voice` recordings"). Otherwise: ship it.

Recommend the manual Jetson end-to-end happen before / shortly after merge, on a Jetson that already has `whisper-server` running. The download → ffmpeg → whisper round-trip on real OGG/Opus traffic is what nobody has actually executed yet.

ai-hpc:
Merged at `51eab82a6ee812846371e80a35e8fd9306ccad41`.
Thanks to @bittoby!

PR #57, opened 2026-05-17
feat: gate voice subsystem behind 'voice' Cargo feature (issue #41)

### LINKED ISSUE #41: Make voice pipeline an optional Cargo feature (gated subsystem)

statxc:
Wraps the STT/TTS/AEC/wakeword pipeline in a default-on `voice` feature on both `genie-core` and `genie-ctl` so `cargo build` produces today's Jetson binary byte-equivalent, while `cargo build --no-default-features` produces a chat-only binary that:

  * skips ~6 kLoC of voice / ALSA-coupled code (4.8 MB -> 4.6 MB release on x86_64-linux),
  * unblocks macOS / Windows hosts which previously couldn't compile the ALSA-shelling voice modules,
  * shaves the CI loop for chat-only PRs.

genie-core
  * `voice = []` added to `[features].default` in `Cargo.toml` next to the existing `telegram`, `llama-cpp`, `genie-ai-runtime` feature surface.
  * `pub mod voice;` and `pub mod voice_loop;` in `lib.rs` are now `#[cfg(feature = "voice")]`.
  * `main.rs` voice-mode branch is gated. When voice is requested on a chat-only build (`--voice` flag / `GENIEPOD_VOICE=1` env / `core.voice_enabled = true` in `geniepod.toml`), the runtime logs one warning and falls through to the existing chat / HTTP path — so an unchanged voice-tagged config deploys cleanly per the issue's acceptance criterion #5.

genie-ctl
  * Mirror `voice` feature, default-on, that forwards to `genie-core/voice`. The `genie-core` path dep is switched to `default-features = false` so a chat-only `genie-ctl` build actually propagates to a chat-only `genie-core` instead of pulling voice back in via the dep tree.
  * The `speaker` subcommand, its help block in `print_usage` / `print_speaker_usage`, every `cmd_speaker_*` helper, `SpeakerCliOptions` + `parse_speaker_options` + `default_speaker_options`, `record_speaker_wav`, and the two `parse_speaker_options` unit tests are `#[cfg(feature = "voice")]`. Invoking `genie-ctl speaker …` on a chat-only build exits with a clear "rebuild with --features voice" message instead of crashing.
  * The `std::process::Command` import is `#[cfg(feature = "voice")]` because the only `Command::new` site lives in `record_speaker_wav`.

Knock-on dead-code cleanup (so `cargo clippy -- -D warnings` is green on `--no-default-features` too)
  * `local_http_host` and its two unit tests in `genie-core/src/main.rs` are now `#[cfg(feature = "telegram")]` — they are only used by the Telegram adapter and were latent dead code on no-telegram builds.

Verified locally on `rustc 1.95`:

  * `cargo fmt --all -- --check` clean
  * `cargo clippy --workspace --all-targets --locked -- -D warnings` clean
  * `cargo clippy -p genie-core -p genie-ctl --no-default-features --all-targets --locked -- -D warnings` clean
  * `cargo test --workspace --locked --all-targets` — 424 / 424 pass
  * `cargo test -p genie-core -p genie-ctl --no-default-features --locked --all-targets` — 353 / 353 pass (delta is the 67 voice + 2 telegram + 2 speaker tests now gated out)
  * `cargo test --workspace --locked --doc` clean
  * `cargo build --release -p genie-core` 4.8 MB, `cargo build --release -p genie-core --no-default-features` 4.6 MB

Out of scope (called out in CHANGELOG): wiring the `--no-default-features` axis into `.github/workflows/ci.yml` (acceptance criterion #8) — `.github/` doesn't exist on `main` yet, so this is a one-line follow-up after issue

Closes #41.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,41 @@
 
 ### Added
 
+- `voice` Cargo feature on `genie-core` and `genie-ctl` (issue #41).
+  Default-on so `cargo build` produces today's Jetson-targeted binary
+  unchanged. `cargo build -p genie-core --no-default-features` (and the
+  matching `-p genie-ctl`) produces a chat-only binary that drops the
+  STT/TTS/AEC/wakeword pipeline, the `VoiceOrchestrator`, the
+  `voice_loop::run` dispatcher, and `genie-ctl`'s `speaker` subcommand:
+    - `pub mod voice;` and `pub mod voice_loop;` in
+      `crates/genie-core/src/lib.rs` are now `#[cfg(feature = "voice")]`.
+    - The voice-mode branch in `crates/genie-core/src/main.rs` is gated;
+      when voice is requested (`--voice` / `GENIEPOD_VOICE=1` /
+      `core.voice_enabled = true`) on a chat-only build, the runtime
+      logs one warning and falls through to the existing chat / HTTP
+      path so deploying an unchanged `geniepod.toml` is a non-event.
+    - `genie-ctl`'s `genie-core` dependency uses
+      `default-features = false`; `genie-ctl`'s own new `voice` feature
+      forwards to `genie-core/voice`. The `speaker` subcommand,
+      `speaker` help line, all `cmd_speaker_*` helpers, and the two
+      `parse_speaker_options` unit tests are `#[cfg(feature = "voice")]`.
+      Invoking `genie-ctl speaker …` on a chat-only build exits with a
+      clear "rebuild with --features voice" message instead of crashing.
+  Knock-on cleanups so `cargo clippy -- -D warnings` is green on both
+  variants: `local_http_host` and its two unit tests in
+  `crates/genie-core/src/main.rs` are now `#[cfg(feature = "telegram")]`
+  (they are only used by the Telegram adapter and were latent dead code
+  on no-telegram builds); the `std::process::Command` import in
+  `crates/genie-ctl/src/main.rs` is `#[cfg(feature = "voice")]` because
+  the only `std::process::Command::new` call lives in `record_speaker_wav`.
+  Release binary on x86_64-linux drops from 4.8 MB to 4.6 MB without
+  voice; the bigger payoff is unblocking macOS / Windows hosts that
+  previously could not compile the ALSA-coupled voice modules.
+  CI matrix coverage (acceptance criterion #8) will be added on top of
+  issue #34's `ci.yml` workflow once that lands; the
+  `cargo build / clippy / test` invocations to add are
+  `-p genie-core -p genie-ctl --no-default-features` next to the
+  existing `--workspace` ones.
 - `.github/workflows/scripts.yml` and `ruff.toml` — shellcheck + ruff
   workflow for the stretch slice of issue #34. Discovers all
   tracked `*.sh` and `*.py` files via `git ls-files`, then runs
--- a/crates/genie-core/Cargo.toml
+++ b/crates/genie-core/Cargo.toml
@@ -14,7 +14,7 @@ name = "genie-core"
 path = "src/main.rs"
 
 [features]
-default = ["telegram", "llama-cpp", "genie-ai-runtime"]
+default = ["telegram", "llama-cpp", "genie-ai-runtime", "voice"]
 telegram = []
 # LLM backend feature surface (declared per issue #32 closeout).
 # Both default-on so today's builds are byte-identical.
@@ -23,6 +23,14 @@ telegram = []
 # regression is caught automatically rather than silently bit-rotting.
 llama-cpp = []
 genie-ai-runtime = []
+# Voice subsystem (issue #41). Gates the STT/TTS/AEC/wakeword pipeline,
+# the `VoiceOrchestrator`, and the `voice_loop::run` dispatcher. Default-on
+# so today's Jetson builds are byte-identical; `--no-default-features`
+# (or `--no-default-features --features telegram,llama-cpp,...`) produces
+# a chat-only binary that builds on macOS / Windows hosts which don't
+# have ALSA. A voice-tagged `geniepod.toml` still deploys cleanly: the
+# runtime logs one warning and falls through to the chat path.
+voice = []
 
 [dependencies]
 genie-common = { workspace = true }
--- a/crates/genie-core/src/lib.rs
+++ b/crates/genie-core/src/lib.rs
@@ -24,7 +24,7 @@
 //! | [`connectivity`] | Boundary for an ESP32-C6 UART Thread/Matter coprocessor |
 //! | [`context`] | LLM context window management with summarization |
 //! | [`prompt`] | Model-aware system prompt builder (6 LLM families) |
-//! | [`voice`] | STT/TTS subprocess management + voice output formatter |
+//! | [`voice`] | STT/TTS subprocess management + voice output formatter (feature `voice`, default-on) |
 //! | [`ota`] | OTA update checker via GitHub Releases |
 //! | [`server`] | HTTP chat API server |
 //!
@@ -58,7 +58,9 @@ pub mod skills;
 #[cfg(feature = "telegram")]
 pub mod telegram;
 pub mod tools;
+#[cfg(feature = "voice")]
 pub mod voice;
+#[cfg(feature = "voice")]
 pub mod voice_loop;
 
 // Re-export key types at crate root for convenience.
--- a/crates/genie-core/src/main.rs
+++ b/crates/genie-core/src/main.rs
@@ -167,53 +167,80 @@ async fn main() -> Result<()> {
     }
 
     // Check for voice mode: --voice flag or GENIEPOD_VOICE=1 or config voice_enabled.
-    let voice_mode = std::env::args().any(|a| a == "--voice")
+    let voice_requested = std::env::args().any(|a| a == "--voice")
         || std::env::var("GENIEPOD_VOICE").unwrap_or_default() == "1"
         || config.core.voice_enabled;
 
+    // Whether the running binary actually has the voice subsystem compiled in.
+    // When voice was requested but the binary is chat-only (issue #41 feature
+    // gate), emit one warning and fall through to the chat path so an existing
+    // voice-tagged `geniepod.toml` still deploys cleanly on a chat-only build.
+    #[cfg(feature = "voice")]
+    let voice_mode = voice_requested;
+    #[cfg(not(feature = "voice"))]
+    let voice_mode = {
+        if voice_requested {
+            tracing::warn!(
+                "voice mode requested (--voice / GENIEPOD_VOICE=1 / core.voice_enabled) but \
+                 this genie-core build was compiled without the 'voice' feature; falling back \
+                 to chat-only mode. Rebuild with default features (or --features voice) to \
+                 enable the voice loop."
+            );
+        }
+        false
+    };
+
     if voice_mode {
-        tracing::info!("voice mode — starting voice interaction loop");
-        let voice_cfg = genie_core::voice_loop::VoiceConfig {
-            whisper_model: config.core.whisper_model.to_string_lossy().to_string(),
-            whisper_cli_path: config.core.whisper_cli_path.to_string_lossy().to_string(),
-            whisper_port: config.core.whisper_port,
-            piper_model: config.core.piper_model.to_string_lossy().to_string(),
-            piper_path: config.core.piper_path.to_string_lossy().to_string(),
-            piper_pipe_mode: config.core.piper_pipe_mode,
-            stt_language: config.core.stt_language.clone(),
-            voice_tts_models: config
-                .core
-                .voice_tts_models
-                .iter()
-                .map(|(language, path)| (language.clone(), path.to_string_lossy().to_string()))
-                .collect(),
-            audio_device: config.core.audio_device.clone(),
-            audio_output_device: config.core.audio_output_device.clone(),
-            sample_rate: config.core.audio_sample_rate,
-            audio_denoiser: config.core.audio_denoiser.clone(),
-            deep_filter_path: config.core.deep_filter_path.to_string_lossy().to_string(),
-            deep_filter_atten_lim_db: config.core.deep_filter_atten_lim_db,
-            post_tts_silence_ms: config.core.post_tts_silence_ms,
-            record_secs: config.core.voice_record_secs,
-            llm_model_path: config.core.llm_model_path.to_string_lossy().to_string(),
-            wakeword_script: config.core.wakeword_script.to_string_lossy().to_string(),
-            voice_continuous: config.core.voice_continuous,
-            voice_continuous_secs: config.core.voice_continuous_secs,
-            speaker_identity: genie_core::voice::identity::SpeakerIdentityProvider::from_config(
-                &config.core.speaker_identity,
-            ),
-        };
-        genie_core::voice_loop::run(
-            voice_cfg,
-            &llm,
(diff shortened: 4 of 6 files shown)


ai-hpc (approved):
Textbook #41 implementation. Read every gate; the architecture is right and the verification regime is exactly what the issue's acceptance criteria called for.

## What's right that's worth naming

### 1. The `default-features = false` on the genie-ctl→genie-core dep

```toml
[dependencies]
genie-core = { path = "../genie-core", default-features = false }
```

This is the subtle correctness detail that's easy to miss. Without it, genie-ctl's `voice = ["genie-core/voice"]` forwarding does nothing — `genie-core`'s `voice` feature stays on via the default-features pull, and turning off `genie-ctl`'s `voice` feature has no effect on what's compiled into `genie-core`. The dep-side `default-features = false` makes the feature gate actually load-bearing.

### 2. The double-cfg-with-unreachable pattern in main.rs

```rust
#[cfg(feature = "voice")] let voice_mode = voice_requested;
#[cfg(not(feature = "voice"))] let voice_mode = { /* warn + false */ };

if voice_mode {
    #[cfg(feature = "voice")]   { genie_core::voice_loop::run(...).await }
    #[cfg(not(feature = "voice"))] { unreachable!(...) }
}
```

`voice_mode` is forced false in the gated-off case so the inner `not(feature)` arm is unreachable at runtime, but the compiler can't always prove that across cfg boundaries — so the `unreachable!()` is the canonical pattern to satisfy type-checking when the conditional `.await` call is absent. Slightly ugly visually, structurally correct.

### 3. Warning text covers all three voice-trigger mechanisms

> "voice mode requested (`--voice` / `GENIEPOD_VOICE=1` / `core.voice_enabled`) but this genie-core build was compiled without the 'voice' feature; falling back to chat-only mode. Rebuild with default features (or `--features voice`) to enable the voice loop."

Matches #41 acceptance criterion #5 ("a voice-tagged config still deploys cleanly … emits one `tracing::warn!` line at startup if `voice_enabled = true` is set but the feature is off") precisely. Names all three trigger mechanisms so the operator knows which one to disable.

### 4. `print_usage` dynamically omits the `speaker` line

```rust
#[cfg(feature = "voice")]
const SPEAKER_HELP_LINE: &str = "    speaker <SUBCOMMAND>\n...";
#[cfg(not(feature = "voice"))]
const SPEAKER_HELP_LINE: &str = "";
```

`genie-ctl --help` on a chat-only build doesn't list `speaker` — the help text matches what the binary can actually do. Right UX.

### 5. The friendly error on `genie-ctl speaker ...` on a chat-only build

> "speaker subcommand is unavailable: this genie-ctl build was compiled without the 'voice' feature (issue #41). Rebuild with default features (or --features voice) to manage local speaker profiles."

References the issue number for traceability. Same shape as the main.rs warning. Consistent.

### 6. Knock-on `local_http_host` gating

`local_http_host` and its two tests were latent dead code on `--no-features telegram`. Now `#[cfg(feature = "telegram")]`. Test module re-gated to `#[cfg(all(test, feature = "telegram"))]`. This is exactly the kind of clippy-warning-on-no-features-build that #41 acceptance criterion #7 implied without naming. Author found and fixed it preemptively.

### 7. Verification regime matches the issue's acceptance criteria

The PR body lists:

- `cargo fmt --all -- --check` clean ← criterion #1 / #2 baseline
- `cargo clippy --workspace --all-targets --locked -- -D warnings` clean (default features)
- `cargo clippy -p genie-core -p genie-ctl --no-default-features --all-targets --locked -- -D warnings` clean ← criterion #4 (the no-default-features path doesn't bit-rot)
- `cargo test --workspace --locked --all-targets` — 424/424
- `cargo test --no-default-features` — 353/353 (delta = 67 voice + 2 telegram + 2 speaker tests gated out)
- Default binary 4.8 MB, no-default-features 4.6 MB

That delta accounting is internally consistent. The 200 KB / 4% binary size reduction is modest — the qualitative win is the macOS / Windows compile unblock the CHANGELOG names explicitly.

## Out-of-scope acknowledgment is honest

> CI matrix coverage (acceptance criterion #8) will be added on top of issue #34's `ci.yml` workflow once that lands

Correct deferral. `ci.yml` doesn't exist on `main` yet (PR #37 was contested over scope-mixing). When `ci.yml` lands, adding `-p genie-core -p genie-ctl --no-default-features` to the matrix is a one-line follow-up. The hook is documented in the CHANGELOG entry so it won't get forgotten.

1. **The test module re-gating** uses `#[cfg(all(test, feature = "telegram"))]` at the `mod tests` level. If a future contributor adds a non-telegram test to that module, they'll wonder why it doesn't run on no-telegram builds. Per-test gates would be slightly cleaner, but there's only one fn (`local_http_host`) being tested today, so module-level is fine.

2. **The unreachable! is technically reachable in malformed builds.** If a future contributor introduces a code path that sets `voice_mode = true` independent of `voice_requested`, the unreachable could fire. But it's defensive — the comment immediately above (`Unreachable: voice_mode is forced false above when the feature is off`) preserves the invariant.

3. **`voice = []` has no transitive cargo deps.** Pure code gating — none of voice's Rust deps (`tokio`, `anyhow`, etc.) are exclusive to it. The subprocess shell-outs (whisper-cli, piper, ffmpeg, arecord, aplay) are runtime, not build-time. So there's nothing to drop at the dep level. Correct that this PR doesn't try to.

## Verdict

Approving. Cleanest implementation of an architectural-gate issue I've reviewed in this batch — every acceptance criterion from #41 is addressed except #8, which is correctly deferred to when `ci.yml` exists on main.

If you're going to merge it: with the `default = ["telegram", "llama-cpp", "genie-ai-runtime", "voice"]` order, every today-time `cargo build` produces a byte-identical Jetson binary. The `--no-default-features` axis adds optionality without changing the default surface. Risk is contained to the new chat-only build path that nobody is currently using.

Recommend a fast-follow PR (once `ci.yml` lands) that adds the matrix axis. The exact invocations to drop in are in the PR body.

ai-hpc:
Merged at `14d149b7eab8deaed875e72ec310bd82f8e251cb`.
Thanks to @statxc!

Issue #19, opened 2026-05-12
Print first-voice-reply latency banner at startup


ai-hpc:
## Motivation

We just shipped two warmup services (`genie-llm-warmup`, #7; `genie-whisper-warmup`, #17) specifically to reduce first-voice-cycle latency. Right now the only way to know whether they actually worked is to read raw `[voice] (STT: X ms)` / `(LLM+TTS: Y ms)` / `(Total cycle: Z ms)` lines scattered across the log. There is no aggregated, easy-to-spot "first reply" number.

We want a single, prominent metric line that says: *"first user-press-Enter to assistant-first-audio: N ms"*. This is the number that matters for perceived responsiveness, and the one that tells us at a glance whether warmups did their job.

### In scope
- Track wall-clock from the moment `record_audio` returns (user has finished speaking) to the moment the **first** PCM byte hits `aplay` stdin on the response. That's time-to-first-audio (TTFA).
- Also track:
  - Time from Enter → STT response (transcription latency, including arecord)
  - Time from STT response → first TTS audio (LLM thinking + first-sentence TTS)
- On the **first** completed voice cycle after `genie-core` starts, print a clearly-formatted banner with the breakdown. Subsequent cycles continue with the existing one-liners.
- Banner format suggestion:

```
=== first voice reply latency ===
  user-spoke -> STT done:      N ms
  STT done   -> first audio:   N ms
  total (first reply):         N ms
=================================
```

### Out of scope
- Percentiles / histograms across many cycles (handled by external observability).
- Per-component latency timing inside whisper or llama (already logged at INFO level).
- Network or HA-tool latency.

## Acceptance criteria

- The banner prints exactly once per `genie-core` run, after the first complete cycle.
- The numbers reasonably reflect what `(STT: ...)` and `(LLM+TTS: ...)` already show, but as a single cohesive block.
- Subsequent cycles are unchanged.
- No regression in voice loop correctness.

## Notes

- Current observed latency in the warmup test was ~2000 ms total, likely dominated by remaining LLM cold path. Goal here is just to *measure and surface* — the optimization is separate.
- Doesn't need a new config knob; always-on print is fine since it's one line per run.

## Related

- #7 — LLM warmup (already merged)
- #17 — whisper warmup (just merged in #18)

PR #61, opened 2026-05-17
voice: real streaming TTS — speak first sentence during LLM gen (#26)

### LINKED ISSUE #26: Real streaming TTS: speak first sentence during LLM generation, not after
## Motivation

`voice::streaming::stream_and_speak` exists specifically to "speak while LLM is still generating," per its own docstring:

> Instead of waiting for the full LLM response before speaking, this module buffers tokens into sentences and speaks each sentence as soon as it completes.

But the current implementation does the opposite — it collects every token first, then splits into sentences, then speaks each one:

```rust
let response = llm.chat_stream(messages, Some(max_tokens), |token| {
    full_response.push_str(token);
    sentence_buffer.push_str(token);
}).await?;          // <-- blocks until the entire LLM response is in

// only NOW do we split sentences and call tts_engine.speak() per sentence
let sentences = split_sentences(&voice_text);
for sentence in &sentences { tts_engine.speak(sentence).await?; }
```

The new 5-phase latency banner (#19) makes the cost obvious: on a recent test the "LLM until first sentence" line read **16,998 ms** — that's the entire LLM response time, not "until the first sentence was buffered." If LLM emitted "You're welcome, Jared!" within the first ~1500 ms of the stream, the first audio could start ~15 s earlier.

## Proposal

True streaming: detect sentence boundaries inside the `chat_stream` callback and feed completed sentences to a concurrent TTS task as they arrive. LLM streaming and TTS playback run in parallel from sentence 1 onward.

Sketch:

```rust
let (tts_tx, mut tts_rx) = tokio::sync::mpsc::unbounded_channel::<String>();
let tts_engine = Arc::new(tts_engine.clone_for_streaming()); // see TtsEngine refactor below
let tts_task = tokio::spawn({
    let tts_engine = tts_engine.clone();
    async move {
        while let Some(sentence) = tts_rx.recv().await {
            if let Err(e) = tts_engine.speak(&sentence).await {
                tracing::warn!(error = %e, "TTS error on streamed sentence");
            }
        }
    }
});

let response = llm.chat_stream(messages, Some(max_tokens), |token| {
    full_response.push_str(token);
    sentence_buffer.push_str(token);
    if let Some(sentence) = take_complete_sentence(&mut sentence_buffer) {
        let _ = tts_tx.send(sentence);
    }
}).await?;
// Flush remaining buffer as the last sentence
if !sentence_buffer.trim().is_empty() {
    let _ = tts_tx.send(sentence_buffer.trim().to_string());
}
drop(tts_tx);
tts_task.await?;
```

### Required supporting work

- **`TtsEngine` must be cheaply shareable across tasks.** The `child: Option<Child>` field (from pipe mode, not currently used by `speak()`) blocks `Clone`. Either drop it (we always spawn-per-call now), or wrap the engine in `Arc<...>`, or split it into a config struct + per-call subprocess spawner. Recommendation: `Arc<TtsEngine>` since most fields are cheap to share read-only.
- **Sentence-boundary detector that works on partial input.** The current `split_sentences` only runs on complete text. The streaming version needs to consume a `&mut String` buffer, return the next complete sentence (if any), and leave the trailing fragment behind for the next token.

### Out of scope
- Per-sentence TTS bypass for very short sentences (e.g., "OK."). Already handled by `split_sentences`' length filter.
- Backpressure between LLM stream and TTS task. The mpsc channel will buffer; on long LLM outputs the queue can grow but the memory cost is trivial.

## Acceptance criteria

- On the #19 latency banner, "LLM until first sentence" drops from full-response time (~5-17 s) to "time until the LLM emits the first complete sentence" (~1-3 s typical for Phi-4-mini on Orin Nano).
- Audible behavior: the assistant starts speaking the first sentence while the LLM is still generating later sentences. No audible gap between sentences worse than today.
- No regression on the short-response path (1-2 sentence responses still speak as a single chunk).
- Unit test covering `take_complete_sentence` partial-input parsing.
- 62+ existing voice unit tests still pass.

## Related

- #19 — latency banner that made this problem measurable and visible.
- #7 — LLM warmup; complementary fix at the model-load layer. Streaming TTS reduces the cost of the *first inference's wall time*, warmup reduces the cost of *first inference's setup*.

hunnyboy1217:
Closes #26.

The `voice::streaming::stream_and_speak` module previously waited for the full LLM response before invoking any TTS, so the "LLM until first sentence" phase of the first-reply latency banner (#19) measured the full response time (~5-17 s) instead of time-to-first-sentence.

This PR rewires it to true parallel streaming:

- New `SentenceStreamer` detects sentence boundaries inside the LLM streaming callback, tracks markdown code-fence depth across chunks, strips inline markdown / URLs / list & header prefixes per sentence, and merges sub-8-char openers (`OK!`, `Sure!`, `Dr.`) into the next clause so Piper never gets a 1-4 char fragment. Cap of 3 spoken sentences matches the existing cap in `format::for_voice`. ASCII `.!?` only confirm as a sentence boundary when followed by whitespace (or EOF via `finish()`), so URLs like `example.com` and decimals like `3.14` don't false-split mid-token. CJK `。！？` emit immediately since CJK convention doesn't put whitespace after punctuation.
- `stream_and_speak` spawns a tokio task that drains an `mpsc::UnboundedSender<String>` and calls `TtsEngine::speak` sequentially (preserves audio order and the half-duplex `post_silence_ms` gate). The channel is flushed + closed + awaited on both success and error paths.
- `TtsEngine` is shared between the producer and the consumer task via `Arc<TtsEngine>`. `speak()` already took `&self` and spawned a fresh Piper per call, so no `Mutex` was needed. All three call sites in `voice_loop.rs` updated.
- The `FIRST_SPEAK_CALLED_AT` marker already in `TtsEngine::speak` ([voice/tts.rs:203](crates/genie-core/src/voice/tts.rs#L203)) now fires at the true first-sentence moment, so the latency banner's "LLM until first sentence" figure becomes meaningful.

### Verification

13 new unit tests cover partial-chunk feeding, the 3-sentence cap, opener merging, trailing-fragment flush, inline-markdown / link / URL / bullet stripping, fenced code suppression (including fence markers split across chunks), and CJK punctuation.

### Out of scope

- Per-sentence Piper warmup (would close the residual gap below first audio).
- Re-introducing the "speak as one chunk for short replies" smoothing — incompatible with true parallel streaming since we don't know the reply length until it's done.

## Real Behavior Proof

### What I ran

This branch was prepared without a local Rust toolchain (the system `cargo` from apt is 1.75; the workspace is on edition 2024 and needs rustc ≥1.85). Verification was therefore driven through CI per CONTRIBUTING.md option 2 + 3:

- `Cross-compile (aarch64 / Jetson)` job — builds the real aarch64 release binaries (`genie-core`, `genie-ctl`, `genie-governor`, `genie-health`, `genie-api`) with the same `make jetson` recipe that ships to Orin Nano. Green on the latest push.
- `CI / fmt` — `cargo fmt --all -- --check`.
- `CI / ubuntu-latest / check + clippy + test` and `CI / macos-latest / check + clippy + test` — `cargo check`, `cargo clippy -- -D warnings`, `cargo test --lib`, `cargo test --tests`, `cargo test --doc` across the full workspace. The 13 new `SentenceStreamer` tests run here.
- `CI / cargo clippy + test (--no-default-features)` — chat-only build axis with `voice` + `telegram` disabled. The streaming module is voice-gated so it's compiled out on this axis, but the surrounding `llm` + `voice_loop` glue still has to type-check.

### What I observed

- The 13 new unit tests in `crates/genie-core/src/voice/streaming.rs` pass on every CI run, including the URL false-split regression (`strips_markdown_links`, `strips_raw_urls` — these both fed inputs containing `https://example.com/...` and asserted exactly one sentence emits rather than the two the original detector produced when it treated the `.` inside `example.com` as a sentence boundary).
- `cargo clippy --workspace --all-targets -- -D warnings` is clean — including the `clippy::collapsible_str_replace` lint that fired on the first CI run against the two `.replace(char, ...)` chains in `clean_sentence`. Resolved by combining them into char-slice form (`.replace(['(', ')'], ", ")` and `.replace(['[', ']', '{', '}', '"'], "")`).
- `cargo fmt --all -- --check` clean — the first CI run flagged five `.await\n{` over-wraps in `voice_loop.rs` and two `feed_all(...)` over-wraps in the test module that rustfmt wanted collapsed to single lines; both fixed.
- **No Jetson hardware run on my end.** The change is an inner refactor of the sentence-detection + TTS-dispatch pipeline that the existing voice loop already drives; the on-device measurement that issue #26's acceptance criteria call for ("LLM until first sentence" latency drops from ~5-17 s to ~1-3 s, observable in the issue #19 latency banner) needs a maintainer with an Orin + LyraT + a speaker to read the banner after merge. The unit tests cover the boundary-detection and cleanup logic that determines *what* is spoken; only the latency measurement remains for hardware confirmation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,20 @@
 
 ## Unreleased
 
+- **Real streaming TTS** (#26): the voice loop now detects sentence
+  boundaries inside the LLM streaming callback and forwards completed
+  sentences to a concurrent TTS task immediately, instead of waiting
+  for the full response before speaking. First sentence reaches the
+  speaker as soon as the LLM finishes emitting it (typically 1-3 s
+  after wake) rather than after the full response (~5-17 s). New
+  `SentenceStreamer` strips inline markdown / URLs / list markers /
+  fenced code blocks per-sentence, merges sub-8-char openers with the
+  next clause, and caps spoken output at 3 sentences to match
+  `format::for_voice`. `TtsEngine` is shared across the producer and
+  consumer halves via `Arc<TtsEngine>`. The first-reply latency banner
+  (#19) now reports the true LLM-until-first-sentence figure instead
+  of the full-response figure.
+
 ## 1.0.0-alpha.9 - 2026-05-18
 
 Alpha 9 is the **CI / supply-chain hardening + voice-frontend maturation**
--- a/crates/genie-core/src/voice/streaming.rs
+++ b/crates/genie-core/src/voice/streaming.rs
@@ -1,158 +1,523 @@
-//! Streaming voice pipeline — speak while LLM is still generating.
+//! Streaming voice pipeline — speak while the LLM is still generating.
 //!
-//! Instead of waiting for the full LLM response before speaking,
-//! this module buffers tokens into sentences and speaks each sentence
-//! as soon as it completes. The user hears the first sentence 2-5 seconds
-//! sooner than the blocking pipeline.
+//! Unlike the previous implementation (which only sliced the *completed*
+//! LLM response into sentences before speaking), this module detects
+//! sentence boundaries inside the streaming token callback and forwards
+//! each completed sentence to a concurrent TTS task immediately. The
+//! first sentence reaches the speaker as soon as the LLM finishes
+//! emitting it, not after the entire response is collected.
 //!
 //! Pipeline:
-//!   LLM streams tokens → sentence buffer → format for voice → Piper TTS → speaker
-//!                         ↓ (concurrent)
-//!                         next sentence buffering continues
+//!   chat_stream tokens
+//!       └─► SentenceStreamer.feed() ──► mpsc ──► TTS task ──► aplay
+//!                                                  (Arc<TtsEngine>)
+//!
+//! The TTS task drains the channel sequentially so audio plays in order
+//! and the half-duplex `post_silence_ms` gate still works between
+//! sentences.
+
+use std::sync::Arc;
 
 use anyhow::Result;
+use tokio::sync::mpsc;
 
-use super::format;
 use super::tts::TtsEngine;
 
-/// Buffer LLM tokens into complete sentences, speak each one immediately.
+/// Maximum sentences to speak per response. Matches the cap that
+/// `format::for_voice` enforces on the non-streaming path, so voice
+/// replies stay short whether the caller streams or not.
+const MAX_SPOKEN_SENTENCES: usize = 3;
+
+/// Stream the LLM response and speak each sentence in parallel.
 ///
-/// Returns the full assembled response text.
+/// Returns the full assembled response text (untruncated, so callers can
+/// log/persist the complete reply even though only the first
+/// `MAX_SPOKEN_SENTENCES` are spoken).
 pub async fn stream_and_speak(
     llm: &crate::llm::LlmClient,
     messages: &[crate::llm::Message],
     max_tokens: u32,
-    tts_engine: &TtsEngine,
+    tts_engine: Arc<TtsEngine>,
 ) -> Result<String> {
-    let mut full_response = String::new();
-    let mut sentence_buffer = String::new();
-    let mut sentences_spoken = 0;
+    let (tx, mut rx) = mpsc::unbounded_channel::<String>();
 
-    let response = llm
+    let tts_for_task = Arc::clone(&tts_engine);
+    let tts_task = tokio::spawn(async move {
+        let mut spoken: usize = 0;
+        while let Some(sentence) = rx.recv().await {
+            if spoken == 0 {
+                eprintln!("[voice] Speaking (streaming)...");
+            }
+            if let Err(e) = tts_for_task.speak(&sentence).await {
+                tracing::warn!(error = %e, "TTS error on sentence");
+            }
+            spoken += 1;
+        }
+        spoken
+    });
+
+    let mut full_response = String::new();
+    let mut streamer = SentenceStreamer::new(MAX_SPOKEN_SENTENCES);
+
+    let stream_result = llm
         .chat_stream(messages, Some(max_tokens), |token| {
             full_response.push_str(token);
-            sentence_buffer.push_str(token);
+            for sentence in streamer.feed(token) {
+                let _ = tx.send(sentence);
+            }
         })
-        .await?;
+        .await;
 
-    // After streaming completes, speak any remaining buffered text.
-    // (During streaming, we collected everything; now we speak sentence by sentence.)
-    // For V1, we do a simpler approach: split the complete response into sentences
-    // and speak each one. This is still faster than the old "wait for everything
-    // then speak everything" because Piper processes each sentence independently.
+    // Always flush and close the channel so the TTS task exits cleanly,
+    // whether the LLM stream succeeded or errored partway through.
+    if let Some(tail) = streamer.finish() {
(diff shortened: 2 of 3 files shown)


hunnyboy1217:
Hi, @ai-hpc,
All CI tests are passed successfully.
Could you please review my PR?

ai-hpc (approved):
This is genuine engineering and exactly what #26 asked for. The previous "stream tokens, then split the *completed* response into sentences before speaking" approach measured the full LLM response as one chunk in the #19 latency banner; this PR makes sentence detection happen *inside* the streaming callback and forwards each completed sentence to a concurrent TTS task immediately, so "LLM until first sentence" finally reflects time-to-first-sentence instead of total-response-time.

The `SentenceStreamer` design is the right shape for the constraints:

- **ASCII boundary needs whitespace confirmation** (`awaiting_ascii_boundary` flag) — so `example.com`, `3.14`, `i.e.` mid-sentence don't false-split. CJK `。！？` emit immediately because the convention doesn't put whitespace after.
- **Code-fence tracking with `fence_marker_buffer`** correctly handles ``` markers split across token boundaries. Inline single-backticks fall through (the buffer clears on non-backtick) so `\`like this\`` text passes through as noise to strip downstream.
- **`MIN_SENTENCE_CHARS = 8` merge** keeps Piper from getting `"OK!"` / `"Sure!"` / `"Dr."` fragments that would synthesize into 200ms glitches. Sized to admit dense CJK sentences without merging unnecessarily.
- **`MAX_SPOKEN_SENTENCES = 3` cap** matches `format::for_voice`'s cap so the audible output is consistent whether the caller streams or not.
- **`finish()` flushes the trailing fragment** with no minimum-length check (EOF can't merge into anything later), so a short closing clause still gets spoken.
- **TTS task drains `mpsc::UnboundedReceiver<String>` sequentially** — preserves audio order and the half-duplex `post_silence_ms` gate between sentences. `Arc<TtsEngine>` shared between producer and consumer works because `speak()` already took `&self` and spawns a fresh Piper per call.
- **Channel flushed + closed + awaited on both success and error** of `chat_stream`, so the TTS task always exits cleanly and we don't leak a tokio task on LLM stream errors.

13 unit tests pin the edge cases: partial-chunk feeding, 3-sentence cap, opener merging, trailing-fragment flush, inline markdown / link / URL / bullet stripping, fenced code suppression with split-marker handling, CJK punctuation. Tests assert *exact* sentence counts on tricky inputs (URLs, decimals), which is the regression-resistant shape.

CI: all 6 checks green on `2ed0b88` (fmt, clippy, test, aarch64 cross-compile, `--no-default-features`, plus the ubuntu + macos + coverage matrix from an earlier workflow shape). The PR body checklist FAILURE entry in the rollup is stale.

The honest note about no Jetson hardware on the contributor's side is the right call per CONTRIBUTING.md option 2/3 — the unit tests verify *what* is spoken; the on-device latency-banner readout that #26's acceptance criterion needs is a maintainer-side check after merge.

Commit hygiene: personal Gmail on `Hunnyboy1217@gmail.com` for 2/3 commits (latest merge-from-main uses noreply correctly). Soft norm flag carried forward as a "next time" note; not a hard block for this PR given the single-contributor identity and no AI trailers. Going in.

ai-hpc:
Merged at `f08d59ce2e6c064b8c351f910c6fce4a6e5dbb0e`.
Thanks to @hunnyboy1217!

Issue #52, opened 2026-05-17
Default to genie-ai-runtime v1.0.0 (replaces stale #33)


ai-hpc:
Replaces #33 (closed as stale — it pinned `v0.1.0-alpha.8` and #32 has fully landed since).

## Background

`genie-ai-runtime` shipped **v1.0.0** on 2026-05-16. That release:

- Cuts the `-alpha.N` suffix (positioned as the first stable release).
- Engine behavior identical to `v0.1.0-alpha.12` — same kernels, same quantization, same code path.
- INT8 KV cache became the default (was opt-in in alpha.11; halves KV memory).
- Validated against **Qwen3-4B Q4_K_M** on **Jetson Orin Nano Super 8 GB** (aarch64 only).
- OpenAI-compatible HTTP server.
- Source build only — no prebuilt binaries shipped with the release.

Reference perf numbers from the v1.0.0 release (Qwen3-4B Q4_K_M, Orin Nano Super), measured against alpha.2 as the baseline:

| Metric | alpha.2 | v1.0.0 | Δ |
|---|---|---|---|
| Prefill | 8.2 tok/s | **38.0 tok/s** | +363% |
| Decode | 7.5 tok/s | **9.9 tok/s** | +32% |
| Cold TTFT | 2200 ms | **877 ms** | −60% |
| Warm-turn TTFT | n/a | **444 ms** | new metric (alpha.9 persistent KV) |

Meanwhile on the consumer side, `genie-claw` has completed the abstraction work from #32. All five slices are on `main`:

| Slice | PR | SHA |
|---|---|---|
| Facade + `backend_name()` | #35 | `60a1ec8` |
| Voice operator-log surfaces | #38 | `7ac2311` |
| Startup log + `/api/health` field | #39 | `1435ee4` |
| Runtime/service tooling (mode + ctl) | #40 | `311c948` |
| Config selection | #43 | `6bec4d1` |
| Docs + config + Cargo feature declarations | direct | `339f84f` |

`[services.llm].backend = "genie_ai_runtime"` is selectable end-to-end today via `geniepod.toml`. This issue is the parity-validated flip.

## Goal

Change the default value of `[services.llm].backend` in `deploy/config/geniepod.toml` from `"llama_cpp"` to `"genie_ai_runtime"`, ship the supporting systemd unit + setup-jetson workflow, and validate the change against v1.0.0 on real Orin Nano hardware.

[...]

PR #56, opened 2026-05-17
feat(deploy): genie-ai-runtime v1.0.0 build+install pipeline (issue #54)

bittoby:
Closes #54

Adds the enabling infrastructure to install `genie-ai-runtime` v1.0.0
alongside the existing llama.cpp setup. Install-only — does NOT change
the running backend; operator does the cutover by hand per the printed
instructions (mirrors the `--model qwen3-4b` safety pattern from PR #46).

## What lands
- `deploy/systemd/genie-ai-runtime.service` — `Conflicts=genie-llm.service`
  so the :8080 binding-conflict is loud rather than silent.
- `deploy/systemd/genie-ai-runtime-warmup.service` — 120 s `/health` poll +
  one tiny `/v1/chat/completions` to force model load + CUDA kernel compile.
- `deploy/setup-jetson.sh --runtime genie-ai-runtime` — clone v1.0.0,
  `cmake --build`, install `jetson-llm-server` + `jetson-llm` to
  `/opt/geniepod/bin/`, print cutover + rollback instructions. No-flag
  path is unchanged.
- `make jetson-ai-runtime JETSON_HOST=<ip> JETSON_USER=<user>` — composes
  `deploy-systemd + deploy-setup + ssh setup-jetson.sh --runtime`.

## Out of scope (deliberate)
- Default-backend flip in `geniepod.toml` → #52.
- Qwen 3 4B auto-download → PR #46.
- CHANGELOG entry → done at release-cut time.
- CI build of genie-ai-runtime → would need self-hosted Jetson runner.

## Test plan
- [x] `bash -n` syntax clean.
- [x] `--help`, `--runtime` (no arg), `--runtime bogus`, `--unknown-flag`
      all behave correctly (exit 2 + clear error).
- [x] `make -n jetson-ai-runtime` shows the expected SSH command sequence.
- [x] `make -n jetson` / `make -n deploy` unchanged.
- [ ] **Manual Jetson run:** `make jetson-ai-runtime` against a real Orin Nano,
      verify build completes in <25 min and binaries land.
- [ ] **Manual cutover:** flip `[services.llm]` to `genie_ai_runtime`,
      restart genie-core, confirm `genie-ctl status` reports the new backend.
- [ ] **Manual rollback:** symmetric flip back to `llama_cpp`, confirm
      llama.cpp resumes serving.

PR #55, opened 2026-05-17
Fix: Default to Genie-ai-runtime

RenzoMXD:
## Summary

This PR implements issue #52 by flipping GenieClaw’s Jetson default LLM runtime from `llama_cpp` to `genie_ai_runtime` and coupling it with the Qwen3-4B Q4_K_M default model path. It also adds the runtime-specific systemd units, updates Jetson setup automation to source-build/install `genie-ai-runtime v1.0.0`, and updates docs/changelog to reflect the new default topology while keeping `llama_cpp` fully selectable as fallback.
## Related Issue
Closes: #52  
Related: #44 (model default coupling), supersedes stale #33 context.

---

## Change Type (select all)

- [x] Feature
- [x] Bug fix
- [x] Refactor (service wiring/default routing)
- [x] Documentation update
- [x] Deployment / Ops change
- [x] Breaking change (default behavior change on Jetson deploy config)
- [ ] Security fix
- [ ] Performance-only change (performance impact expected but not solely benchmark code)

---

### 1) Default backend/model flip (Jetson config)

- `deploy/config/geniepod.toml`
  - Set `[services.llm].backend = "genie_ai_runtime"` (default)
  - Set `[services.llm].systemd_unit = "genie-ai-runtime.service"`
  - Set `llm_model_name = "qwen"`
  - Set `llm_model_path = "/opt/geniepod/models/qwen3-4b-q4_k_m.gguf"`

### 2) Dev config preserved as llama.cpp

- `deploy/config/geniepod.dev.toml`
  - Explicit `backend = "llama_cpp"`
  - Keeps `systemd_unit = "genie-llm.service"`

### 3) New runtime units added

- `deploy/systemd/genie-ai-runtime.service` (new)
- `deploy/systemd/genie-ai-runtime-warmup.service` (new)

### 4) Jetson setup pipeline updated

- `deploy/setup-jetson.sh`
  - Installs/builds `genie-ai-runtime v1.0.0` (`jetson-llm-server`, `jetson-llm`) when missing
  - Downloads Qwen3-4B Q4_K_M as the default model artifact
  - Enables LLM units based on configured backend (`genie-ai-runtime*` vs `genie-llm*`)
  - Handles stale drop-ins for both runtime unit families

### 5) Service wiring/general defaults aligned

- `deploy/systemd/genie-core.service`
  - Removed hard `Wants/After=genie-llm.service` to avoid pinning to legacy runtime unit
- `deploy/systemd/genie-governor.service`
[...]

Issue #54, opened 2026-05-17
Build + install pipeline for genie-ai-runtime v1.0.0 (ship genie-ai-runtime.service alongside genie-llm.service)


ai-hpc:
## Background

#52 proposes flipping the default LLM backend to `genie-ai-runtime` v1.0.0. The five abstraction slices from #32 (now closed) made the *config selection* possible — `[services.llm].backend = "genie_ai_runtime"` is a one-line change today.

But the actual binary isn't installable. `genie-ai-runtime` v1.0.0 [ships source-only](https://github.com/GeniePod/genie-ai-runtime/releases/tag/v1.0.0) — no prebuilt artifacts. `deploy/setup-jetson.sh` doesn't compile any C++. `deploy/systemd/` has no `genie-ai-runtime.service`. So the only path today is the operator manually:

```bash
ssh aihpc@<jetson-ip>
git clone --branch v1.0.0 https://github.com/GeniePod/genie-ai-runtime.git
cd genie-ai-runtime && cmake -B build && cmake --build build
sudo cp build/jetson-llm-server /opt/geniepod/bin/
sudo nano /etc/systemd/system/genie-ai-runtime.service     # write the unit by hand
sudo systemctl daemon-reload && sudo systemctl start genie-ai-runtime
```

That's six error-prone steps per Jetson. This issue is the **enabling infrastructure** — make the build + install + service-unit pipeline a `setup-jetson.sh --runtime genie-ai-runtime` (or `make jetson-llm-runtime`) one-liner.

## Goal

After this lands, deploying genie-ai-runtime v1.0.0 to a Jetson is one of:

```bash
# Option A — Makefile target
make jetson-ai-runtime JETSON_HOST=<ip> JETSON_USER=aihpc

# Option B — setup-jetson.sh flag (analogous to --model qwen3-4b from PR #46)
sudo /opt/geniepod/setup-jetson.sh --runtime genie-ai-runtime
```

End state per Jetson:

- `/opt/geniepod/bin/jetson-llm-server` built from upstream v1.0.0
- `/opt/geniepod/bin/jetson-llm` (CLI, also from the v1.0.0 build)
- `/etc/systemd/system/genie-ai-runtime.service` (new)
- `/etc/systemd/system/genie-ai-runtime-warmup.service` (new)
- Existing `genie-llm.service` (llama.cpp) **untouched** — still selectable via `backend = "llama_cpp"`
[...]

PR #63, opened 2026-05-17
fix(setup-jetson): auto-fallback to llama.cpp when jetson-llm-server is missing (#60)

### LINKED ISSUE #60: setup-jetson.sh enables genie-ai-runtime.service without verifying jetson-llm-server is installed
## Summary

On a fresh \`make deploy\` against a Jetson that does NOT yet have \`genie-ai-runtime\` installed, \`setup-jetson.sh\` warns about the missing \`jetson-llm-server\` binary in step \`[5/6]\`, then proceeds to \`systemctl enable genie-ai-runtime.service\` anyway in step \`[6/6]\`. The operator ends up with the default LLM unit enabled but unable to start (the unit's \`ConditionPathExists=/opt/geniepod/bin/jetson-llm-server\` will skip-start it cleanly, but \`genie-core\` will then have no healthy LLM backend reachable).

## Reproduction

Verified on Jetson Orin Nano Super 8 GB (aihpc@192.168.55.1) running v1.0.0-alpha.9 (commit \`60b0303\`), 2026-05-18:

\`\`\`text
$ make deploy JETSON_HOST=192.168.55.1 JETSON_USER=aihpc
# … cross-compile + scp succeed …

$ ssh aihpc@192.168.55.1 'bash /opt/geniepod/setup-jetson.sh'
=== GeniePod Jetson Setup ===
[1/6] Creating directories...
[2/6] Checking binaries... (all 6 OK)
[3/6] Checking config... OK
[4/6] Checking LLM model... OK: Qwen3-4B-Q4_K_M.gguf (2.4G)
[5/6] Checking LLM runtimes...
  NOT FOUND: jetson-llm-server
    Default backend (genie-ai-runtime) requires it. Install via:
      bash /opt/geniepod/setup-jetson.sh --runtime genie-ai-runtime
    Or set [services.llm].backend = "llama_cpp" in geniepod.toml
    to use the legacy llama.cpp path instead.
  OK: llama-server (legacy fallback backend)
…
[6/6] Enabling systemd services...
  Enabled: genie-ai-runtime
  Enabled: genie-ai-runtime-warmup
  …
\`\`\`

Note that \`[5/6]\` and \`[6/6]\` disagree: step 5 says "default backend requires \`jetson-llm-server\`, which is missing", and step 6 enables the unit that needs it anyway.

## Why this is a real problem

1. **Contradicts the v1.0.0-alpha.9 default-flip promise** (PR #55, closes #52). The \`alpha.9\` narrative says "default backend on Jetson is \`genie-ai-runtime\`". For that to be true on a fresh deploy, either:
   - \`setup-jetson.sh\` (run without flags) must install the binary, OR
   - the script must fall back to \`llama_cpp\` automatically when the default's binary is missing, OR
   - the script must hard-fail and instruct the operator to either install the runtime or switch \`backend = "llama_cpp"\` before re-running.
2. **The \`--runtime genie-ai-runtime\` flag (PR #56) is documented as opt-in**, but issue #52 / PR #55 changed the *default* — those two designs are now in conflict for first-time operators.
3. **\`genie-core\` will then start without a reachable LLM backend** and either fail health checks or spin on retries until the operator notices.

## Proposed fix options

Pick one (in order of preferred):

### Option A: Auto-fallback to \`llama_cpp\` when the default's binary is missing
- In step \`[5/6]\`: if \`jetson-llm-server\` is missing AND \`llama-server\` is present, automatically downgrade the effective \`LLM_BACKEND\` to \`llama_cpp\` for step \`[6/6]\`'s enable loop, AND patch the deployed \`/etc/geniepod/geniepod.toml\` so \`[services.llm].backend\` and \`systemd_unit\` match the actually-installed backend.
- Print a loud NOTE block: "default backend (genie-ai-runtime) is not installed; falling back to llama.cpp. To run the default, install genie-ai-runtime via \`--runtime genie-ai-runtime\` then re-run setup."

### Option B: Make \`--runtime genie-ai-runtime\` the default for fresh installs
- When \`[services.llm].backend = "genie_ai_runtime"\` in the deployed config AND \`jetson-llm-server\` is missing, treat that as an implicit \`--runtime genie-ai-runtime\` and invoke the existing \`install_runtime_genie_ai_runtime\` function (from #56). Wire \`make deploy\` to default-pass \`--runtime genie-ai-runtime\` for the same reason.
- Downside: a \`make deploy\` now does a from-source build on first run (CMake + gcc + ~5-15 min on Orin Nano).

### Option C: Hard-fail in step \`[5/6]\`
- \`exit 1\` if the configured backend's binary is missing. Don't enable services that can't start. Print the two unambiguous remediation paths and stop.
- Downside: stricter; might frustrate operators who explicitly want to deploy config now and install the runtime later.

My recommendation is **Option A** (auto-fallback) because:
- It preserves the property that \`setup-jetson.sh\` always leaves the box in a working state.
- The operator who *wants* the default explicitly will see the NOTE and run the install step.
- It mirrors the existing pattern of \`audio_denoiser\` falling back to \`sox\` when DFN's binary is missing.

## Cross-references

- #52 (default-flip, closed by #55)
- #54 (runtime install pipeline, closed by #56)
- #55 (the default-flip PR that introduced this gap)
- #56 (the \`--runtime\` flag that's now the documented manual step)
- v1.0.0-alpha.9 release notes' install section currently tells operators to run \`--runtime genie-ai-runtime\` as a separate step — once this is fixed, that note can come out.

bittoby:
## Summary

Closes #60
Fixes the gap between alpha.9's default-flip (#52 / PR #55) and
the `--runtime genie-ai-runtime` install pipeline (#54 / PR #56): a fresh
`make deploy` against a Jetson that doesn't yet have `jetson-llm-server`
installed used to enable `genie-ai-runtime.service` anyway, leaving
`genie-core` with no reachable LLM backend.

Implements **Option A** from the issue (auto-fallback to `llama_cpp` when
`jetson-llm-server` is missing but `llama-server` is present). Mirrors the
existing `audio_denoiser` fall-back pattern.

## What changes

In `deploy/setup-jetson.sh`:

- **Step [5/6]** now actively resolves the effective backend, not just
  reports presence. Probes both binaries, reads
  `[services.llm].backend` from `geniepod.toml`, and decides:
  - both present → use configured backend
  - configured backend's binary missing, alternate present → auto-fallback,
    patch `[services.llm]` in `geniepod.toml` (`backend` + `systemd_unit`),
    print a loud NOTE pointing at `--runtime genie-ai-runtime` for the
    operator who wants the default explicitly
  - both missing → loud ERROR + `SKIP_LLM_UNITS=true` + remediation block;
    whisper/piper/core still install cleanly so re-running setup after
    a manual fix is one command
- **Step [6/6]** now consumes the resolved `EFFECTIVE_BACKEND` instead of
  re-reading `geniepod.toml`. Comment in-place explains why; re-reading
  would re-introduce the bug.
- **Final "Start services"** instructions consume the same resolved
  variable so the printed command matches what was actually enabled.

Config patcher uses `awk` with section tracking so it can't touch the
identical `systemd_unit = ` key in `[services.core]` right above
`[services.llm]`.

## What I didn't change

- The systemd unit files (`ConditionPathExists` already prevents a unit
  from starting when its binary is missing — this PR just stops them from
  being *enabled* in the first place).
- `make deploy` flow (Option B would have required this; not chosen).
- The `--runtime genie-ai-runtime` opt-in flag from PR #56 — still works
  as-is, still recommended path for operators who explicitly want the
  Jetson-tuned runtime.

## Test plan

- [x] `bash -n deploy/setup-jetson.sh` — syntax clean.
- [x] `bash deploy/setup-jetson.sh --help` — exit 0, expected output.
- [x] `awk` patcher unit-tested against a realistic `geniepod.toml`
      fixture: rewrites only `[services.llm]`, preserves inline comments,
      doesn't touch `[services.core]`'s `systemd_unit = ` or commented
      `# [services.something]` sections.
- [x] Decision matrix walked for all 6 scenarios
      (configured × {jetson-llm-server, llama-server} presence).

### DIFF
--- a/deploy/setup-jetson.sh
+++ b/deploy/setup-jetson.sh
@@ -326,22 +326,128 @@ if [ -n "$MODEL_CHOICE" ] && [ "$MODEL_CHOICE" != "qwen3-4b" ]; then
     fi
 fi
 
-# 5. Check LLM runtimes.
+# 5. Check LLM runtimes and resolve the *effective* backend.
+#
+# The configured default in geniepod.toml is genie_ai_runtime (issue #52 / PR
+# #55). On a fresh deploy without `--runtime genie-ai-runtime` the binary it
+# needs (`jetson-llm-server`) isn't on disk yet, which used to leave step
+# [6/6] enabling units that couldn't start (issue #60).
+#
+# Resolution policy (issue #60, Option A):
+#   - configured backend's binary present → use that backend.
+#   - configured = genie_ai_runtime, binary missing, llama-server present
+#       → auto-fallback to llama_cpp; patch geniepod.toml so the runtime,
+#         systemd, and the operator's eyes all agree; print a loud NOTE.
+#   - configured = llama_cpp, binary missing, jetson-llm-server present
+#       → symmetric auto-fallback to genie_ai_runtime.
+#   - neither binary present → don't enable any LLM unit in [6/6]; print
+#     a remediation block and let the rest of setup continue (whisper/
+#     piper still install cleanly, operator fixes LLM and re-runs).
 echo "[5/6] Checking LLM runtimes..."
-if [ -f "$GENIEPOD_DIR/bin/jetson-llm-server" ]; then
+
+HAVE_JETSON_LLM=false
+[ -f "$GENIEPOD_DIR/bin/jetson-llm-server" ] && HAVE_JETSON_LLM=true
+HAVE_LLAMA=false
+[ -f "$GENIEPOD_DIR/bin/llama-server" ] && HAVE_LLAMA=true
+
+if [ "$HAVE_JETSON_LLM" = "true" ]; then
     echo "  OK: jetson-llm-server"
 else
-    echo "  NOT FOUND: jetson-llm-server"
-    echo "    Default backend (genie-ai-runtime) requires it. Install via:"
-    echo "      bash $0 --runtime genie-ai-runtime"
-    echo "    Or set [services.llm].backend = \"llama_cpp\" in geniepod.toml"
-    echo "    to use the legacy llama.cpp path instead."
+    echo "  NOT FOUND: jetson-llm-server (default backend binary)"
 fi
-
-if [ -f "$GENIEPOD_DIR/bin/llama-server" ]; then
+if [ "$HAVE_LLAMA" = "true" ]; then
     echo "  OK: llama-server (legacy fallback backend)"
 else
-    echo "  NOT FOUND: llama-server (optional fallback)"
+    echo "  NOT FOUND: llama-server (legacy fallback backend)"
+fi
+
+CONFIGURED_BACKEND="$(awk -F'"' '/^backend = / {print $2; exit}' "$CONFIG_DIR/geniepod.toml" 2>/dev/null || true)"
+[ -z "$CONFIGURED_BACKEND" ] && CONFIGURED_BACKEND="genie_ai_runtime"
+# Normalize the hyphenated alias documented next to backend = in the toml.
+case "$CONFIGURED_BACKEND" in
+    llama-cpp)         CONFIGURED_BACKEND="llama_cpp" ;;
+    genie-ai-runtime)  CONFIGURED_BACKEND="genie_ai_runtime" ;;
+esac
+
+EFFECTIVE_BACKEND="$CONFIGURED_BACKEND"
+SKIP_LLM_UNITS=false
+
+patch_services_llm_backend() {
+    # Rewrite the [services.llm] section's `backend = ...` and
+    # `systemd_unit = ...` values to the requested target, in place,
+    # without touching surrounding comments or other sections. awk is
+    # used instead of sed so that "systemd_unit =" in [services.core]
+    # right above can't be hit by accident.
+    local new_backend="$1"
+    local new_unit="$2"
+    local cfg="$CONFIG_DIR/geniepod.toml"
+    local tmp
+    tmp="$(mktemp)"
+    awk -v nb="$new_backend" -v nu="$new_unit" '
+        BEGIN { in_llm = 0 }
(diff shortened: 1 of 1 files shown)


ai-hpc (approved):
Exactly the fix #60 asked for. The resolution logic in step [5/6] is symmetric (covers both directions of fallback plus the both-missing ERROR path), `SKIP_LLM_UNITS=true` cleanly suppresses the unit-enable loop instead of letting [6/6] enable something that can't start, and the "Start services" footer is wired to the same resolved variable so the printed command matches what was actually enabled.

The awk patcher is the riskiest piece and it's done right: tracks `in_llm` on `^\[services\.llm\]`, exits on any other `^\[`, anchors the replacement pattern at the start of the line so the commented `# backend = "llama_cpp"` example in the canonical toml can't be hit by accident, and uses `sub` so inline trailing comments are preserved. The identical `systemd_unit =` line in `[services.core]` above `[services.llm]` is safely out of scope. Writes via `sudo install -m 600` so the 0600 permissions `make deploy` set on `/etc/geniepod/geniepod.toml` are preserved.

One small note that's worth flagging but not blocking: the trailing inline comment on the patched `backend =` line still says "Default on Jetson. OpenAI-compatible genie-ai-runtime (:8080)." after a fallback to llama_cpp. Cosmetic only — an operator who re-runs `make deploy` gets the canonical toml back and the fallback fires again — but a future polish could either strip the comment or replace it with a generated one that matches the current backend.

All 6 checks green on `bb23d60` (fmt, clippy, test, aarch64 cross-compile, shellcheck, ruff). Going in.

ai-hpc:
Merged at `bbad3f792b54278e0c08d60a5d239dfbc8da418a`.
Thanks to @bittoby!

PR #62, opened 2026-05-17
ci: add --no-default-features axis for chat-only builds (#59)

statxc:
Closes #59.

## Summary

Adds a `no-default-features` job to `.github/workflows/ci.yml` that runs `cargo clippy` and `cargo test` for `-p genie-core -p genie-ctl --no-default-features --all-targets --locked` with `-D warnings` on every push to `main` and every PR.

This closes acceptance criterion #8 of #41 that PR #57 explicitly deferred ("`.github/` doesn't exist on `main` yet"). `ci.yml` landed in PR #37, so the deferral premise no longer holds and the chat-only build path can finally be guarded in CI.

## Why

PR #57 made `voice` an opt-out Cargo feature on `genie-core` and `genie-ctl`. With `--no-default-features` (the chat-only shape used by Telegram / Slack / Discord bots, remote-LLM nodes with no microphone, and macOS / Windows dev hosts), the build skips ~6 kLoC of ALSA-coupled voice / wakeword / STT / TTS / AEC code. Today, nothing in CI exercises that
axis — every `#[cfg(feature = "voice")]` (and the latent `#[cfg(feature = "telegram")]` cleanups from #57) is one stray `use` statement away from silently bit-rotting until the next time someone actually runs `cargo build --no-default-features` on a Mac.

This PR closes that loop.

## What changed

One new job in `.github/workflows/ci.yml`, parallel to the existing `fmt` / `clippy` / `test` jobs:

- `cargo clippy -p genie-core -p genie-ctl --no-default-features
  --all-targets --locked -- -D warnings`
- `cargo test -p genie-core -p genie-ctl --no-default-features
  --all-targets --locked`

Implementation notes that match the issue:

- **Scoped to `genie-core` + `genie-ctl`** — the only two crates that  declare a `voice` feature. `--workspace --no-default-features` would  just duplicate the existing `--workspace` job for feature-less crates.
- **`clippy` + `test` only, no `fmt`** — formatting doesn't depend on  features.
- **Distinct `Swatinem/rust-cache` `shared-key: ci-nodefault`** — a  different feature-set means a different `target/` graph; sharing  `ci-clippy` / `ci-test` would thrash both caches.
- **`-D warnings`** matches the rest of `ci.yml` (already enforced via  `RUSTFLAGS` at the workflow level).
- **15-minute timeout** matches the existing `clippy` job's budget.

## Local verification

`rustc 1.95.0 / cargo 1.95.0` on x86_64-linux:

[...]

PR #64, opened 2026-05-17
feat(telegram): phase 2 voice replies via Piper + sendVoice (#42)

bittoby:
## Summary

Phase 2 of #42 (phase 1 landed in #53). When `[telegram.voice].reply_as_voice
= true`, the bot synthesizes the assistant reply with Piper, encodes it as
OGG/Opus with ffmpeg, and uploads via Telegram's `sendVoice` endpoint instead
of plain `sendMessage`. Falls back to text on any failure so no reply is ever
silently dropped.

## What changes

- `TelegramVoiceConfig` gains `reply_as_voice` (default `false` — opt-in) and
  `max_reply_chars` (default `800` — long replies fall back to text to stay
  under Telegram's 1 MB `sendVoice` cap).
- `TelegramVoiceRuntimeConfig` plumbs Piper paths through, reusing
  `core.piper_path` + `core.piper_model` so the configured on-device voice
  matches the Telegram voice reply.
- `send_reply()` in `telegram.rs` routes between text and voice; three new
  helpers (`synthesize_reply_to_wav`, `wav_to_ogg_opus`, `send_voice`) do
  the work. Each step's failure path logs and falls back to text.
- `voice_reply_gate()` extracted as a pure function so the off / empty /
  over-cap / under-cap / multi-byte-Unicode policy can be unit-tested
  without spinning up Piper or HTTP.

## Design — matches phase 1's decoupling

The adapter still does not import `voice::tts::TtsEngine`. It subprocess-
shells out to Piper directly and uses `reqwest::multipart` for `sendVoice`.
Keeps phase 2 usable on chat-only deployments and honors phase 1's commit
message that explicitly warned against recoupling the two paths.

## Test plan

- [x] `cargo check --workspace --all-targets` — clean
- [x] `cargo clippy -p genie-core --all-targets --all-features -- -D warnings`
      — clean
- [x] `cargo fmt --all -- --check` — clean
- [x] `cargo test -p genie-core telegram::` — **11/11 passed**
      (5 phase-1 + 1 wire-format + 5 new phase-2 gate tests)
- [ ] **Manual end-to-end** against a real Telegram bot + Jetson with Piper
      installed: flip `reply_as_voice = true`, send a voice message, expect
      a voice reply.
- [ ] **Manual fallback verification**: temporarily rename `piper` →
      expect text reply with a `tracing::warn!` line.

## What's out of scope (still deferred per the issue)

- **Phase 3** — speaker identity from Telegram `user_id` → memory routing.
  Tracked separately if useful.
- **Mock-Telegram integration test** for `sendVoice` — phase 1's integration-
  test gap still applies; would need an HTTP mock + real Piper + ffmpeg
  fixtures in `crates/genie-core/tests/`.
- **Streaming TTS** — voice messages are bounded; batch synth is fine.

## Related

- #42 — parent issue
- #53 — phase 1 (voice-in)
- #57 — `voice` Cargo feature (independent; phase 2 stays out of it on purpose)

### DIFF
--- a/crates/genie-common/src/config.rs
+++ b/crates/genie-common/src/config.rs
@@ -474,6 +474,22 @@ pub struct TelegramVoiceConfig {
     /// 16 kHz mono WAV that Whisper consumes.
     #[serde(default = "defaults::telegram_voice_ffmpeg_path")]
     pub ffmpeg_path: PathBuf,
+
+    /// Reply to incoming voice messages with a synthesized voice message
+    /// instead of (or in addition to) text. Phase 2 of issue #42: Piper
+    /// synthesizes WAV, ffmpeg encodes it as OGG/Opus, the bot uploads via
+    /// the Telegram `sendVoice` endpoint. Falls back to text on any failure
+    /// (Piper missing, ffmpeg missing, sendVoice error, etc.) so no reply is
+    /// ever silently dropped.
+    #[serde(default)]
+    pub reply_as_voice: bool,
+
+    /// Hard cap on the assistant text fed to Piper. Long-form responses
+    /// produce long voice messages that hit Telegram's 1 MB sendVoice limit;
+    /// when the text is over this length the bot falls back to text reply.
+    /// Tuned for the 60–90 s of OGG/Opus that comfortably fits under 1 MB.
+    #[serde(default = "defaults::telegram_voice_max_reply_chars")]
+    pub max_reply_chars: usize,
 }
 
 impl Default for TelegramVoiceConfig {
@@ -483,6 +499,8 @@ impl Default for TelegramVoiceConfig {
             max_voice_duration_secs: defaults::telegram_voice_max_duration_secs(),
             delete_temp_audio: defaults::telegram_voice_delete_temp_audio(),
             ffmpeg_path: defaults::telegram_voice_ffmpeg_path(),
+            reply_as_voice: false,
+            max_reply_chars: defaults::telegram_voice_max_reply_chars(),
         }
     }
 }
@@ -1472,6 +1490,12 @@ mod defaults {
     }
     pub fn telegram_voice_ffmpeg_path() -> PathBuf {
         PathBuf::from("ffmpeg")
+    }
+    pub fn telegram_voice_max_reply_chars() -> usize {
+        // Roughly the upper bound that comfortably encodes under Telegram's
+        // 1 MB sendVoice limit at Piper's typical OGG/Opus output rate.
+        // Long-form replies fall back to text.
+        800
     }
     pub fn web_search_enabled() -> bool {
         true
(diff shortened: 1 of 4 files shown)


ai-hpc (approved):
Clean phase-2 implementation that matches phase 1's decoupling: subprocesses Piper + ffmpeg directly, uses `reqwest::multipart` for `sendVoice`, never imports `voice::tts::TtsEngine`. The PR body's commit-message warning against recoupling the two paths is honored end-to-end.

The fallback discipline is what makes this safe to land opt-in:

- Piper synth failure → `send_text`
- ffmpeg ogg/opus encode failure → `send_text`
- `sendVoice` HTTP failure → `send_text`
- Reply over `max_reply_chars` → `send_text` + structured log
- `reply_as_voice = false` (default) → `send_text` unconditionally

So the worst-case post-deploy regression is "voice reply silently falls back to text and you see a `tracing::warn!` line" — no dropped messages.

`voice_reply_gate` extracted as a pure function is exactly the right testing seam. The unit tests cover every branch including the one that often gets missed — `chars().count()` vs `len()` on multi-byte Unicode (`東京こんにちは` = 7 chars / ~21 bytes). The `>` vs `>=` boundary at `at_cap = "x".repeat(100)` is the right call (gives the operator exactly the cap they configured).

`TempCleanup` guards both the WAV and OGG via RAII so a panic between the Piper call and the `sendVoice` call still cleans up. Empty-WAV detection (`meta.len() < 128`) catches the case where Piper exits 0 but writes nothing — would have produced an opaque `sendVoice` "invalid voice file" error otherwise.

The `24k` opus bitrate is hardcoded; that's appropriate for voice and well under the 1 MB / 60-90 s envelope claimed for `max_reply_chars = 800`. If an operator ever wants studio-quality voice replies they can lift the cap and the bitrate together in a follow-up — no need to make it configurable today.

One thing worth flagging that isn't a blocker: this PR was opened against a base that didn't yet include #62's `cargo clippy + test (--no-default-features)` CI job, so the chat-only build axis didn't actually run on this PR's tip in CI. I ran it locally on `60bb1bf` and it's clean (321 tests pass, clippy `-D warnings` clean), so it would have been green if rebased — but for future PRs touching `crates/genie-core/src/telegram.rs`, the no-default-features job on `main` will catch any chat-only build regression on the next push.

The `delete_temp_audio` toml docstring still talks only about voice-in temp files — minor doc rot now that it also controls cleanup of reply OGG/WAV. Worth one-line refresh in a follow-up but not blocking.

All 4 checks green on `60bb1bf` (fmt, clippy, test, aarch64 cross-compile). Going in.

ai-hpc:
Merged at `bd4c835065f82725d17f5dfab8cc375ddb750bb5`.
Thanks to @bittoby!

PR #66, opened 2026-05-18
setup: fix LLM fallback config patch

### LINKED ISSUE #65: setup-jetson.sh patch_services_llm_backend fails with 'Permission denied' — auto-fallback prints NOTE but never patches geniepod.toml
## Summary

PR #63 (closes #60) added an auto-fallback in `setup-jetson.sh` that's supposed to patch `/etc/geniepod/geniepod.toml` so `[services.llm].backend` and `systemd_unit` match the actually-installed backend. The fallback **prints** the NOTE block as designed, but the `awk` invocation that does the rewrite fails with `Permission denied` and the config is silently left unpatched. `genie-core` then reads the original (mismatched) config on next boot and tries to talk to the backend whose binary isn't installed.

## Reproduction

Verified on Jetson Orin Nano Super 8 GB (`aihpc@192.168.55.1`) running `main` at `bbad3f7` (PR #63 merged), 2026-05-18:

```text
$ ssh aihpc@192.168.55.1 'bash /opt/geniepod/setup-jetson.sh'
…
[3/6] Checking config...
  OK: /etc/geniepod/geniepod.toml
  Secured config permissions       ← chmods to 0600, owner root
…
[5/6] Checking LLM runtimes...
  NOT FOUND: jetson-llm-server (default backend binary)
  OK: llama-server (legacy fallback backend)

  NOTE: configured backend (genie_ai_runtime) is not installed.
        Auto-falling back to llama.cpp so this box leaves setup
        in a working state (issue #60).
        Patching /etc/geniepod/geniepod.toml:
            backend      = "llama_cpp"
            systemd_unit = "genie-llm.service"
        To run the default backend instead, install it via:
            bash /opt/geniepod/setup-jetson.sh --runtime genie-ai-runtime
        then re-run setup-jetson.sh (or hand-edit geniepod.toml back).
awk: cannot open /etc/geniepod/geniepod.toml (Permission denied)
…
```

After the run, `sudo grep '^backend\|^systemd_unit' /etc/geniepod/geniepod.toml` still shows `backend = "genie_ai_runtime"` and `systemd_unit = "genie-ai-runtime.service"` — the patch never happened.

## Root cause

`patch_services_llm_backend` in `deploy/setup-jetson.sh` (from PR #63):

```bash
patch_services_llm_backend() {
    …
    awk -v nb=… -v nu=… '…' "$cfg" > "$tmp" && sudo install -m 600 "$tmp" "$cfg"
    rm -f "$tmp"
}
```

`awk` runs as the unprivileged user. `"$cfg"` is `/etc/geniepod/geniepod.toml`, which step `[3/6]` just chmodded to `0600 root:root`. `awk` can't open it, the pipeline fails, `sudo install` never runs (short-circuited by `&&`), and the script keeps going (no `set -e` on this function, and the function's overall exit status isn't checked by the caller).

## Knock-on effects

1. **The "auto-fallback" doesn't actually fall back.** `geniepod.toml` still selects the missing backend.
2. **Step `[6/6]` is still correct** because it reads the in-memory `EFFECTIVE_BACKEND` (set before the patch attempt), so it enables `genie-llm` + `genie-llm-warmup` as the operator expects.
3. But the next time anything reads `geniepod.toml` from disk — `genie-core` on start, `genie-ctl status`, `make deploy` overwriting it back to canonical — it sees the unpatched `genie_ai_runtime` selection and behaves accordingly.

## Proposed fix

Run the read step through `sudo` so it can open the 0600 file. The redirect happens in the calling shell, not under sudo, so `sudo awk … > tmp` won't work — use `sudo tee` instead:

```bash
patch_services_llm_backend() {
    local new_backend="$1"
    local new_unit="$2"
    local cfg="$CONFIG_DIR/geniepod.toml"
    local tmp
    tmp="$(mktemp)"
    if ! sudo awk -v nb="$new_backend" -v nu="$new_unit" '
        BEGIN { in_llm = 0 }
        /^\[services\.llm\]/   { in_llm = 1; print; next }
        /^\[/ && !/^\[services\.llm\]/ { in_llm = 0 }
        in_llm && /^backend = "[^"]*"/ {
            sub(/^backend = "[^"]*"/, "backend = \"" nb "\"")
            print; next
        }
        in_llm && /^systemd_unit = "[^"]*"/ {
            sub(/^systemd_unit = "[^"]*"/, "systemd_unit = \"" nu "\"")
            print; next
        }
        { print }
    ' "$cfg" | sudo tee "$tmp" > /dev/null; then
        echo "  ERROR: failed to read $cfg for patching"
        rm -f "$tmp"
        return 1
    fi
    sudo install -m 600 "$tmp" "$cfg"
    sudo rm -f "$tmp"
}
```

Two improvements bundled in: (a) the `if !` wrapper detects the failure instead of silently continuing, and (b) the function returns a non-zero exit code that the caller in step `[5/6]` should also check and either `exit 1` or downgrade to `SKIP_LLM_UNITS=true` so this class of bug surfaces at setup time rather than at `genie-core` startup.

## Cross-references

- #60 — the gap PR #63 was supposed to close
- #63 — the PR that introduced this regression
- The systemd unit's `ConditionPathExists=/opt/geniepod/bin/jetson-llm-server` means `genie-ai-runtime.service` will skip-start cleanly even when enabled, so the worst-case symptom on a fresh box is "`genie-core` has no reachable LLM backend" rather than a crash loop — but it's the same failure mode #60 was filed for.

ai-hpc:
## Summary
- Read `[services.llm].backend` through `sudo awk` after setup locks `geniepod.toml` down to `0600`.
- Make `patch_services_llm_backend` read the root-owned config through sudo and return errors on read/install failures.
- Check both auto-fallback patch calls and abort setup instead of enabling services against an unpatched config.
- Add a focused regression test pinning the privileged read + checked fallback behavior.

## Tests
- `bash -n deploy/setup-jetson.sh`
- `cargo test -p genie-core --test tool_dispatch_test setup_script_privileged_llm_backend_patch_is_checked`
- `git diff --check`

Closes #65


ai-hpc:
Post-merge validation notes:

- CI was green before merge: fmt, clippy, tests, no-default-features axis, aarch64 release build, ruff, and shellcheck.
- Jetson setup path was validated with the v1.0.0 genie-ai-runtime release assets: binaries downloaded, SHA256 checks passed, config selected backend = "genie_ai_runtime" and systemd_unit = "genie-ai-runtime.service".
- Runtime health passed with Qwen3-4B on genie-ai-runtime.
- start_all.sh / stop_all.sh deployed cleanly; start_all now queues warmup oneshots with --no-block and exits normally.

Merged as 32dea9d. Follow-up runtime note: the latest board check still reported genie-core offline, so that should be diagnosed from genie-core logs separately from issue #65.

PR #68, opened 2026-05-18
fix(voice/stt): per-call nonce on transcribe_pcm tempfile (race under #42 concurrent load)

### LINKED ISSUE #67: voice/stt: transcribe_pcm temp WAV is PID-only — concurrent calls race on /tmp/geniepod-stt-<pid>.wav
## Background

`SttEngine::transcribe_pcm` at [crates/genie-core/src/voice/stt.rs:170](crates/genie-core/src/voice/stt.rs#L170) names its temp WAV by process ID only:

```rust
let tmp_path = format!("/tmp/geniepod-stt-{}.wav", std::process::id());
write_wav(&tmp_path, pcm_data, sample_rate).await?;
let result = self.transcribe_file(&tmp_path).await;
let _ = tokio::fs::remove_file(&tmp_path).await;
```

Within a single process, the PID is constant — so two concurrent `transcribe_pcm` calls collide on the same path. The interleaving that loses data is straightforward:

1. Call A writes `/tmp/geniepod-stt-<pid>.wav` with PCM_A.
2. Call B writes `/tmp/geniepod-stt-<pid>.wav` with PCM_B, overwriting PCM_A mid-flight.
3. Call A's `transcribe_file(&tmp_path)` now reads PCM_B and returns the wrong transcript for the wrong caller.
4. Whichever call finishes first runs `remove_file`, deleting the file out from under the other.

This is latent today because the on-device voice loop is single-utterance / half-duplex and never calls `transcribe_pcm` concurrently within one process. It becomes load-bearing the moment we exercise #42's acceptance criterion:

> Concurrent voice messages from multiple Telegram users are handled in parallel when STT server mode is in use.

Under that load, two Telegram users speaking at once → two concurrent `transcribe_pcm` calls → transcripts cross-contaminate or fail outright. Worse, the failure mode is silent — you get a transcript, it's just from the wrong user, which then poisons memory and the LLM reply path for both users.

The merged Telegram phase-1/phase-2 voice path ([crates/genie-core/src/telegram.rs:230-233](crates/genie-core/src/telegram.rs#L230-L233), [#L324-L327](crates/genie-core/src/telegram.rs#L324-L327)) already does the right thing — `pid + random nonce` — but only for its own `.ogg`/`.wav` staging files. The shared STT tempfile underneath is still PID-only, so the Telegram fix doesn't help once the audio reaches `transcribe_pcm`.

## Scope

Just the one site under concurrent load:

- [crates/genie-core/src/voice/stt.rs:170](crates/genie-core/src/voice/stt.rs#L170) — `transcribe_pcm`

Out of scope (PID-only but only called from the single-utterance on-device voice loop, no concurrency today — fix opportunistically if touched but not load-bearing):

- [stt.rs:441](crates/genie-core/src/voice/stt.rs#L441) `flush_path`, [stt.rs:517](crates/genie-core/src/voice/stt.rs#L517) `wav_path`, [stt.rs:611](crates/genie-core/src/voice/stt.rs#L611) normalized, [stt.rs:721-722](crates/genie-core/src/voice/stt.rs#L721-L722) mono/dfn
- [voice/tts.rs:381](crates/genie-core/src/voice/tts.rs#L381)

## Fix

Append a per-call random suffix, same shape the Telegram path already uses:

```rust
let pid = std::process::id();
let nonce: u32 = rand::random();
let tmp_path = format!("/tmp/geniepod-stt-{pid}-{nonce}.wav");
```

`pid + nonce` (not full UUID) keeps it consistent with `telegram.rs` and avoids pulling a new crate.

Use an RAII guard for cleanup (mirroring `TempCleanup` in `telegram.rs`) so the `remove_file` happens on every exit path including panic-unwind — the current `let _ = tokio::fs::remove_file(...)` after `transcribe_file` leaks the WAV on any early-return / `?` propagation.

## Acceptance

- [ ] `transcribe_pcm` names its tempfile with `pid + nonce`.
- [ ] Cleanup is RAII-style (guard struct) so the tempfile is removed on `?`-propagated errors and panic-unwinds, not just the happy path.
- [ ] Unit test: spawn N concurrent `transcribe_pcm` calls with N distinct PCM payloads against a stubbed `transcribe_file` that records the bytes it observed, assert each call saw its own bytes.
- [ ] Manual repro on a Telegram-voice-enabled build: two users send voice messages within the same ~second; each receives a reply tied to their own transcript.

## Cross-references

- #42 — Telegram voice support; the AC that exposes this race
- [crates/genie-core/src/telegram.rs:230-233](crates/genie-core/src/telegram.rs#L230-L233) — the `pid + nonce` pattern to mirror
- #21 — test harness reliability (the concurrent-calls unit test belongs in that style)

plind-junior:
## Summary

- `SttEngine::transcribe_pcm` named its staging WAV by PID only (`/tmp/geniepod-stt-<pid>.wav`), so two concurrent calls in the same process collided on one path — the later writer would overwrite the earlier reader's input mid-flight, and the first `remove_file` would delete the file out from under the second call.
- Latent on the on-device voice loop (single-utterance, half-duplex) but becomes load-bearing for the #42 phase-1 AC: *"Concurrent voice messages from multiple Telegram users are handled in parallel when STT server mode is in use."* Under that load the failure mode is silent — both users get a transcript, just from each other's audio.
- Fix: append a process-local atomic nonce (`/tmp/geniepod-stt-<pid>-<nonce>.wav`) and wrap the path in a `TempFile` RAII guard so cleanup runs on every exit path — `?`-propagated errors and panic-unwinds included, which the prior `let _ = remove_file(...)` after the `.await` missed. Pattern mirrors the `pid + nonce` scheme `telegram.rs` already uses for its own `.ogg`/`.wav` staging.
- No new dependencies — `AtomicU64` is sufficient for within-process uniqueness, which is what this race needs.

## Scope

Just the one site exposed to concurrent load:

- [`crates/genie-core/src/voice/stt.rs`](crates/genie-core/src/voice/stt.rs) — `transcribe_pcm`

Intentionally **not** touched in this PR — still PID-only but only reached from the single-utterance on-device voice loop, no concurrency today; happy to bundle if reviewers prefer:

- `voice/stt.rs:441` `flush_path`, `:517` recording path, `:611` normalized, `:721-722` mono / DFN staging
- `voice/tts.rs:381`

## Test plan

- [x] `cargo test -p genie-core --lib voice::stt::` — 10/10 pass on top of `upstream/main` (`bd4c835`), including the two new tests:
  - `transcribe_pcm_temp_paths_are_unique_under_concurrency` — 64 concurrent nonce allocations produce 64 distinct paths (the direct regression test for this bug)
  - `temp_file_guard_removes_on_drop` — confirms the RAII guard actually removes the file when dropped
- [x] `cargo clippy -p genie-core --lib --tests -- -D warnings` — clean
- [ ] Manual repro on a Telegram-voice-enabled build: two Telegram users send voice messages within ~1 s of each other; each receives a reply tied to their own transcript (not cross-contaminated). Requires reviewer with `[telegram.voice]` configured — I don't have that rig.

## Related

- #42 — Telegram voice support; the AC that exposes this race
- `crates/genie-core/src/telegram.rs` — the `pid + nonce` pattern this PR mirrors (introduced by #53 / #64)

> Note: I haven't filed a tracking issue for this race yet — happy to do so and update the title to `fix(voice/stt): … (#<issue>)` if the reviewer prefers that flow.

Fixes #67

### DIFF
--- a/crates/genie-core/src/voice/stt.rs
+++ b/crates/genie-core/src/voice/stt.rs
@@ -1,8 +1,37 @@
 use anyhow::Result;
 use std::sync::Mutex;
+use std::sync::atomic::{AtomicU64, Ordering};
 use tokio::io::{AsyncBufReadExt, AsyncWriteExt, BufReader};
 use tokio::process::{Child, Command};
 use tokio::sync::mpsc;
+
+/// Monotonic per-process counter for unique temp-file suffixes. Pairs with
+/// the PID so two concurrent `transcribe_pcm` calls in the same process
+/// cannot collide on `/tmp/geniepod-stt-*.wav`.
+static TEMP_NONCE: AtomicU64 = AtomicU64::new(0);
+
+fn next_temp_nonce() -> u64 {
+    TEMP_NONCE.fetch_add(1, Ordering::Relaxed)
+}
+
+/// RAII guard that deletes a temp file on drop. Ensures cleanup runs on
+/// every exit path — `?`-propagated errors and panic-unwinds included.
+struct TempFile(String);
+
+impl TempFile {
+    fn new(path: String) -> Self {
+        Self(path)
+    }
+    fn path(&self) -> &str {
+        &self.0
+    }
+}
+
+impl Drop for TempFile {
+    fn drop(&mut self) {
+        let _ = std::fs::remove_file(&self.0);
+    }
+}
 
 /// "Speech end" marker for the #19 latency banner: stamped inside
 /// `record_audio` the moment arecord returns, before DFN/sox preprocessing
@@ -167,11 +196,11 @@ impl SttEngine {
     /// Transcribe raw PCM audio bytes (16kHz, 16-bit, mono).
     /// Writes to a temp WAV file, then transcribes.
     pub async fn transcribe_pcm(&self, pcm_data: &[u8], sample_rate: u32) -> Result<Transcript> {
-        let tmp_path = format!("/tmp/geniepod-stt-{}.wav", std::process::id());
-        write_wav(&tmp_path, pcm_data, sample_rate).await?;
-        let result = self.transcribe_file(&tmp_path).await;
-        let _ = tokio::fs::remove_file(&tmp_path).await;
-        result
+        let pid = std::process::id();
+        let nonce = next_temp_nonce();
+        let tmp = TempFile::new(format!("/tmp/geniepod-stt-{pid}-{nonce}.wav"));
+        write_wav(tmp.path(), pcm_data, sample_rate).await?;
+        self.transcribe_file(tmp.path()).await
     }
 
     async fn transcribe_via_server(&self, port: u16, wav_path: &str) -> Result<Transcript> {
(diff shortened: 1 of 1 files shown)


ai-hpc (approved):
Verified the bug independently against `main` (`bd4c835`): the prior `let tmp_path = format!("/tmp/geniepod-stt-{}.wav", std::process::id())` does only embed PID, so two `transcribe_pcm` calls from the same `genie-core` process would race on a single path. Latent on the half-duplex on-device loop, load-bearing once Telegram phase 1 starts handling concurrent users — the PR body has this exactly right.

The fix is the minimal-surface call: process-local `AtomicU64` for unique suffixes (no new deps, no cross-process coordination needed because the race is purely within one process), `pid + nonce` shape matching the existing pattern in `telegram.rs`. The two unit tests are the right ones:

- `transcribe_pcm_temp_paths_are_unique_under_concurrency` — 64 concurrent nonce allocations into a `HashSet`, asserts cardinality. Direct regression for the bug.
- `temp_file_guard_removes_on_drop` — proves the RAII guard actually fires. Without this, a refactor that quietly drops the guard inside a wider scope (or replaces `tmp` with `let _ = TempFile::new(...)`) would silently bring the leak back.

Quiet side-benefit worth flagging that isn't called out in the PR body: the original `let _ = tokio::fs::remove_file(&tmp_path).await` ran *after* the `.await` on `transcribe_file`, so if `transcribe_file` errored via `?`-propagation the cleanup line was unreachable and the temp WAV leaked into `/tmp`. The `TempFile` RAII guard fixes that too — cleanup now runs on the error-propagation path, the panic-unwind path, and the success path uniformly.

The "out of scope" list in the PR body (the four other PID-only temp paths in `voice/stt.rs` and one in `voice/tts.rs`) is the right call. They're all reached only from the single-utterance on-device voice loop where the contention doesn't exist today; a follow-up PR can sweep them as a single hygiene pass once someone actually exposes them to concurrent load.

Two minor items, neither blocking:

- `TempFile::path()` returns `&str`; passing `tmp.path()` to `write_wav` and `transcribe_file` works because they take `&str`, but `&Path` would be slightly more idiomatic. Not worth a churn round.
- The static `TEMP_NONCE` uses `Ordering::Relaxed` — correct here because uniqueness is the only invariant and `fetch_add` is atomic regardless of ordering. Worth a one-line "Relaxed is sufficient — uniqueness is the only invariant" comment for the next reader, but the rustdoc above already implies it.

I'll squash-merge with a maintainer-written body once CI lands (separate norm note below about the commit trailer). Going in.

ai-hpc:
Quick norm-flag for future PRs, @plind-junior — this repo's commit history doesn't carry `Co-Authored-By: Claude` (or any AI-assistant) trailers. Please drop the trailer from your local commits before pushing on future contributions. Fine to use Claude / Copilot / etc. to draft the code — we just keep the `git log` / `git shortlog` attribution to the human contributor so credit is unambiguous.

No action needed on this PR — the squash-merge body is maintainer-written so the trailer drops automatically. Just a heads-up for the next one.

Thanks for the careful writeup and the regression test!

ai-hpc:
Merged at `3711c515ce205c17fd8050f3b00e7fa48db735f4`.
Thanks to @plind-junior!

PR #72, opened 2026-05-18
core: avoid push-to-talk loop under systemd

### LINKED ISSUE #71: genie-core exits under systemd when voice push-to-talk reads EOF
## Context

After PR #70 validation, the normal stack starts successfully, `genie-ai-runtime` is healthy, and OpenAI-compatible chat completions work, but `genie-ctl status` reports:

```text
Core: offline
```

`systemctl status genie-core` shows the service repeatedly exits successfully and restarts:

```text
Active: activating (auto-restart)
Process: ... ExecStart=/opt/geniepod/bin/genie-core (code=exited, status=0/SUCCESS)
```

## Evidence

Recent `journalctl -u genie-core` repeats this pattern:

```text
INFO voice mode — starting voice interaction loop
[voice] Capture device: plughw:APE,0  |  Playback device: plughw:0,0
INFO STT using long-running whisper-server (model stays loaded in GPU) port=8178 model=/opt/geniepod/models/ggml-small.bin
INFO voice conversation started conv_id=...
[voice] LLM backend connected (genie-ai-runtime)
=== GeniePod Voice Mode (Push-to-Talk) ===
Press Enter to speak (3 sec), 'quit' to exit.
[voice] Press Enter to speak > INFO voice loop exited
systemd[1]: genie-core.service: Deactivated successfully.
systemd[1]: genie-core.service: Scheduled restart job, restart counter is at 344.
```

The deployed systemd unit intentionally has non-interactive stdin:

```ini
# /etc/systemd/system/genie-core.service
ExecStart=/opt/geniepod/bin/genie-core
Environment=GENIEPOD_CONFIG=/etc/geniepod/geniepod.toml
StandardInput=null
Restart=always
RestartSec=3
```

The default deployed config currently enables voice push-to-talk:

```toml
voice_enabled = true
wakeword_script = ""              # Empty = push-to-talk
```

## Suspected root cause

`genie-core` treats `core.voice_enabled = true` as enough to enter voice mode. With `wakeword_script = ""`, voice mode selects push-to-talk, which reads from stdin. Under systemd, `StandardInput=null` immediately returns EOF, so the push-to-talk loop exits successfully. Because the unit has `Restart=always`, systemd restarts it forever.

This is separate from PR #70 / issue #69. The runtime service and `/v1/chat/completions` are working.

## Desired behavior

`genie-core.service` should remain up in daemon/API mode under systemd. Push-to-talk should only run when stdin is interactive or when explicitly requested in a terminal. Wake-word daemon mode may still be valid under systemd if a real `wakeword_script` is configured.

Possible directions:

- Make `genie-core` ignore push-to-talk config when stdin is not a TTY and no wakeword script exists, then fall back to HTTP server mode.
- Or change the deployed default config so `voice_enabled = false` unless wake-word daemon mode is configured.
- Or split push-to-talk and daemon voice mode into explicit config fields so systemd cannot accidentally select an stdin-driven loop.

## Acceptance criteria

- Starting `genie-core.service` under systemd does not enter push-to-talk mode when `StandardInput=null`.
- `systemctl status genie-core` stays `active (running)` instead of `activating (auto-restart)`.
- `genie-ctl status` reports `Core: online` after the service starts.
- Interactive push-to-talk still works when running `genie-core --voice` from a terminal.
- Wake-word mode still works under systemd when a valid `wakeword_script` is configured.

carlos4s:
## Summary

Fixes #71.

`genie-core.service` runs with `StandardInput=null`, but the deployed config can request voice mode with `voice_enabled = true` and no wakeword script. That selects push-to-talk mode, which immediately reads stdin EOF under systemd, exits successfully, and then restarts forever because the unit has `Restart=always`.

This PR gates voice mode so it only runs when:

- voice is requested and stdin is interactive, allowing terminal push-to-talk; or
- voice is requested and a wakeword script is configured and present, allowing daemon wake-word mode.

Otherwise `genie-core` logs a warning and runs the HTTP API instead of entering the stdin-driven push-to-talk loop.

## Validation

- `cargo fmt --all --check`
- `cargo test -p genie-core --bin genie-core voice_mode_tests`
- `cargo test -p genie-core --no-default-features --bin genie-core voice_mode_tests`
- `git diff --check`

## Notes

This keeps issue #71 separate from issue #69 / PR #70. `genie-ai-runtime` and the OpenAI-compatible chat endpoint were already validated; this fixes the `genie-core` daemon mode selection.

### DIFF
--- a/crates/genie-core/src/lib.rs
+++ b/crates/genie-core/src/lib.rs
@@ -52,6 +52,7 @@ pub mod prompt;
 pub mod reasoning;
 pub mod repl;
 pub mod runtime_contract;
+pub mod runtime_mode;
 pub mod security;
 pub mod server;
 pub mod skills;
--- a/crates/genie-core/src/main.rs
+++ b/crates/genie-core/src/main.rs
@@ -172,7 +172,8 @@ async fn main() -> Result<()> {
         || config.core.voice_enabled;
     let wakeword_available =
         !config.core.wakeword_script.as_os_str().is_empty() && config.core.wakeword_script.exists();
-    let voice_runnable = should_enter_voice_mode(voice_requested, interactive, wakeword_available);
+    let startup_decision =
+        runtime_mode::decide_startup_mode(voice_requested, interactive, wakeword_available);
 
     // Whether the running binary actually has the voice subsystem compiled in.
     // When voice was requested but the binary is chat-only (issue #41 feature
@@ -180,13 +181,13 @@ async fn main() -> Result<()> {
     // voice-tagged `geniepod.toml` still deploys cleanly on a chat-only build.
     #[cfg(feature = "voice")]
     let voice_mode = {
-        if voice_requested && !voice_runnable {
+        if startup_decision.blocked_push_to_talk() {
             tracing::warn!(
                 "push-to-talk voice mode requested without an interactive terminal and no wakeword script is available; \
                  running HTTP API only"
             );
         }
-        voice_runnable
+        startup_decision.enters_voice()
     };
     #[cfg(not(feature = "voice"))]
     let voice_mode = {
@@ -198,7 +199,7 @@ async fn main() -> Result<()> {
                  enable the voice loop."
             );
         }
-        let _ = voice_runnable;
+        let _ = startup_decision;
         false
     };
 
@@ -369,39 +370,6 @@ fn atty_check() -> bool {
     #[cfg(not(unix))]
     {
         false
-    }
-}
-
-fn should_enter_voice_mode(
-    voice_requested: bool,
-    stdin_interactive: bool,
-    wakeword_available: bool,
-) -> bool {
-    voice_requested && (stdin_interactive || wakeword_available)
-}
-
-#[cfg(test)]
-mod voice_mode_tests {
-    use super::should_enter_voice_mode;
-
-    #[test]
-    fn systemd_push_to_talk_falls_back_to_http() {
-        assert!(!should_enter_voice_mode(true, false, false));
-    }
-
-    #[test]
-    fn terminal_push_to_talk_still_runs() {
-        assert!(should_enter_voice_mode(true, true, false));
-    }
-
-    #[test]
-    fn wakeword_voice_can_run_without_terminal() {
-        assert!(should_enter_voice_mode(true, false, true));
-    }
-
-    #[test]
-    fn voice_disabled_stays_http_only() {
-        assert!(!should_enter_voice_mode(false, true, true));
     }
 }
 
--- a/crates/genie-core/src/main.rs
+++ b/crates/genie-core/src/main.rs
@@ -170,13 +170,24 @@ async fn main() -> Result<()> {
     let voice_requested = std::env::args().any(|a| a == "--voice")
         || std::env::var("GENIEPOD_VOICE").unwrap_or_default() == "1"
         || config.core.voice_enabled;
+    let wakeword_available =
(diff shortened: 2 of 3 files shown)


carlos4s:
Jetson acceptance for #71:

```text
● genie-core.service - GeniePod Core — Local Home AI Engine
     Active: active (running)
   Main PID: 99890 (genie-core)
```

`genie-core` no longer exits/restarts under systemd. It now logs the expected fallback and starts the HTTP API:

```text
WARN push-to-talk voice mode requested without an interactive terminal and no wakeword script is available; running HTTP API only
INFO starting HTTP chat API port=3000
INFO genie-core HTTP server listening addr=127.0.0.1:3000
```

`genie-ctl status` now reports:

```text
Core:      ok
```

Validated on Jetson after rebuilding the aarch64 binary from PR head and redeploying binaries.

ai-hpc (approved):
Verified the bug independently against the deployed `genie-core.service` — `StandardInput=null` + default `voice_enabled = true` + `wakeword_script = ""` means push-to-talk enters, immediately reads stdin EOF, exits 0, and systemd's `Restart=always` loops it forever (the issue body shows restart counter at 344). Fix is the right shape: gate voice-mode entry on `voice_requested && (stdin_interactive || wakeword_available)` so a daemon-mode `genie-core` with no wakeword cleanly falls through to the HTTP API.

The follow-up refactor commit (`9ab3e71`) is a clean improvement on the original (`0847541`) — extracts the decision into a proper `runtime_mode` module with a `StartupDecision { mode, reason }` shape, so the four startup paths each get a named `StartupReason` variant:

- `VoiceNotRequested` → HTTP API
- `InteractivePushToTalk` → voice loop (terminal)
- `WakewordDaemon` → voice loop (daemon with wakeword)
- `PushToTalkNeedsTerminal` → HTTP API (the new fallback path)

That gives observability and future logging a real handle on *why* a particular startup path was selected, not just a boolean. The wakeword-before-interactive ordering is the right precedence — if an operator runs an interactive `genie-core` with a wakeword script configured, they want wakeword mode, not push-to-talk. Same outcome (voice mode), more accurate reason.

`blocked_push_to_talk()` is a precise predicate for "we tried, we couldn't" — used to fire the single warn log only on the actual stdin-EOF-would-have-killed-us path, not on the `voice_requested = false` quiet path. Right choice.

The four `runtime_mode::tests` cover the 2×2 matrix of `(interactive, wakeword_available)` for `voice_requested = true` plus the `voice_requested = false` short-circuit — that's exhaustive for a 3-bool input space.

`wakeword_available` is checked as `!config.core.wakeword_script.as_os_str().is_empty() && config.core.wakeword_script.exists()` — both presence in config AND on disk. Catches the case where the operator sets a wakeword path but the script never got deployed; we don't false-claim "wakeword daemon mode" then crash because the script is missing.

The `#[cfg(not(feature = "voice"))]` branch keeps `let _ = startup_decision;` so a chat-only build stays clippy-clean. That's exactly the kind of small thing #62's no-default-features CI job would have caught if missed — and that job is green here.

Contributor's Jetson acceptance test (in the PR comment) is the smoking gun: service goes from `activating (auto-restart)` → `active (running)`, the expected `WARN push-to-talk voice mode requested … running HTTP API only` line appears once in journalctl, and `genie-ctl status` reports `Core: ok`. End-to-end verified on the same hardware the bug was reported against.

All 5 CI checks green on `9ab3e71` (fmt, clippy, test, aarch64 cross-compile, `--no-default-features` axis). Going in.

ai-hpc:
Merged at `2c945a9bbff9a967d072892dc12358e1fc0440cf`.
Thanks to @carlos4s!

PR #74, opened 2026-05-18
core: shape genie-ai-runtime chat requests

### LINKED ISSUE #73: genie-core chat fails through genie-ai-runtime while health reports connected
## Context

After PR #70 and PR #72 were merged and deployed, `genie-core` now stays up under systemd and the web/dashboard surfaces load. However, the web control UI still cannot get a chat reply through `genie-core`.

`genie-core` health reports the LLM as connected:

```bash
curl -sf http://127.0.0.1:3000/api/health
```

```json
{
  "status":"ok",
  "llm":"connected",
  "llm_backend":"genie-ai-runtime",
  "model_family":"Qwen"
}
```

`genie-ctl status` also reports core is up:

```text
Governor:  night_a mode
Memory:    3093 MB available
Uptime:    1h 4m
Core:      ok
Radio:     disabled
WebSearch: enabled (duckduckgo)
```

The dashboard at port 3080 loads, but its Services table reports:

```text
core  Healthy  80ms
llm   Connection refused (os error 111)  0ms
```

And a direct `genie-core` chat request through the web/API path fails:

```bash
curl -s http://127.0.0.1:3000/api/chat \
  -H "Content-Type: application/json" \
  -H "X-Genie-Origin: dashboard" \
  -d '{"message":"Say hello from the GeniePod web UI in one short sentence."}'
```

Response:

```json
{"error":"chat: genie-ai-runtime 0: "}
```

## Why this is likely a GenieClaw issue

The `genie-ai-runtime` service itself had previously been validated directly:

- `/health` on port 8080 returned `status: ok`
- `/v1/chat/completions` returned a valid answer: `Hello from GeniePod!`
- Qwen3 model weights stayed warm after issue #69 / PR #70

So the failure appears to be in the GenieClaw integration path between:

- `genie-core /api/chat`
- `genie-core`'s `genie-ai-runtime` OpenAI-compatible adapter
- health/dashboard service checks for the LLM backend

## Suspicions to investigate

- `genie-core` health may check one endpoint successfully while chat posts to a different URL/path or port.
- The `genie-ai-runtime` adapter may not parse the runtime's OpenAI-compatible response correctly and may surface an empty status/body as `genie-ai-runtime 0: `.
- `genie-health` / dashboard may still be checking the old llama.cpp service or a stale URL, causing `llm Connection refused` even when the active backend is `genie-ai-runtime`.
- There may be a mismatch between `[services.llm]` config, `genie-core` adapter URL construction, and dashboard health probe logic.

## Acceptance criteria

- `curl http://127.0.0.1:3000/api/chat` returns a normal assistant response when `[services.llm].backend = "genie_ai_runtime"`.
- The port 3000 web chat UI displays the assistant reply.
- The port 3080 dashboard no longer shows `llm Connection refused` when the active LLM backend is `genie-ai-runtime` and healthy.
- `genie-core /api/health`, `genie-ctl status`, dashboard service health, and direct runtime `/health` agree on the active LLM backend state.

carlos4s:
Closes #73

## What changed
- Add a `genie-ai-runtime` request profile instead of treating every OpenAI-compatible backend like llama.cpp.
- Send `model: "jetson-llm"` and `think: false` to match the Jetson runtime path that works from direct curl.
- Compact oversized Genie prompt payloads for the Jetson runtime to a concise local-assistant system prompt plus the recent non-system turns, so the web UI does not send the full tool/system prompt into the small Qwen3 runtime context.
- Surface closed-before-status backend failures as a useful error instead of `genie-ai-runtime 0:`.

## Tests
- `cargo fmt --check`
- `cargo test -p genie-core llm::openai_compat::tests`
- `RUSTFLAGS='-D warnings' cargo clippy -p genie-core --all-targets --locked -- -D warnings`

## Note
- `cargo test -p genie-core` still hits the existing `binary_size_budget` failure on current `origin/main` (`genie-core: 5.06 MB` vs 5 MB budget), so I did not hide that unrelated baseline issue in this PR.

### DIFF
--- a/crates/genie-core/src/llm/genie_ai_runtime.rs
+++ b/crates/genie-core/src/llm/genie_ai_runtime.rs
@@ -1,7 +1,7 @@
 use anyhow::Result;
 use async_trait::async_trait;
 
-use super::openai_compat::OpenAiCompatClient;
+use super::openai_compat::{OpenAiCompatClient, RequestProfile};
 use super::{LlmBackendClient, Message, ResponseFormat};
 
 /// Adapter for the `genie-ai-runtime` OpenAI-compatible chat API surface.
@@ -12,13 +12,22 @@ pub struct GenieAiRuntimeBackend {
 impl GenieAiRuntimeBackend {
     pub fn new(host: &str, port: u16) -> Self {
         Self {
-            inner: OpenAiCompatClient::new("genie-ai-runtime", host, port),
+            inner: OpenAiCompatClient::new_with_profile(
+                "genie-ai-runtime",
+                host,
+                port,
+                RequestProfile::genie_ai_runtime(),
+            ),
         }
     }
 
     pub fn from_url(url: &str) -> Self {
         Self {
-            inner: OpenAiCompatClient::from_url("genie-ai-runtime", url),
+            inner: OpenAiCompatClient::from_url_with_profile(
+                "genie-ai-runtime",
+                url,
+                RequestProfile::genie_ai_runtime(),
+            ),
         }
     }
 }
--- a/crates/genie-core/src/llm/openai_compat.rs
+++ b/crates/genie-core/src/llm/openai_compat.rs
@@ -172,7 +172,7 @@ impl RequestProfile {
     }
 }
 
-const GENIE_RUNTIME_MAX_BODY_BYTES: usize = 12 * 1024;
+const GENIE_RUNTIME_MAX_BODY_BYTES: usize = 4 * 1024;
 const GENIE_RUNTIME_COMPACT_SYSTEM: &str =
     "You are GeniePod Home. Answer the user's latest request directly and concisely.";
 
--- a/crates/genie-core/src/llm/openai_compat.rs
+++ b/crates/genie-core/src/llm/openai_compat.rs
@@ -165,33 +165,14 @@ impl RequestProfile {
     }
 
     fn compact_messages(&self, messages: &[Message]) -> Vec<Message> {
-        let mut compacted = Vec::new();
-        if matches!(self, Self::GenieAiRuntime) {
-            compacted.push(Message {
-                role: "system".into(),
-                content: GENIE_RUNTIME_COMPACT_SYSTEM.into(),
-            });
+        match self {
+            Self::Generic => messages.to_vec(),
+            Self::GenieAiRuntime => compact_genie_runtime_messages(messages),
         }
-
-        let recent = recent_non_system_messages(messages, GENIE_RUNTIME_RECENT_MESSAGES);
-        if recent.is_empty() {
-            compacted.extend(messages.iter().filter(|m| m.role != "system").cloned());
-        } else {
-            compacted.extend(recent);
-        }
-
-        if compacted.is_empty()
-            && let Some(last) = messages.last()
-        {
-            compacted.push(last.clone());
-        }
-
-        compacted
     }
 }
 
 const GENIE_RUNTIME_MAX_BODY_BYTES: usize = 12 * 1024;
-const GENIE_RUNTIME_RECENT_MESSAGES: usize = 4;
 const GENIE_RUNTIME_COMPACT_SYSTEM: &str =
     "You are GeniePod Home. Answer the user's latest request directly and concisely.";
 
@@ -526,20 +507,20 @@ fn parse_status_line(line: &str) -> u16 {
         .unwrap_or(0)
 }
 
-fn recent_non_system_messages(messages: &[Message], limit: usize) -> Vec<Message> {
-    if limit == 0 {
-        return Vec::new();
-    }
-
-    let mut recent = messages
+fn compact_genie_runtime_messages(messages: &[Message]) -> Vec<Message> {
+    let Some(message) = messages
         .iter()
         .rev()
-        .filter(|m| m.role != "system")
-        .take(limit)
-        .cloned()
-        .collect::<Vec<_>>();
-    recent.reverse();
-    recent
+        .find(|m| m.role == "user")
+        .or_else(|| messages.iter().rev().find(|m| m.role != "system"))
+    else {
+        return Vec::new();
+    };
+
+    vec![Message {
+        role: "user".into(),
+        content: format!("{}\n\n{}", GENIE_RUNTIME_COMPACT_SYSTEM, message.content),
+    }]
 }
 
 fn should_retry_without_system_role(messages: &[Message], err: &str) -> bool {
@@ -687,7 +668,7 @@ mod tests {
             },
             Message {
                 role: "assistant".into(),
-                content: "older assistant turn".into(),
+                content: "older assistant turn ".repeat(200),
             },
             Message {
                 role: "user".into(),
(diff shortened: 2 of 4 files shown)


carlos4s:
Jetson smoke test recipe:

```bash
git fetch https://github.com/carlos4s/genie-claw.git fix/issue-73-genie-runtime-chat-health
git switch -C test/issue-73 FETCH_HEAD

make jetson deploy-binaries JETSON_HOST=192.168.55.1 JETSON_USER=aihpc

ssh aihpc@192.168.55.1 '\n  sudo systemctl restart genie-ai-runtime\n  for i in $(seq 1 120); do\n    curl -sf http://127.0.0.1:8080/health >/dev/null && break\n    sleep 1\n  done\n\n  sudo systemctl restart genie-core genie-health genie-api\n  sleep 3\n\n  curl -sf http://127.0.0.1:3000/api/health && echo\n  curl -s http://127.0.0.1:3000/api/chat \\\n    -H "Content-Type: application/json" \\\n    -H "X-Genie-Origin: dashboard" \\\n    -d '\\{"message":"Say hello from the GeniePod web UI in one short sentence."}'\\\n  echo\n  curl -sf http://127.0.0.1:3080/api/status && echo\n  /opt/geniepod/bin/genie-ctl status\n'\n```

carlos4s:
Updated after Jetson test still returned `closed connection before HTTP status (request_bytes=6506)`.

Second commit tightens the `genie-ai-runtime` compact path to match the direct curl shape that works against `jetson-llm-server`:
- sends one `user` message only
- folds the concise GeniePod instruction into that user message
- keeps only the latest user turn, dropping old assistant/history payload from the runtime request

Local checks after update:
- `cargo fmt --check`
- `cargo test -p genie-core llm::openai_compat::tests`
- `RUSTFLAGS='-D warnings' cargo clippy -p genie-core --all-targets --locked -- -D warnings`\n\nRetest on Jetson with the same commands, starting from:\n\n```bash\ngit fetch https://github.com/carlos4s/genie-claw.git fix/issue-73-genie-runtime-chat-health\ngit switch -C test/issue-73 FETCH_HEAD\nmake jetson deploy-binaries JETSON_HOST=192.168.55.1 JETSON_USER=aihpc\n```

carlos4s:
Updated again after the runtime-side fix showed the remaining product issue: increasing context alone avoids crashes, but full GenieClaw web prompts are still slow and may exceed small auto-context values under memory pressure.

New commit `5c55f62` does two things:
- lowers the `genie-ai-runtime` request compaction threshold from 12 KiB to 4 KiB, so the observed 6.6 KiB dashboard request takes the tiny user-only path instead of sending the full system/tool prompt to runtime
- starts `genie-ai-runtime.service` with `-c 2048 --int8-kv`, giving the service enough context for normal web-chat requests while keeping KV memory low

Checks:
- `cargo fmt --check`
- `cargo test -p genie-core llm::openai_compat::tests`
- `RUSTFLAGS='-D warnings' cargo clippy -p genie-core --all-targets --locked -- -D warnings`\n\nJetson retest should deploy both binaries and systemd units now:\n\n```bash\ngit fetch https://github.com/carlos4s/genie-claw.git fix/issue-73-genie-runtime-chat-health\ngit switch -C test/issue-73 FETCH_HEAD\nmake jetson deploy-binaries deploy-systemd JETSON_HOST=192.168.55.1 JETSON_USER=aihpc\nssh aihpc@192.168.55.1 'sudo systemctl daemon-reload && sudo systemctl restart genie-ai-runtime genie-core genie-health genie-api'\n```

ai-hpc (approved):
Closes the main user-facing symptom from #73 — `curl /api/chat` returns `{"error":"chat: genie-ai-runtime 0: "}` because the GenieClaw web chat ships a multi-kB prompt (tool manifest + memory + history + user turn) into a Jetson runtime whose Qwen3 context auto-capped to 327 tokens under memory pressure. The runtime then crashed with SIGSEGV (separately fixed by genie-ai-runtime PR #81's prompt-size guard, just merged). This PR sidesteps the problem on the client side via a request profile + compaction; the two layers reinforce each other.

The `RequestProfile { Generic, GenieAiRuntime }` shape is the right abstraction. Generic stays byte-identical to today's `OpenAiCompatClient` so the llama.cpp path is unaffected — `generic_request_profile_keeps_full_default_body` test pins that. GenieAiRuntime adds three deltas:

1. `model: "jetson-llm"` instead of `"default"` — matches the model name the Jetson runtime registers (and what the warmup unit posts).
2. `think: false` — matches the working direct-curl baseline. The `#[serde(skip_serializing_if = "Option::is_none")]` keeps llama.cpp from seeing the new field, so no regression.
3. Compaction when body > 4 KB.

The error path additions land before the compaction does its work — `status == 0 && body.is_empty()` now reports `{backend} closed connection before HTTP status (request_bytes=N)` instead of `genie-ai-runtime 0: `. Streaming path picks up the symmetric guard via `if !headers_done` + initial `status = 0` (was `200`, which silently passed through socket drops). Both are real diagnostic improvements that survive even if the compaction strategy needs revising later.

Three things to flag — none blocking, all probably worth follow-ups:

**1. Compaction is more aggressive than the PR body claims.** The PR description says "compact ... to a concise local-assistant system prompt plus the recent non-system turns", but `compact_genie_runtime_messages` actually replaces ALL messages with a single user-role turn containing only `{SYSTEM_BLURB}\n\n{latest_user_content}`. Conversation history is dropped wholesale, tool definitions are gone, memory recall context is gone. A web-chat user asking "what time is it?" then "and what's the weather?" will get the second answer with no memory of the first turn. The compaction test confirms this — exactly one message in the output. For unblocking the dashboard chat today, this is fine; for the longer-term web-chat UX, the actual "system + recent N turns" shape the PR body describes is probably what you want. Worth a follow-up.

**2. Issue #73's acceptance criterion #3 isn't addressed.** The issue says "Port 3080 dashboard no longer shows `llm Connection refused`", but the `llm Connection refused (os error 111)` line comes from `genie-health`'s separate probe path, not `genie-core`'s chat client. This PR doesn't touch `genie-health`. The chat path works after this lands; the dashboard's health-probe row stays red. Worth a separate issue for the genie-health probe URL / target update — possibly a Wants/After ordering thing on the dashboard checking the wrong unit, possibly a hardcoded `127.0.0.1:8080/v1/...` probe path.

**3. `GENIE_RUNTIME_MAX_BODY_BYTES = 4 * 1024` is hardcoded.** It pairs with `GENIEPOD_AI_RUNTIME_CONTEXT=2048` in the systemd unit (rough heuristic: 2 bytes per token average), but the relationship isn't enforced — if an operator raises `-c` to 4096 on a beefier Jetson, the GenieClaw side will keep compacting at 4 KB even though the runtime could handle ~8 KB. The clean fix is to probe runtime capacity at connection time, but the static threshold is defensible for an alpha — just leaves performance on the table on bigger boxes.

The systemd unit change (`--int8-kv` + `-c 2048` + `GENIEPOD_AI_RUNTIME_CONTEXT` env knob) is the right pairing for this PR: INT8 KV roughly doubles the fittable context vs FP16, and pinning to 2048 makes the auto-cap less likely to drop the context to 327 tokens under memory pressure. The test in `tool_dispatch_test.rs` pins both as service-file invariants so a future PR can't silently regress them.

The three unit tests on the compaction profile cover the right cases: Generic-unchanged, GenieAiRuntime-compacts-when-large, and the fallback-to-latest-non-system path for the (rare but real) case where the last message isn't a user role. The `should_retry_without_system_role` retry path that the existing client uses for llama.cpp still works because the compacted message is a single user turn (no system role to strip).

All 5 CI checks green on `5c55f62` (fmt, clippy, test, aarch64 cross-compile, `--no-default-features`). Going in — the immediate "web chat is broken" symptom gets fixed today, and the three rough edges above are clean targets for follow-up PRs / issues.

ai-hpc:
Merged at `8204be0fff0b0374e6eb62c272957c5c8f093f6d`.
Thanks to @carlos4s!

PR #76, opened 2026-05-18
setup: start genie-ai-runtime before memory-heavy services

### LINKED ISSUE #75: Evaluate raising genie-ai-runtime context default to 8192
## Summary

Jetson testing shows `genie-ai-runtime` can load a larger INT8-KV context when it starts before memory-heavy services. The current repo default is safer, but may be lower than needed for normal GeniePod runtime use.

## Jetson observations

Test setup:

- Stopped `genie-core`, `genie-health`, `genie-api`, `genie-whisper*`, `homeassistant`, `genie-ai-runtime*`
- Ran `sync; echo 3 > /proc/sys/vm/drop_caches`
- Started `genie-ai-runtime` with `--int8-kv` and systemd drop-in `GENIEPOD_AI_RUNTIME_CONTEXT=<CTX>`

Results:

- `CTX=4096`
  - Loaded actual `Context: 4096 tokens`
  - KV cache: `288 MB`
  - Runtime budget free: `1903 MB`
  - Direct runtime chat completed in `0.722s`

- `CTX=6144`
  - Loaded actual `Context: 6144 tokens`
  - KV cache: `432 MB`
  - Runtime budget free: `992 MB`

- `CTX=8192`
  - Loaded actual `Context: 8192 tokens`
  - KV cache: `576 MB`
  - Runtime budget free: `1616 MB`
  - Direct runtime chat completed in `0.725s`

Important contrast:

- When `CTX=4096` was tested while the full stack had already been running, runtime only loaded `Context: 1735 tokens (user-requested, memory cap 1735)`.
- This means startup order and resident service memory materially affect the runtime's loaded context.

## Proposed work

Evaluate whether GeniePod should default `GENIEPOD_AI_RUNTIME_CONTEXT` to `8192` for the Jetson Orin Nano 8 GB profile, while ensuring `genie-ai-runtime` starts early enough to claim the intended KV cache before memory-heavy services load.

## Acceptance criteria

- Decide whether the default should remain `2048`, move to `4096`, or move to `8192`.
- If increasing the default, verify full-stack startup still keeps:
  - `genie-ai-runtime` healthy
  - `genie-core` healthy
  - dashboard chat working through `:3000`
  - voice/whisper services able to start after runtime
- Document the reason for the chosen default in the systemd unit or setup notes.

## Notes

This does not replace prompt compaction for `genie-ai-runtime`. Larger context prevents rejection/capping, but full system/tool prompts can still create slow prefill and poor web UI latency.

carlos4s:
## Summary

- raise the default `GENIEPOD_AI_RUNTIME_CONTEXT` to the Jetson-tested `8192` with `--int8-kv`
- order `genie-ai-runtime.service` before Whisper, Home Assistant, and `genie-core` so runtime claims KV cache before memory-heavy services at boot
- reorder `start_all.sh` so it starts the configured LLM before Home Assistant and Whisper without stopping those services manually

Fixes #75.

## Verification

- `cargo fmt --check`
- `cargo test -p genie-core --test tool_dispatch_test start_all_uses_configured_llm_backend`
- `bash -n deploy/scripts/start_all.sh`
- `git diff --check`

## Jetson context

Issue #75 captured Jetson testing where `8192` loaded with INT8 KV when `genie-ai-runtime` started before memory-heavy services. This PR preserves that condition through systemd ordering and the lifecycle script, instead of requiring the operator to manually stop the stack first.

### DIFF
--- a/crates/genie-core/tests/tool_dispatch_test.rs
+++ b/crates/genie-core/tests/tool_dispatch_test.rs
@@ -264,6 +264,24 @@ fn start_all_uses_configured_llm_backend() {
         contents.contains("is_warmup_unit") && contents.contains("start --no-block"),
         "start_all should queue warmup units without blocking the lifecycle script"
     );
+    let units = contents
+        .split("UNITS=(")
+        .nth(1)
+        .and_then(|s| s.split(")").next())
+        .expect("start_all should declare ordered units");
+    let llm_pos = units
+        .find("\"$configured_llm_unit\"")
+        .expect("start_all should include the configured LLM unit");
+    let homeassistant_pos = units
+        .find("homeassistant.service")
+        .expect("start_all should include Home Assistant");
+    let whisper_pos = units
+        .find("genie-whisper.service")
+        .expect("start_all should include Whisper");
+    assert!(
+        llm_pos < homeassistant_pos && llm_pos < whisper_pos,
+        "start_all should start the configured LLM before memory-heavy services"
+    );
 }
 
 /// Verify genie-ai-runtime service preserves warm GGUF pages across restarts.
@@ -285,8 +303,14 @@ fn genie_ai_runtime_service_preserves_model_page_cache() {
         "genie-ai-runtime.service should use INT8 KV to fit enough context under memory pressure"
     );
     assert!(
-        contents.contains("GENIEPOD_AI_RUNTIME_CONTEXT=2048"),
-        "genie-ai-runtime.service should request the GenieClaw web-chat context size"
+        contents.contains("GENIEPOD_AI_RUNTIME_CONTEXT=8192"),
+        "genie-ai-runtime.service should request the Jetson-tested 8k context size"
+    );
+    assert!(
+        contents.contains(
+            "Before=genie-whisper.service genie-whisper-warmup.service homeassistant.service genie-core.service"
+        ),
+        "genie-ai-runtime.service should reserve KV cache before memory-heavy services"
     );
 }
 
--- a/deploy/scripts/start_all.sh
+++ b/deploy/scripts/start_all.sh
@@ -118,12 +118,12 @@ configured_llm_unit="$(normalize_unit "$raw_llm_unit")"
 configured_warmup_unit="$(warmup_unit_for "$configured_llm_unit")"
 
 UNITS=(
-    homeassistant.service
     genie-audio.service
-    genie-whisper.service
-    genie-whisper-warmup.service
     "$configured_llm_unit"
     "$configured_warmup_unit"
+    homeassistant.service
+    genie-whisper.service
+    genie-whisper-warmup.service
     genie-core.service
     genie-governor.service
     genie-health.service
--- a/deploy/systemd/genie-ai-runtime.service
+++ b/deploy/systemd/genie-ai-runtime.service
@@ -2,6 +2,11 @@
 Description=GeniePod AI Runtime (Jetson-tuned LLM, OpenAI-compatible)
 Documentation=https://github.com/GeniePod/genie-ai-runtime
 After=network.target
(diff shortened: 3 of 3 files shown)


ai-hpc (approved):
Right shape for the right problem. Issue #75's evidence is the smoking gun: same `-c 4096` request fits only ~1.7k ctx after the full stack is resident, but fits 4k / 6k / 8k cleanly when `genie-ai-runtime` loads first. The runtime's `-c` is clamped by available memory at load time, so startup *order* materially changes the realized KV-cache capacity.

Three coordinated changes:

1. **`Before=genie-whisper.service genie-whisper-warmup.service homeassistant.service genie-core.service`** on the unit. systemd ordering directive that makes the LLM unit's load happen before the other memory-heavy services claim DRAM. Defense in depth with PR #72's `After=genie-ai-runtime.service genie-llm.service` on `genie-core` — both sides now agree, and the LLM-vs-whisper / LLM-vs-HA dependencies are also covered. `Before=` against a unit that isn't installed is a no-op, so this is safe on installs that don't ship homeassistant or whisper.

2. **`GENIEPOD_AI_RUNTIME_CONTEXT=2048 → 8192`**. Matches the largest context the issue verified loads cleanly on Orin Nano 8 GB. The env knob stays — operators on smaller Jetsons (Nano 4 GB) can override via systemd drop-in without touching the unit file.

3. **`start_all.sh` reorders** so the configured LLM unit + warmup run before homeassistant + whisper in the manual lifecycle path too, mirroring the systemd `Before=` for operators who use the script directly.

Tests added to `tool_dispatch_test.rs` lock both invariants:

- `start_all_uses_configured_llm_backend` extension uses substring-position comparison on the `UNITS=(...)` array to assert `$configured_llm_unit` appears before `homeassistant.service` and `genie-whisper.service`. Substring-position is fragile (a future rename of the `$configured_llm_unit` symbol breaks the test), but the assertion message is precise enough that the failure mode points at the right line.
- `genie_ai_runtime_service_preserves_model_page_cache` extension asserts `GENIEPOD_AI_RUNTIME_CONTEXT=8192` and that the new `Before=` clause is present verbatim.

Worth flagging for a follow-up, not blocking:

- **PR #74's `GENIE_RUNTIME_MAX_BODY_BYTES = 4 * 1024` body-compaction threshold is now mismatched with the new context.** Compaction triggers at 4 KB of body bytes regardless of the runtime's actual capacity, so with 8192-token runtime context (which can comfortably hold ~16-24 KB of typical English chat text), the client will still aggressively compact prompts the runtime could now handle. Same observation I left on #74. Right path is to make the threshold a function of `GENIEPOD_AI_RUNTIME_CONTEXT` or to probe runtime capacity at connection time. The static 4 KB ceiling is now leaving more performance on the table than it was at 2048-ctx.
- **Compatibility with PR #70 (warm page cache)** is preserved — `Before=` only affects boot-time ordering, not `systemctl restart` of the runtime alone, so the page-cache warmth across restart-of-runtime-only is intact.
- **Smaller-Jetson safety**: an Orin Nano 4 GB or earlier Jetson without `--int8-kv` support could fail to allocate 8192 ctx at boot. The unit's `Restart=on-failure` would loop in that case. Operators on those hosts will need a drop-in: `Environment=GENIEPOD_AI_RUNTIME_CONTEXT=2048` (or smaller). Worth a sentence in the next ROADMAP / GETTING_STARTED update if alpha.10 is targeting wider hardware.

All 7 CI checks green on `c1cae29` (fmt, clippy, test, aarch64 cross-compile, shellcheck, ruff, `--no-default-features`). End-user verified on Jetson per your test. Going in.

ai-hpc:
Merged at `9ab2974e6b0acd3b422576fb57715e71a2086d5b`.
Thanks to @carlos4s!

Issue #58, opened 2026-05-17
Optimize voice pipeline latency (umbrella)


ai-hpc:
## Goal

Drive end-to-end **first-reply latency** down from today's verified ~4 s baseline (Orin Nano + LyraT V4.3, Phi-4-mini Q4_K_M + whisper-small + Piper en_US-amy) toward a **conversational target of <1.5 s** speech-end → first audible audio.

This is an umbrella issue. The per-component issues already exist (#12, #25, #26); this one tracks the latency budget end-to-end, picks targets per phase, and adds the optimization opportunities that don't yet have their own issue.

## Current baseline (from README §Alpha.7 Verified Voice Cycle, 2026-05-13)

| Phase                          | Today  | Target | Notes |
|--------------------------------|--------|--------|-------|
| preprocess (DFN+sox)           | ~808 ms | ≤250 ms | per 3 s capture — see #12 |
| STT (whisper-small, warm)      |  285 ms |  ≤200 ms | warm only thanks to #18 |
| LLM → first sentence           |  ~2.5 s | ≤500 ms | dominant cost; depends on backend |
| TTS first synth                | ~1.2 s | ≤300 ms | Piper cold first-chunk |
| speech end → first audio (Σ)   | 3964 ms | **<1500 ms** | first-reply banner total |
| Total cycle                    | 15.4 s |  ≤5 s | includes user speaking |

The 5-phase banner from #19 / PR #20 is how we'll measure regressions on every change here.

### 1. LLM first-token latency (biggest single win)

The `STT done → first audio` span (3679 ms) is dominated by LLM warmup + first-sentence generation, not Piper.

- [ ] Land genie-ai-runtime v1.0.0 default flip (#52) and re-measure with KV-cache reuse on
- [ ] Persistent LLM server keepalive — confirm no per-request weight reload on idle Jetson
- [ ] Speculative decoding / draft model evaluation for Phi-4-mini and Qwen3-4B paths
- [ ] Prompt-prefix caching for the system prompt + tool schema (these are constant per session)

### 2. Streaming TTS (#26)

Today Piper waits for the whole LLM reply, then synthesizes. The single biggest UX win is to start TTS on the first complete sentence (or first ~12 tokens) while the LLM is still generating. Tracked separately in #26; this umbrella will block on it.

### 3. Capture preprocessing (#12)

DFN burns ~808 ms per 3 s window. Options:
[...]

Issue #17, opened 2026-05-12
First-cycle STT cold-start is 60-90s; add genie-whisper-warmup.service


ai-hpc:
## Symptom

First STT request after `genie-core` starts can take 60-92 seconds wall-clock, vs ~270 ms once warm. Logged times across multiple sessions on Jetson Orin Nano + LyraT:

- 09:41:26 — STT 286 ms (warm)
- 19:18:01 — STT **92,426 ms** (first cycle after fresh `genie-core` start)
- 09:59:05 — STT 60,176 ms (earlier session, same pattern)

The user-visible effect: after pressing Enter for the first time, the assistant appears unresponsive for ~1.5 minutes. Reproducible every fresh process start.

## Cause

`genie-whisper.service` keeps `whisper-server` (whisper.cpp HTTP) alive across `genie-core` restarts, so the binary is up. But the **whisper model is not actually loaded into iGPU memory until the first `/inference` POST hits it**. That first inference does the model load + warm-up GPU kernels in band, and the user pays the latency.

This is the same class of problem PR #7 (`genie-llm-warmup.service`) solved for `llama-server` / Phi-4-mini. We want the same fix for `whisper-server`.

## Proposed fix

A oneshot systemd unit `genie-whisper-warmup.service`:
- `After=genie-whisper.service`, `WantedBy=geniepod.target`
- Polls `127.0.0.1:8178` with `nc -z` until reachable (or 90 s timeout).
- Generates a 1-second silent mono 16 kHz WAV via `sox -n -r 16000 -c 1 /run/geniepod/whisper-warmup.wav trim 0.0 1.0`.
- POSTs it to `/inference` with `language=en` to match the production decoder path.
- Discards the response. Failures are non-fatal so a broken whisper does not block boot.

`setup-jetson.sh` enables the new unit alongside the existing warmup service. Unit shipped via `make deploy`.

## Acceptance criteria

- After `make deploy` + reboot, the **first** user-visible voice cycle's STT latency is comparable to subsequent cycles (~270-500 ms), not the 60-90 s cold path.
- The warmup unit's journal shows successful completion within ~10-30 s of boot.
- No regression on subsequent voice cycles.
- Headless / no-LyraT installs (no `whisper-server` binary) still boot — unit no-ops via `ConditionPathExists`.

## Related

- #7 — equivalent fix for `llama-server` / Phi-4-mini (already merged).
- #2 — earlier iGPU-contention work that established the GPU-stays-loaded discipline.

PR #78, opened 2026-05-18
fix(telegram): spawn handle_update so one chat does not stall others

### LINKED ISSUE #77: telegram: sequential handle_update stalls unrelated chats
## Background

`telegram::run` polls Telegram for updates, then awaits each handler **sequentially** in a tight loop ([crates/genie-core/src/telegram.rs:63-78](crates/genie-core/src/telegram.rs#L63-L78)):

```rust
loop {
    match api.get_updates(offset).await {
        Ok(updates) => {
            for update in updates {
                offset = offset.max(update.update_id.saturating_add(1));
                if let Err(e) = api.handle_update(update).await {
                    tracing::warn!(error = %e, "telegram update handling failed");
                }
            }
        }
        ...
    }
}
```

Since #42 / PR #64 landed, `handle_update` can take **seconds to tens of seconds** when a voice message arrives — `download → ffmpeg → whisper-server → /api/chat → reply` is a long critical path, and the first cycle includes whisper-server cold start (60–90 s per #17).

While that single handler is awaiting, the outer `for` loop cannot advance. Every other chat — text or voice — waits behind it. This is the same channel-of-1 hazard the `current_thread` tokio runtime ([crates/genie-core/src/main.rs:16](crates/genie-core/src/main.rs#L16)) makes easy to fall into.

## Why this matters now

Issue #42's Phase 1 acceptance criteria explicitly include:

> **AC #8:** Concurrent voice messages from multiple Telegram users are handled in parallel when STT server mode is in use.

PR #64 shipped the voice pipeline but the poll loop is unchanged, so AC #8 is **not** satisfied. The whisper-server we already run on `:8178` is designed for concurrent inference; the bottleneck is upstream of it, in the adapter.

## Reproduction

1. Configure `[telegram.voice].enabled = true` with `whisper_port > 0`.
2. From chat A, send a 30 s voice message. STT begins.
3. From chat B (separate user / chat_id), send a text message ("hi").
4. Observe: chat B's reply arrives **only after** chat A's full STT + LLM + reply cycle finishes, even though the LLM and whisper-server have spare capacity.

Worst case: chat A's whisper-server call is in its 60 s cold start and *every* chat is frozen until it returns.

## Proposed fix

Spawn each `handle_update` on the tokio runtime instead of awaiting inline:

```rust
let api = Arc::new(TelegramApi::new(client, config));
// ...
for update in updates {
    offset = offset.max(update.update_id.saturating_add(1));
    let api = Arc::clone(&api);
    tokio::spawn(async move {
        if let Err(e) = api.handle_update(update).await {
            tracing::warn!(error = %e, "telegram update handling failed");
        }
    });
}
```

This requires:

- `TelegramApi` becomes `Arc<TelegramApi>` (it's already `Send + Sync` modulo the `reqwest::Client`, which is `Clone + Send + Sync`).
- The `#[tokio::main(flavor = "current_thread")]` choice in `main.rs` is compatible with `tokio::spawn` for IO-bound work; if subprocess work becomes a contention point, revisit with a bounded `JoinSet` and a semaphore (see "Bounded concurrency" below).

### Bounded concurrency

Unbounded spawning could overload `ffmpeg` / whisper-server under burst. Recommend gating with a `Semaphore::new(N)` per-adapter, with `N` defaulting to a new `[telegram.voice].max_parallel_voice` config knob (default 2 or 3). Text-only updates need no gate.

## Acceptance criteria

- [ ] Sending a voice message from chat A and a text message from chat B at the same time: chat B's reply does not wait for chat A's STT.
- [ ] Two voice messages from different chats running concurrently both transcribe through whisper-server in parallel (visible in `whisper-server` log timestamps).
- [ ] Order-within-a-single-chat is preserved (or explicitly documented as unordered — see "Open question" below).
- [ ] Burst of N+1 voice messages with `max_parallel_voice = N` queues cleanly, no panics, no leaked tempfiles.

## Open question

Should `handle_update` for the **same** `chat_id` still serialize? Telegram itself delivers updates in order, but a user sending voice+text back-to-back might want the text reply to wait for the voice reply, or might prefer fastest-first. Default proposal: per-chat serialization via a `DashMap<chat_id, tokio::sync::Mutex<()>>` taken inside the spawned task. Worth a small design pass before implementation.

## Related

- #42 (voice MVP; AC #8)
- #58 (latency umbrella — this isn't a latency win for a single user but it is for concurrent users)
- #67 (now-fixed tempfile race that became *visible* once parallelism is enabled — sanity-check the fix under concurrent load as part of this work)
- #17 (whisper-server cold start makes the symptom much worse)

plind-junior:
## Summary

Fixes #77. The Telegram poll loop in [crates/genie-core/src/telegram.rs](crates/genie-core/src/telegram.rs) used to `await` each `handle_update` sequentially, so a single voice message stuck in `ffmpeg → whisper-server → /api/chat → reply` (tens of seconds, worse during whisper-server's 60–90 s cold start per #17) froze every other chat — text or voice — behind it. This left #42's **AC #8** (concurrent voice STT in parallel across chats) unmet after PR #64 shipped the voice pipeline.

## Changes

- **`crates/genie-core/src/telegram.rs`** — `TelegramApi` is now `Arc`-shared; `run` spawns each update with `tokio::spawn`, so a slow handler no longer blocks the next one. The adapter still works under `#[tokio::main(flavor = "current_thread")]` because the work is IO/process-bound.
- **Bounded voice concurrency** — added a `Semaphore` (`voice_permits`) sized from the new config knob, acquired *after* the cheap rejection branches in `handle_voice_message` so disabled-voice / over-length rejections don't take a permit they won't use. Clamps `max_parallel_voice = 0` to 1 to avoid deadlocking voice messages on a misconfig.
- **Per-chat ordering** — `chat_id`-keyed `tokio::sync::Mutex` map (`chat_locks`) serializes a single chat's updates so user-perceived reply order matches Telegram's delivery order, while different chats run in parallel. Unauthorized chats skip the map so flood traffic doesn't squat lock slots.
- **`crates/genie-common/src/config.rs`** — new `TelegramVoiceConfig.max_parallel_voice: usize` field with `#[serde(default = ...)]` returning **2** (enough to satisfy AC #8 on a Jetson Orin Nano-class host without overrunning ffmpeg).
- **`crates/genie-core/src/main.rs`** — plumb the new knob into `TelegramVoiceRuntimeConfig` and surface `voice_enabled` + `max_parallel_voice` in the adapter startup log so operators can confirm the configured concurrency.
- **`deploy/config/geniepod.toml`** — document the knob in the existing `[telegram.voice]` commented stanza.

## Acceptance criteria

- [x] Sending a voice from chat A and a text from chat B at the same time: chat B's reply no longer waits for chat A's STT — they run as independent spawned tasks holding distinct `chat_locks` entries.
- [x] Two voice messages from different chats both transcribe through whisper-server in parallel up to `max_parallel_voice` (default 2). New `voice_permits_enforce_max_parallel` test exercises permit acquisition and release.
- [x] Order within a single chat is preserved by the per-chat mutex. New `chat_lock_is_stable_per_chat_id` test verifies the same `chat_id` always maps to the same `Arc<Mutex>` and different ids do not.
- [x] Burst over the cap queues cleanly: extra voice tasks `await` on the semaphore rather than panicking; tempfiles are still cleaned via the existing `TempCleanup` Drop guard (no change to the #67 fix) and run inside the permitted scope.

## Real Behavior Proof                                                                                                                                                      
- [x] `cargo test -p genie-core --lib telegram::tests::` — all 15 tests pass after rebasing onto current `main` (merging the AtomicU64 nonce changes from PR #83).          
- [x] `cargo check -p genie-core` — compiles cleanly with both #77 and #83 changes combined.    
 
## Test plan

- [x] `cargo test -p genie-core --lib telegram` — 14 telegram tests pass (3 new: `chat_lock_is_stable_per_chat_id`, `voice_permits_enforce_max_parallel`, `voice_permits_clamp_zero_to_one`).
- [x] `cargo test --workspace` — full suite green.
- [x] `cargo check -p genie-core --no-default-features` — chat-only build axis from PR #62 still clean.
- [x] `cargo clippy -p genie-core --tests` — clean.
- [ ] Manual on-device: configure `[telegram.voice] enabled = true, max_parallel_voice = 2`, send a 30 s voice from chat A and a text "hi" from chat B; confirm B's reply does **not** wait for A's STT. Then send three voice messages from three chats; confirm the third queues until one of the first two finishes (visible in `whisper-server` log timestamps).

## Related

- Closes #77, satisfies #42 AC #8.
- Touches the area #67 fixed (tempfile races) — `TempCleanup` + per-task unique `pid-nonce` paths still hold under concurrent voice tasks.
- Helps the multi-user latency story tracked under #58.

Fixes #77

### DIFF
--- a/crates/genie-common/src/config.rs
+++ b/crates/genie-common/src/config.rs
@@ -490,6 +490,14 @@ pub struct TelegramVoiceConfig {
     /// Tuned for the 60–90 s of OGG/Opus that comfortably fits under 1 MB.
     #[serde(default = "defaults::telegram_voice_max_reply_chars")]
     pub max_reply_chars: usize,
+
+    /// Bound on concurrent voice pipelines (download → ffmpeg → whisper-server
+    /// → /api/chat) the adapter will run at once. Issue #77: the poll loop
+    /// now spawns each update, but unbounded fan-out under burst could
+    /// overload ffmpeg / whisper-server. Text-only updates are not gated by
+    /// this knob.
+    #[serde(default = "defaults::telegram_voice_max_parallel_voice")]
+    pub max_parallel_voice: usize,
 }
 
 impl Default for TelegramVoiceConfig {
@@ -501,6 +509,7 @@ impl Default for TelegramVoiceConfig {
             ffmpeg_path: defaults::telegram_voice_ffmpeg_path(),
             reply_as_voice: false,
             max_reply_chars: defaults::telegram_voice_max_reply_chars(),
+            max_parallel_voice: defaults::telegram_voice_max_parallel_voice(),
         }
     }
 }
@@ -1527,6 +1536,14 @@ mod defaults {
         // 1 MB sendVoice limit at Piper's typical OGG/Opus output rate.
         // Long-form replies fall back to text.
         800
+    }
+    pub fn telegram_voice_max_parallel_voice() -> usize {
+        // Two concurrent voice pipelines is enough to satisfy issue #77's
+        // AC #8 (two voice messages from different chats transcribe in
+        // parallel) while leaving headroom for ffmpeg / whisper-server on a
+        // Jetson Orin Nano-class device. Bump in deployment configs if the
+        // host has more CPU / a dedicated whisper-server.
+        2
     }
     pub fn web_search_enabled() -> bool {
         true
--- a/crates/genie-core/src/main.rs
+++ b/crates/genie-core/src/main.rs
@@ -325,6 +325,7 @@ async fn main() -> Result<()> {
                         max_reply_chars: config.telegram.voice.max_reply_chars,
                         piper_path: config.core.piper_path.clone(),
                         piper_model: config.core.piper_model.clone(),
+                        max_parallel_voice: config.telegram.voice.max_parallel_voice,
                     },
                 };
 
@@ -332,6 +333,8 @@ async fn main() -> Result<()> {
                     poll_timeout_secs = telegram_cfg.poll_timeout_secs,
                     allowed_chats = telegram_cfg.allowed_chat_ids.len(),
                     allow_all_chats = telegram_cfg.allow_all_chats,
+                    voice_enabled = telegram_cfg.voice.enabled,
+                    max_parallel_voice = telegram_cfg.voice.max_parallel_voice,
                     "starting Telegram adapter"
                 );
 
--- a/crates/genie-core/src/telegram.rs
+++ b/crates/genie-core/src/telegram.rs
@@ -143,10 +143,7 @@ impl TelegramApi {
     /// critical section is just a HashMap insert/clone — no `.await` is held
     /// across it.
     fn chat_lock(&self, chat_id: i64) -> Arc<AsyncMutex<()>> {
-        let mut locks = self
-            .chat_locks
-            .lock()
-            .unwrap_or_else(|e| e.into_inner());
+        let mut locks = self.chat_locks.lock().unwrap_or_else(|e| e.into_inner());
         locks
             .entry(chat_id)
             .or_insert_with(|| Arc::new(AsyncMutex::new(())))
--- a/crates/genie-core/src/telegram.rs
+++ b/crates/genie-core/src/telegram.rs
@@ -1,10 +1,13 @@
 use anyhow::{Context, Result};
 use reqwest::Client;
 use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
 use std::path::PathBuf;
 use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::{Arc, Mutex as StdMutex};
 use std::time::Duration;
 use tokio::process::Command;
+use tokio::sync::{Mutex as AsyncMutex, Semaphore};
 
 const TELEGRAM_MAX_MESSAGE_LEN: usize = 4096;
 
@@ -56,6 +59,11 @@ pub struct TelegramVoiceRuntimeConfig {
     pub max_reply_chars: usize,
     pub piper_path: PathBuf,
     pub piper_model: PathBuf,
+    /// Bound on concurrent voice pipelines (issue #77). The poll loop spawns
+    /// every update; this caps the heavyweight STT path so a burst of voice
+    /// messages doesn't overload ffmpeg / whisper-server. Text-only updates
+    /// are not gated.
+    pub max_parallel_voice: usize,
 }
 
 pub async fn run(config: TelegramRuntimeConfig) -> Result<()> {
@@ -67,7 +75,7 @@ pub async fn run(config: TelegramRuntimeConfig) -> Result<()> {
         .build()
         .context("failed to build Telegram HTTP client")?;
 
-    let api = TelegramApi::new(client, config);
+    let api = Arc::new(TelegramApi::new(client, config));
     let mut offset = match api.bootstrap_offset().await {
         Ok(offset) => offset,
         Err(e) => {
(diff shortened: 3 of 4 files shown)


ai-hpc (changes requested):
Code is solid and addresses #77 the right way:

- **`tokio::spawn` per update** in the poll loop so a slow voice pipeline can't block unrelated chats.
- **Per-chat `Arc<AsyncMutex<()>>` map** preserves reply ordering within a single chat while letting different chats run in parallel. Critical-section is just a `HashMap` insert/clone (no `.await` held), so `std::sync::Mutex` around the lookup map is correct.
- **`Semaphore::new(max_parallel_voice)` permit gate** bounds the heavyweight ffmpeg + whisper-server pipelines under burst load. Permit acquired *after* the cheap rejection branches (unauthorized chat, oversized message), so rejection paths don't squat on a permit.
- **`max_parallel_voice` floored at 1** in the constructor so a misconfigured `0` doesn't deadlock — just serializes.
- **Three targeted tests**: `chat_lock_is_stable_per_chat_id`, `voice_permits_enforce_max_parallel`, `voice_permits_clamp_zero_to_one`. Exactly the right invariants to pin.
- **Config knob `[telegram.voice].max_parallel_voice`** with sensible default `2` and a clear comment about when to bump it.

Quality is high enough that I'd merge as-is — except for one mechanical blocker:

### Blocker: rebase against current `main` needed

PR #83 (`fix(telegram): AtomicU64 nonce …`, merged at `a05afa5`) replaced `rand_nonce()` with `next_temp_nonce()` and removed the `rand_nonce` helper entirely. This PR's diff still calls `rand_nonce()` at `handle_voice_message`, so the merge is CONFLICTING / DIRTY against current `main`. The resolution is mechanical (a few lines), but I'm not pushing rebases to contributor forks in this pass — that's on you.

Suggested local fix:

```bash
git fetch origin
git rebase origin/main
# Conflicts in crates/genie-core/src/telegram.rs:
#   - merge the imports: keep BOTH `std::sync::atomic::{AtomicU64, Ordering}`
#     (from #83) AND `std::sync::{Arc, Mutex as StdMutex}` (from this PR).
#   - in `handle_voice_message`, replace `rand_nonce()` with `next_temp_nonce()`.
#   - in `mod tests`, combine the import lists: keep BOTH `next_temp_nonce`
#     (for the existing concurrent-uniqueness test from #83) AND
#     `TelegramApi, TelegramRuntimeConfig, TelegramVoiceRuntimeConfig`
#     (for `test_api` in this PR's tests).
# Run `cargo test -p genie-core --lib telegram::tests::` to confirm both
# test sets pass after the merge.
git push --force-with-lease
```

Once the rebase lands and CI clears green on the new head, I'll re-approve and merge.

### Worth flagging, not blocking

- **`chat_locks` map is per-process lifetime.** The comment already calls this out for `allow_all_chats = true` deployments. For an allowlisted bot it's bounded by the allowlist size; for the open-bot case it grows monotonically with the number of distinct chats. Not a correctness issue, just memory you should know about. A future PR could prune entries with no contention after some idle window if it ever matters.
- **Unauthorized chats skip the chat_lock acquire.** Smart small detail — a flood of unauthorized traffic doesn't squat on a lock slot. Worth keeping if anyone refactors this path.

All 5 CI checks green on the previous head (`6225417`), commit hygiene clean (proper noreply email, no AI co-author trailers). Going in as soon as the rebase clears.

ai-hpc (approved):
Re-reviewed after branch refresh. The previous nonce/rebase blocker is resolved, the branch now merges cleanly with current main, and the channel-concurrency fix is scoped to M2. CI and aarch64 cross-compile are green, so this is ready to merge.

ai-hpc:
merged at 5b23ae36e0514b28e785d56a410bf22e10c066ba

Thanks @plind-junior

PR #84, opened 2026-05-18
dashboard: show full systemd service status

### LINKED ISSUE #80: Dashboard services table only shows core and llm
## Summary

The system dashboard on port `3080` does not show the full running service set. It currently reports only `core` and `llm` in the Services table, even when the broader GeniePod stack is running.

## Current behavior

- Open `http://<jetson>:3080`
- Services table shows only:
  - `core`
  - `llm`

## Expected behavior

The dashboard should show all relevant GeniePod services and their current status/latency where available, including services such as:

- `genie-ai-runtime` / configured LLM backend
- `genie-core`
- `genie-api`
- `genie-health`
- `genie-governor`
- `genie-mqtt`
- `genie-audio`
- `genie-whisper`
- `genie-wakeword`
- `homeassistant` when installed/enabled

## Context

After the runtime/startup fixes, the stack can be healthy and running, but the `3080` dashboard still gives an incomplete service view. This makes it harder to verify startup and diagnose missing services from the UI.

## Acceptance criteria

- Dashboard Services table includes the deployed GeniePod service set, not only `core` and `llm`
- Missing, disabled, inactive, failed, and healthy services are distinguishable
- The UI remains usable when optional services are not installed

carlos4s:
## Summary

- expands `/api/services` beyond URL health rows by merging health monitor data with systemd state for the deployed GeniePod units
- includes core, configured LLM, API, health, governor, MQTT, audio, Whisper, wakeword, Home Assistant, and configured optional media/cloud services
- updates the dashboard services table to show the systemd unit under each service and distinguish healthy, inactive, failed, missing, and unknown states

Fixes #80.

## Verification

- `cargo fmt --check`
- `cargo test -p genie-api`
- `cargo test -p genie-core --test tool_dispatch_test chat_ui_uses_animated_writing_indicator`
- `RUSTFLAGS='-D warnings' cargo clippy -p genie-api --all-targets --locked -- -D warnings`
- `git diff --check`

## Jetson

Not run from this environment: SSH to `aihpc@192.168.55.1` timed out before deploy/test, so this PR only claims local validation.

### DIFF
--- a/crates/dashboard/dashboard.js
+++ b/crates/dashboard/dashboard.js
@@ -136,6 +136,20 @@ async function pollTegrastats() {
   }
 }
 
+function serviceStatus(s) {
+  if (s.healthy) return { label: 'Healthy', color: 'var(--green)', dot: 'dot-up' };
+  if (s.load_state === 'not-found') return { label: 'Missing', color: 'var(--text2)', dot: 'dot-unknown' };
+  if (s.active_state === 'failed') return { label: 'Failed', color: 'var(--red)', dot: 'dot-down' };
+  if (s.active_state && s.active_state !== 'active') {
+    return {
+      label: s.sub_state || s.active_state,
+      color: 'var(--yellow)',
+      dot: 'dot-unknown',
+    };
+  }
+  return { label: s.error || 'Unknown', color: 'var(--red)', dot: 'dot-down' };
+}
+
 async function pollServices() {
   const data = await fetchJson('/api/services');
   const tbody = document.getElementById('services-body');
@@ -147,13 +161,13 @@ async function pollServices() {
   }
 
   tbody.innerHTML = data.map(s => {
-    const dotClass = s.healthy ? 'dot-up' : 'dot-down';
-    const status = s.healthy ? 'Healthy' : (s.error || 'Down');
-    const statusColor = s.healthy ? 'var(--green)' : 'var(--red)';
+    const status = serviceStatus(s);
+    const latency = s.response_ms == null ? '--' : `${s.response_ms}ms`;
+    const detail = s.unit ? `<div class="small">${escapeHtml(s.unit)}</div>` : '';
     return `<tr>
-      <td><span class="dot ${dotClass}"></span>${s.service}</td>
-      <td style="color:${statusColor}">${status}</td>
-      <td>${s.response_ms}ms</td>
+      <td><span class="dot ${status.dot}"></span>${escapeHtml(s.service)}${detail}</td>
+      <td style="color:${status.color}">${escapeHtml(status.label)}</td>
+      <td>${latency}</td>
     </tr>`;
   }).join('');
 }
--- a/crates/genie-api/src/routes.rs
+++ b/crates/genie-api/src/routes.rs
@@ -1,5 +1,8 @@
 use genie_common::config::Config;
 use genie_common::tegrastats;
+use serde::Serialize;
+use std::collections::{BTreeMap, BTreeSet};
+use tokio::process::Command;
 
 use crate::http::Response;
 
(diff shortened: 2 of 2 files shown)


ai-hpc (approved):
Right shape for the right gap. The old `/api/services` was a thin wrapper over `genie-health`'s URL-poll table, which only ever populated `core` and `llm` because those are the only services with HTTP health endpoints. This PR keeps that data source but composes it with systemd state for the canonical 10-service set (plus optional nextcloud / jellyfin), so the dashboard finally reflects reality on the box.

Implementation pieces worth calling out:

- **`dashboard_service_targets`** hardcodes the 10 canonical services but reads `core` and `llm` unit names from `config.services.*.systemd_unit`. That means the LLM row automatically switches between `genie-llm.service` (llama.cpp fallback) and `genie-ai-runtime.service` (current default) as the operator flips backends — no extra config plumbing needed.
- **`unique_units`** dedupes before subprocessing — important because two `service` keys can map to one unit and we don't want duplicate `systemctl show` invocations on every dashboard poll.
- **`query_systemd_unit`** uses `systemctl show … --property=LoadState --property=ActiveState --property=SubState --property=UnitFileState --property=Result` with a 2-second `tokio::time::timeout`. If dbus is hung or systemctl is missing entirely (macOS dev hosts), it returns a populated `SystemdRow { error: Some(...), ... }` instead of blocking the dashboard. Stderr is preserved as the error when the exit is non-zero and stderr is non-empty.
- **`read_latest_health_rows`** quietly fixed a real correctness bug in the old query. The original was `SELECT service, healthy, response_ms, error, MAX(ts_ms) FROM health_log GROUP BY service` — non-aggregated columns alongside `MAX(ts_ms)` is undefined per ANSI SQL and SQLite happens to pick *some* row per group, not necessarily the row that matches `MAX(ts_ms)`. The new join-against-latest pattern gets actually-the-latest row per service. Nice quiet improvement.
- **`merge_service_rows`** layers errors in a sensible order: health error → systemd error → derived state ("unit not installed" / sub_state / active_state). `healthy = !missing && systemd_active && health.healthy.unwrap_or(true)` means "no URL probe = trust systemd" — correct for services like `audio`, `mqtt`, `wakeword` that don't expose HTTP endpoints.

Dashboard JS pieces:

- **`serviceStatus(s)`** is the clean state-machine map: Healthy (green) → Missing (text2) → Failed (red) → non-active states (yellow with `sub_state`) → Unknown (red). The fall-through order matches what operators expect to see triaged first.
- **Unit name shown under each service** in a `.small` div is the right diagnostic. Operator clicking through can grep journalctl for the exact unit without guessing.
- **`escapeHtml()` on `s.service`, `s.unit`, and `status.label`** — defense-in-depth even though the values are hardcoded server-side. Good hygiene.
- **`latency = response_ms == null ? '--' : …`** handles the new "systemd-only" rows that don't have URL latency. Old code would have rendered `nullms`.

Tests cover the two key paths:

- `dashboard_targets_include_deployed_stack_services` asserts all 10 canonical services are in the target list. Anyone removing one will fail this loudly.
- `service_rows_merge_health_and_systemd_status` exercises the merge: `core` with health=healthy + systemd=active → healthy true with `source="health+systemd"`; `api` with no health row + systemd=inactive/dead → healthy false with `error="dead"` and `source="systemd"`.

Worth flagging, neither blocking:

- **Per-poll subprocess fanout.** 10+ `systemctl show` invocations per `/api/services` request. The dashboard polls every few seconds, so that's ~120 systemctl forks/minute on a busy operator session. Each one is cheap but they're not free — a future micro-optimization would batch into a single `systemctl show <unit1> <unit2> ...` call (systemctl accepts multiple units and emits sectioned output). Not blocking; the current shape is correct, just slightly chatty.
- **`api` row reflects itself.** The dashboard runs *inside* `genie-api`, so `genie-api` is querying its own systemd state for the `api` row. Tiny circular sniff, no real risk. If `systemctl show genie-api.service` ever times out, the row degrades to "systemctl timed out" but the rest of the response still renders.
- **`audio` is a oneshot** — `genie-audio.service` exits cleanly after configuring I2S routes. Resulting state is `active_state=active sub_state=exited`, which renders as Healthy. Correct, just worth knowing for the next "why does audio say Healthy when nothing is happening" question.

All 5 CI checks green on `70ac985` (fmt, clippy, test, aarch64 cross-compile, `--no-default-features`). Going in.

ai-hpc:
Merged at `d8b1af515fd71868001a094cd343ca92abbce6d2`.
Thanks to @carlos4s!

PR #87, opened 2026-05-18
core: preserve genie runtime prompt context

### LINKED ISSUE #85: Web chat tool dispatch broken — compaction strips tool manifest + memory context, LLM hallucinates name instead of memory_recall
## Summary

The web chat UI on port `3000` is returning bare LLM completions with no tool calls executed, even though `memory` clearly has the data the user is asking about. The model answers from prior chat turns or hallucinates a value, instead of dispatching `memory_recall` / household-preference lookups.

This is the predictable downstream failure mode of the aggressive prompt compaction landed in PR #74 (called out in that PR's review as a follow-up risk).

## Reproduction

Memory state (visible in the UI's memory / household-preference panel):

```text
preference  household.preference  household  normal  recalls 10  5/18/2026, 2:46:59 AM
  Jared is interested in cooking and gardening.
  User's name is Jared
  instead of using the name Jared use the alias Tom
```

Chat transcript through `:3000`:

```text
> hi
Hello! How can I assist you today?

> cool
Thanks for the cool input! How can I assist you today?

> did you remember my name?
Yes, I remember your name. How can I assist you today?

> what is my name?
Your name is GeniePod.
```

Expected:

- "what is my name?" → calls `memory_recall` (or whatever the household-preference lookup is named), reads back `Tom` (or `Jared` if the alias-mapping preference isn't being applied) — not `GeniePod`.
- "did you remember my name?" → likewise a tool-backed answer, not a content-free affirmation.

## Why this is almost certainly the PR #74 compaction path

PR #74's `compact_genie_runtime_messages` (in `crates/genie-core/src/llm/openai_compat.rs`) replaces ALL prior messages — including the system prompt that lists the tool manifest and any injected memory context — with a single user-role turn:

```text
{GENIE_RUNTIME_COMPACT_SYSTEM}\n\n{latest_user_content}
```

where `GENIE_RUNTIME_COMPACT_SYSTEM` is the hardcoded blurb `"You are GeniePod Home. Answer the user's latest request directly and concisely."`. That blurb says nothing about tools, has no memory context, and the model has no way to know `memory_recall` exists. So it just generates a chat completion.

Compaction triggers at `body.size() > GENIE_RUNTIME_MAX_BODY_BYTES (4 KB)`. A real GenieClaw system prompt (tool definitions + household preferences + recent turns) easily exceeds 4 KB, so compaction fires on every web-chat request once the user has any non-trivial memory state.

## Compounding factor — PR #76 widened the runtime context

PR #76 bumped `GENIEPOD_AI_RUNTIME_CONTEXT` from `2048` to `8192` with `--int8-kv`. The Jetson runtime can now comfortably hold ~16–24 KB of typical English chat text — so the 4 KB body-compaction trigger is now far more aggressive than necessary. The client is compacting prompts the runtime could trivially handle.

## Acceptance criteria

- Asking "what is my name?" through the `:3000` chat UI returns the value stored in memory (or in the household-preference store), not a hallucinated answer.
- Tool calls show up in `journalctl -u genie-core` for chat requests that should trigger them (`memory_recall`, household-preference lookups, etc.).
- Web-chat requests preserve the system prompt's tool manifest and at least the immediately-relevant memory context, so the LLM knows what tools exist and what facts it has.

### Option A — Preserve the system prompt during compaction

Keep the original system message (tool manifest + memory injection) and only compact the *history* (older user/assistant turns). Latest user turn plus the system prompt; drop only the oldest turns until under budget. This matches what PR #74's body description claimed but isn't what the code actually does.

### Option B — Raise `GENIE_RUNTIME_MAX_BODY_BYTES` to track context capacity

`GENIE_RUNTIME_MAX_BODY_BYTES` is currently a hardcoded 4 KB constant. With PR #76's 8192-token context, the runtime can comfortably handle ~16–24 KB. Either bump the constant or derive it from `GENIEPOD_AI_RUNTIME_CONTEXT`. Trade-off: simpler change, but compaction still wipes context the *next* time we widen prompts faster than we widen the runtime.

### Option C — Pre-resolve memory tool calls before sending to the LLM

Genie-core's tool-dispatch layer already has memory recall plumbed. Run an explicit `memory_recall` *before* the LLM call when the prompt mentions identity / preferences keywords, inject the recalled values into a small prompt, then send that compact body. Sidesteps the compaction problem entirely for the most-common identity-query path.

I'd recommend **A + B together**: A is the structural fix (preserve system context); B unlocks today's bigger runtime context. C is a useful future direction but heavier.

## Cross-references

- #73 — original issue that PR #74 fixed; web chat going through at all
- #74 — landed the compaction that's now too aggressive
- #75 / #76 — widened runtime context to 8192 tokens
- Related noted in #74 squash body: "current compaction replaces ALL prior turns with the latest user message + system blurb. Web chat returns a working response but doesn't remember previous turns. Should evolve toward 'system blurb + recent N user/assistant pairs' once the runtime's context capacity is more reliable." — that follow-up is exactly this issue.

carlos4s:
## Summary
- raise the Genie AI runtime request body budget from 4KB to 24KB for the 8192-token runtime context
- preserve real system messages during runtime compaction so tool manifests and injected memory context survive
- compact only older non-system history while keeping the latest user turn
- keep core health/dashboard endpoints responsive while a chat turn is waiting on the local LLM

Fixes #85

## Testing
- cargo fmt --check
- cargo test -p genie-core llm::openai_compat::tests
- cargo test -p genie-core server::tests
- cargo clippy -p genie-core --all-targets --locked -- -D warnings

## Real Behavior Proof
- [x] Jetson verified `/api/chat` returned `Your name is Jared` with `tool=memory_recall` after deploying this branch.
- [x] Local tests verify runtime prompt compaction keeps memory/tool context and server checks still pass.

## Jetson
- Pending retest: dashboard/core health should stay responsive while a long chat turn is running.

### DIFF
--- a/crates/genie-core/src/llm/openai_compat.rs
+++ b/crates/genie-core/src/llm/openai_compat.rs
@@ -122,7 +122,7 @@ impl RequestProfile {
             });
         }
 
-        let compacted_messages = self.compact_messages(messages);
+        let compacted_messages = self.compact_messages(messages, GENIE_RUNTIME_MAX_BODY_BYTES);
         let compacted_body =
             self.serialize_body(&compacted_messages, max_tokens, stream, response_format)?;
         Ok(PreparedChatBody {
@@ -164,15 +164,16 @@ impl RequestProfile {
         }
     }
 
-    fn compact_messages(&self, messages: &[Message]) -> Vec<Message> {
+    fn compact_messages(&self, messages: &[Message], max_body_bytes: usize) -> Vec<Message> {
         match self {
             Self::Generic => messages.to_vec(),
-            Self::GenieAiRuntime => compact_genie_runtime_messages(messages),
+            Self::GenieAiRuntime => compact_genie_runtime_messages(messages, max_body_bytes),
         }
     }
 }
 
-const GENIE_RUNTIME_MAX_BODY_BYTES: usize = 4 * 1024;
+const GENIE_RUNTIME_MAX_BODY_BYTES: usize = 24 * 1024;
+const GENIE_RUNTIME_BODY_OVERHEAD_BYTES: usize = 768;
 const GENIE_RUNTIME_COMPACT_SYSTEM: &str =
     "You are GeniePod Home. Answer the user's latest request directly and concisely.";
 
@@ -507,20 +508,65 @@ fn parse_status_line(line: &str) -> u16 {
         .unwrap_or(0)
 }
 
-fn compact_genie_runtime_messages(messages: &[Message]) -> Vec<Message> {
-    let Some(message) = messages
+fn compact_genie_runtime_messages(messages: &[Message], max_body_bytes: usize) -> Vec<Message> {
+    let system_messages = messages
         .iter()
-        .rev()
-        .find(|m| m.role == "user")
-        .or_else(|| messages.iter().rev().find(|m| m.role != "system"))
+        .filter(|m| m.role == "system")
+        .cloned()
+        .collect::<Vec<_>>();
+    let has_system_context = !system_messages.is_empty();
+
+    let Some(latest_idx) = messages
+        .iter()
+        .rposition(|m| m.role == "user")
+        .or_else(|| messages.iter().rposition(|m| m.role != "system"))
     else {
-        return Vec::new();
+        return system_messages;
     };
 
-    vec![Message {
+    let latest_source = &messages[latest_idx];
+    let latest = Message {
         role: "user".into(),
-        content: format!("{}\n\n{}", GENIE_RUNTIME_COMPACT_SYSTEM, message.content),
-    }]
+        content: if has_system_context {
+            latest_source.content.clone()
+        } else {
+            format!(
+                "{}\n\n{}",
+                GENIE_RUNTIME_COMPACT_SYSTEM, latest_source.content
+            )
+        },
+    };
+
+    let mut compacted = system_messages;
+    let body_budget = max_body_bytes.saturating_sub(GENIE_RUNTIME_BODY_OVERHEAD_BYTES);
+    let mut estimated_bytes = estimate_messages_bytes(&compacted) + estimate_message_bytes(&latest);
+    let mut retained_history = Vec::new();
+
+    for message in messages[..latest_idx]
+        .iter()
+        .rev()
+        .filter(|m| m.role != "system")
+    {
+        let message_bytes = estimate_message_bytes(message);
+        if estimated_bytes + message_bytes > body_budget {
+            break;
+        }
+        retained_history.push(message.clone());
+        estimated_bytes += message_bytes;
+    }
+
+    retained_history.reverse();
+    compacted.extend(retained_history);
+    compacted.push(latest);
+    compacted
+}
+
+fn estimate_messages_bytes(messages: &[Message]) -> usize {
+    messages.iter().map(estimate_message_bytes).sum()
+}
+
+fn estimate_message_bytes(message: &Message) -> usize {
+    message.role.len() + message.content.len() + 32
 }
 
 fn should_retry_without_system_role(messages: &[Message], err: &str) -> bool {
@@ -664,11 +710,11 @@ mod tests {
         let messages = vec![
             Message {
                 role: "system".into(),
-                content: "large tool manifest ".repeat(1_000),
+                content: "tool manifest memory_recall household context ".repeat(128),
             },
             Message {
                 role: "assistant".into(),
-                content: "older assistant turn ".repeat(200),
+                content: "older assistant turn ".repeat(2_000),
             },
             Message {
                 role: "user".into(),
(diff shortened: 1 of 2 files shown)


ai-hpc (approved):
Closes #85 with the structurally-right fix, plus a bonus that closes the chat-blocks-health symptom that wasn't a filed issue but was the real reason dashboard polls felt frozen during chat. Two commits:

**Commit 1 — runtime prompt compaction** (`a26a25a`)

Replaces the original PR #74 "compact to a single user turn + hardcoded blurb" with the design the PR description actually claimed:

- Preserves **all** system messages (tool manifest + household-preference injection) so the LLM knows what tools exist and what facts to recall. The reason #85 fired in the first place.
- Keeps the latest user turn untouched.
- Walks older non-system history *backwards from the latest turn* and retains as many older user/assistant pairs as fit in the byte budget. This is the "system blurb + recent N turns" shape the PR #74 body had described but didn't implement.
- Raises `GENIE_RUNTIME_MAX_BODY_BYTES` from `4 KB → 24 KB` to match the 8192-token INT8-KV runtime context from #76 (the rough-budget concern I flagged on #85). New `GENIE_RUNTIME_BODY_OVERHEAD_BYTES = 768` reservation for JSON envelope + serialization overhead, so the budget math doesn't quietly overshoot.
- Falls back gracefully: if there's no user turn at all, returns the system messages alone (test `genie_runtime_compaction_falls_back_to_latest_non_system_message` now asserts this preserves both system context and the latest non-system turn).

`estimate_message_bytes` uses `role.len() + content.len() + 32` — the `+32` is a JSON-envelope nudge that matches the actual serialized overhead per message reasonably well. Cheap to compute, and the budget is enforced inclusively (`> body_budget` breaks the loop) so we stay under the cap.

The Jetson acceptance the author noted in the PR body is the smoking gun: `/api/chat` returned `Your name is Jared` with `tool=memory_recall` — which is exactly what #85's reproduction was missing. Before this PR, the LLM had no way to know `memory_recall` existed because the tool manifest was getting compacted away.

**Commit 2 — concurrent request handling** (`a77d67b`)

Quieter but useful. Changes `ChatServer::serve` from sequential `accept().await; handle(...).await` to `tokio::task::LocalSet::spawn_local` so multiple requests can be in-flight on the single OS thread. Adds a `chat_turn_lock: Mutex<()>` that explicitly serializes only the chat-turn endpoints (`POST /api/chat`, `POST /api/chat/stream`, `POST /v1/chat/completions`). `/api/health`, `/api/services`, `/api/memories`, runtime contract, history list, etc. now run concurrently with an in-flight chat turn.

Net effect: the dashboard's 5s polling against `/api/health` and `/api/services` stays responsive even when an LLM call is taking 4-8 seconds. Previously the whole HTTP server was blocked behind the LLM call (the old comment claimed "LLM calls are seconds, HTTP is microseconds" — true but irrelevant when the queue is serialized).

The change is correctly scoped — chat turns still serialize via `chat_turn_lock`, so concurrent chat requests from two users still queue up (which is what we want; the LLM backend can only handle one prompt at a time anyway and parallel chat requests would just thrash the KV cache).

Knock-on refactor that makes the diff cleaner:

- Inline `(method, path)` tuple matching at every route → `classify_route()` returning a `RequestRoute<'a>` enum. Routes are now named (`RequestRoute::Chat`, `::Health`, etc.) and the dispatch table reads top-to-bottom without re-spelling method/path tuples.
- 4× repeated `if matches!(request_origin, Unknown) { Api } else { request_origin }` ternary → `normalized_origin(request_origin)` helper.
- `with_chat_turn_lock(lock, fut)` helper encapsulates the "lock then await" pattern for chat-turn endpoints.

`ChatServer::serve` signature change from `&self` to `self` (because the body is `Rc::new(self)` for `spawn_local`) — this is a breaking change for any external caller, but the only call sites are in `crates/genie-core/src/main.rs` (4 mutually-exclusive branches that each move once). Verified via the `cargo clippy + test (--no-default-features)` and `cargo test` checks both being green.

The tests are updated correctly:

- `genie_runtime_profile_compacts_large_core_prompt` now expects 2 messages (`system + user`) instead of 1; verifies "memory_recall" and "household context" survive into the wire payload.
- New `genie_runtime_profile_keeps_runtime_prompt_under_expanded_budget` covers the "real-sized 10 KB prompt should NOT compact at all under the new 24 KB budget" case — this is the test that would catch a regression where someone shrinks `GENIE_RUNTIME_MAX_BODY_BYTES` back toward 4 KB.
- `genie_runtime_compaction_falls_back_to_latest_non_system_message` updated for the 2-message output shape.

Worth flagging, neither blocking:

- **Token-budget is byte-based, not token-based.** `GENIE_RUNTIME_MAX_BODY_BYTES = 24 KB` is a rough proxy for "fits in 8192 token context after JSON overhead". For typical English chat text (~4 bytes/token) that's ~6k tokens before overhead, comfortable. For CJK / code-heavy content (denser bytes/token) it could over-fit. Right path eventually is to tokenize on the client side or probe runtime capacity. The current static value is defensible for an alpha.
- **`chat_turn_lock` is a process-wide singleton**, so even semantically-different chat endpoints (web `/api/chat` vs OpenAI-compatible `/v1/chat/completions`) serialize behind each other. That's the right call given the single shared LLM backend, but worth noting if anyone later wants per-conversation parallelism.
- **`LocalSet` + `Rc` is a single-thread design**. That's appropriate for a Jetson appliance with <10 concurrent users, but if someone ever wants to push this to many-thread / many-tenant deployments, the Mutex needs to become `tokio::sync::Mutex` (it already is) and the `Rc` needs to become `Arc`. Worth a one-line note in `serve`'s doc comment, but the doc comment already says "current-thread runtime" which is honest enough.

All 6 CI checks green on `a77d67b` (fmt, clippy, test, aarch64 cross-compile, `--no-default-features`, PR body checklist). Going in.

ai-hpc:
Merged at `b3c4be08c903c5d5403fd3012817e0793268675e`.
Thanks to @carlos4s!

PR #96, opened 2026-05-18
fix(ui): reflect degraded /api/health state in chat header

galuis116:
## Summary

Fixes #95

The embedded chat UI (`chat_ui.html`) now reads `status` and `llm` from `GET /api/health` instead of treating any HTTP 200 as fully connected. When genie-core is up but the LLM is offline, the header shows an amber degraded dot and **LLM offline** (or **Degraded** for other degraded states).

## Changes

- Add `applyHealthStatus()` to map `/api/health` JSON to header dot + label
- Add `.dot.degraded` (amber) between green (ok) and red (offline)
- Add regression test `chat_ui_reflects_api_health_status` in `tool_dispatch_test.rs`

## Real Behavior Proof

- [x] I have built and run the affected code locally (or noted why I could not).
- [x] I have verified the change end-to-end on Jetson hardware **OR** explained the equivalent verification path I used.

### What I ran

```bash
cd genie-claw
cargo build -p genie-core --release
cargo test -p genie-core --test tool_dispatch_test chat_ui_reflects_api_health_status

# LLM stopped; genie-core on port 3001
GENIEPOD_CONFIG=/tmp/geniepod-port3001.toml ./target/release/genie-core

curl -s http://127.0.0.1:3001/api/health | jq '{status, llm}'
```

Browser: `http://127.0.0.1:3001/` (hard refresh). DevTools console:

```javascript
const data = await (await fetch('/api/health')).json();
console.log({ status: data.status, llm: data.llm });
console.log(document.getElementById('status-text').textContent);
console.log(document.getElementById('status-dot').className);
```

### What I observed

- `/api/health` with LLM down: `{"status":"degraded","llm":"offline",...}`
- **Before:** green dot, header text `Connected`; chat POST returned connection error while header stayed connected
- **After:** amber dot (`dot degraded`), header text `LLM offline`
- Regression test passes locally
- No Jetson hardware in this environment; verification is local x86_64 against a running `genie-core` instance with LLM backend stopped

## Test plan

- [ ] `cargo test -p genie-core --test tool_dispatch_test chat_ui_reflects_api_health_status`
- [ ] Stop LLM backend, start `genie-core`, open chat UI — header shows **LLM offline** (amber), not **Connected**
- [ ] Start LLM, confirm header returns to **Connected** (green) when `/api/health` reports `status: ok` and `llm: connected`
- [ ] Stop `genie-core` — header shows **Offline** (red)

## Notes for reviewers

[...]

PR #98, opened 2026-05-18
fix(ui): reflect degraded /api/health state in chat header

### LINKED ISSUE #95: Chat UI header shows “Connected” when `/api/health` reports a degraded stack
## Summary

The embedded web chat UI (`crates/genie-core/src/chat_ui.html`) shows a green status dot and the label **Connected** whenever `/api/health` returns HTTP 200, without reading `status` or `llm` from the JSON body. When genie-core is up but the LLM backend is offline (or connectivity is degraded), the API correctly returns `"status": "degraded"` and `"llm": "offline"`, yet the UI still tells the user the system is fully connected.

## Duplicate check

Searched open and closed issues and pull requests on `GeniePod/genie-claw` (May 2026). **No duplicate found** for this UI status-indicator bug.

| Result | Notes |
| --- | --- |
| No open issue | Queries for `checkHealth`, `status-dot`, `degraded connected`, `misleading health`, `llm offline ui` returned no matches. |
| [#73](https://github.com/GeniePod/genie-claw/issues/73) (closed) | Different problem: core `/api/health` reported `llm: connected` while `/api/chat` failed against `genie-ai-runtime` (integration/adapter bug, fixed in PR #74). |
| [PR #81](https://github.com/GeniePod/genie-claw/pull/81) (merged) | Animated writing indicator only; did not change `checkHealth()`. |
| [PR #84](https://github.com/GeniePod/genie-claw/pull/84) (merged) | Port 3080 dashboard services table; not the port 3000 embedded chat UI header. |

This issue is about the **port 3000 chat UI ignoring health fields**, not about health lying to operators (as in #73).

## Current behavior

1. Start genie-core with the LLM backend stopped or unreachable (for example `genie-llm.service` down, or no runtime on the configured URL).

2. Confirm the API reports degradation:

   ```bash
   curl -s http://127.0.0.1:3000/api/health | jq '{status, llm, llm_backend}'
   ```

   Example when the LLM is down:

   ```json
   {
     "status": "degraded",
     "llm": "offline",
     "llm_backend": "genie-ai-runtime"
   }
   ```

3. Open the chat UI in a browser (same host/port as genie-core, typically `http://127.0.0.1:3000/`).

4. Observe the header status:

   - Green dot (`.dot`, not `.dot.offline`)
   - Text: **Connected**

5. Sending a message fails or returns an error in the chat pane, while the header still shows **Connected**.

`checkHealth()` runs on load and every 10 seconds; it never inspects `data.status` or `data.llm`:

```javascript
async function checkHealth() {
  try {
    const r = await fetch('/api/health');
    const data = await r.json();
    statusDot.className = 'dot';
    statusText.textContent = 'Connected';
  } catch {
    statusDot.className = 'dot offline';
    statusText.textContent = 'Offline';
  }
}
```

## Expected behavior

- **Core unreachable** (fetch error or non-2xx): red dot, label such as **Offline**.
- **`status: "ok"` and `llm: "connected"`**: green dot, **Connected** (or **Ready**).
- **`status: "degraded"` and/or `llm: "offline"`**: warning state (for example amber dot), label such as **Degraded** or **LLM offline**, so users know chat may not work before they send a message.
- Optionally surface connectivity degradation when `connectivity.state` is not `disabled` / `ready` (same semantics as `overall_health_status()` in `crates/genie-core/src/server.rs`).

Align with operator docs: [doc/deployment-and-ops.md](https://github.com/GeniePod/genie-claw/blob/main/doc/deployment-and-ops.md) lists “Chat UI loads but no answers” → check `/api/health`; the UI should reflect that health signal.

## Context

- Health endpoint: `GET /api/health` in `crates/genie-core/src/server.rs` sets `status` via `overall_health_status()` and `llm` to `"connected"` or `"offline"`.
- The chat UI is served from the same binary (`chat_ui.html` embedded in genie-core) and is the primary local operator surface on port 3000.
- `genie-ctl health` already prints separate lines for core reachability and LLM state; only the web UI header is misleading today.
- CSS today only defines `.dot` (green) and `.dot.offline` (red); a third “degraded/warning” style may be needed.

## Acceptance criteria

- With LLM up and `status: "ok"`, header shows a positive connected state (green dot).
- With genie-core up but `llm: "offline"` (and `status: "degraded"`), header does **not** show **Connected**; it shows a degraded/warning state and text that indicates the LLM is unavailable.
- With genie-core down or `/api/health` unreachable, header shows **Offline** (red dot).
- Status updates on the existing 10s poll without a full page reload.
- Add a regression test in `genie-core` (for example assert `chat_ui.html` contains logic that reads `status` or `llm` from the health response, similar to `chat_ui_uses_animated_writing_indicator` in `tool_dispatch_test.rs`).

## Suggested fix (for implementers)

- Parse `data.status` and `data.llm` in `checkHealth()`.
- Map to three UI states: offline / degraded / ok.
- Add `.dot.degraded` (or similar) using an existing or new CSS variable (for example amber) between green and red.

Image

galuis116:
## Summary

Fixes #95

The embedded chat UI (`chat_ui.html`) now reads `status` and `llm` from `GET /api/health` instead of treating any HTTP 200 as fully connected. When genie-core is up but the LLM is offline, the header shows an amber degraded dot and **LLM offline**.

## Changes

- Add `applyHealthStatus()` to map `/api/health` JSON to header dot + label
- Add `.dot.degraded` (amber) between green (ok) and red (offline)
- Add regression test `chat_ui_reflects_api_health_status` in `tool_dispatch_test.rs`

## Real Behavior Proof

### What I ran

```bash
cargo build -p genie-core --release
cargo test -p genie-core --test tool_dispatch_test chat_ui_reflects_api_health_status

GENIEPOD_CONFIG=/tmp/geniepod-port3001.toml ./target/release/genie-core
curl -s http://127.0.0.1:3001/api/health | jq '{status, llm}'
```

Browser: `http://127.0.0.1:3001/` — header shows **LLM offline** (amber) when LLM backend is stopped.

### What I observed

- `/api/health` with LLM down: `status: degraded`, `llm: offline`
- Header: amber dot + **LLM offline** (not **Connected**)
- Regression test passes locally
- No Jetson in this environment; local x86_64 verification only

## Test plan

- [ ] `cargo test -p genie-core --test tool_dispatch_test chat_ui_reflects_api_health_status`
- [ ] Stop LLM, open chat UI — **LLM offline** (amber)
- [ ] Start LLM — **Connected** (green) when health is ok
- [ ] Stop genie-core — **Offline** (red)

## Notes for reviewers

Commit author: `galuis116 <116897328+galuis116@users.noreply.github.com>`. No AI co-author trailers in git log.

### DIFF
--- a/crates/genie-core/src/chat_ui.html
+++ b/crates/genie-core/src/chat_ui.html
@@ -8,7 +8,7 @@
 :root {
   --bg: #0a0e14; --surface: #131920; --border: #1e2730;
   --text: #e0e6ed; --text2: #8899aa; --accent: #00d4ff;
-  --green: #00e676; --red: #ff5252; --user-bg: #1a2332; --bot-bg: #0f1820;
+  --green: #00e676; --amber: #ffab40; --red: #ff5252; --user-bg: #1a2332; --bot-bg: #0f1820;
   --tool-bg: #1a1a2e; --tool-border: #7c4dff;
 }
 * { margin: 0; padding: 0; box-sizing: border-box; }
@@ -36,6 +36,7 @@ header .dot {
   width: 8px; height: 8px; border-radius: 50%;
   background: var(--green); display: inline-block;
 }
+header .dot.degraded { background: var(--amber); }
 header .dot.offline { background: var(--red); }
 header button {
   background: none; border: 1px solid var(--border); color: var(--text2);
@@ -164,12 +165,22 @@ const statusText = document.getElementById('status-text');
 
 let sending = false;
 
+function applyHealthStatus(data) {
+  if (data.status === 'ok' && data.llm === 'connected') {
+    statusDot.className = 'dot';
+    statusText.textContent = 'Connected';
+    return;
+  }
+  statusDot.className = 'dot degraded';
+  statusText.textContent = data.llm === 'offline' ? 'LLM offline' : 'Degraded';
+}
+
 async function checkHealth() {
   try {
     const r = await fetch('/api/health');
+    if (!r.ok) throw new Error('health unavailable');
     const data = await r.json();
-    statusDot.className = 'dot';
-    statusText.textContent = 'Connected';
+    applyHealthStatus(data);
   } catch {
     statusDot.className = 'dot offline';
     statusText.textContent = 'Offline';
--- a/crates/genie-core/tests/tool_dispatch_test.rs
+++ b/crates/genie-core/tests/tool_dispatch_test.rs
@@ -403,6 +403,26 @@ fn chat_ui_uses_animated_writing_indicator() {
     );
 }
 
+/// Verify the chat UI header reflects `/api/health` status and LLM state.
+#[test]
+fn chat_ui_reflects_api_health_status() {
+    let path = workspace_root().join("crates/genie-core/src/chat_ui.html");
+    let contents = std::fs::read_to_string(&path).unwrap();
+
+    assert!(
+        contents.contains("function applyHealthStatus"),
+        "chat UI should map health JSON to header state"
+    );
+    assert!(
+        contents.contains("data.status === 'ok'") && contents.contains("data.llm === 'connected'"),
+        "chat UI should require both ok status and connected LLM before showing Connected"
+    );
+    assert!(
+        contents.contains("LLM offline") && contents.contains("dot degraded"),
+        "chat UI should show a degraded state when the LLM is offline"
+    );
+}
+
 /// Verify the model cache helper can inspect GGUF page-cache residency.
 #[test]
 fn model_cache_status_helper_reports_residency() {


ai-hpc (approved):
Thanks for the clean follow-up — commit author is now your GitHub-noreply (`116897328+galuis116@users.noreply.github.com`), no AI co-author trailer, and you flagged both in "Notes for reviewers" so a maintainer doesn't have to re-check. That's exactly the norm-pattern we wanted to land.

Code is the same clean fix from the closed PR #96, and it correctly closes #95:

- **The actual bug**: `checkHealth()` was setting `statusText = 'Connected'` whenever the HTTP request to `/api/health` succeeded — i.e., whenever genie-core itself was up enough to serve. But genie-core can be up while the LLM backend it depends on is down, and the dashboard had no way to surface that. The user would see "Connected", send a message, and get a connection error from the chat path.
- **The fix**: read `data.status` and `data.llm` from the JSON response. Show green + "Connected" only when both are healthy. Show amber + the specific failure label (`"LLM offline"` when `data.llm === 'offline'`, generic `"Degraded"` otherwise) when something's wrong but genie-core is still alive. Existing red + "Offline" branch is preserved for the network-level failure case.
- **Latent bug also fixed**: `if (!r.ok) throw new Error('health unavailable')` ensures non-2xx responses (5xx from genie-core under load, 503 during boot, etc.) correctly fall through to the catch branch instead of silently `r.json()`ing an empty/garbage body. Subtle but real.

`applyHealthStatus()` extracted as a function with a clean early return for the happy path is good factoring — the regression test can probe the function name + its dual-condition logic directly.

Three-state visual hierarchy (green → amber → red) maps cleanly to the data model (`status: "ok"` → `status: "degraded"` → fetch failed). Naming the CSS class `.dot.degraded` (not `.dot.warn`) matches the API field, so a reader who greps `degraded` finds both the server-side state and the client-side rendering.

Regression test `chat_ui_reflects_api_health_status` follows the same pattern as PR #81's `chat_ui_uses_animated_writing_indicator` — three pinned invariants on the static HTML (function name exists, dual condition is enforced, LLM-offline label + degraded class are present). Smoke-test scope is right for a static-asset change.

Worth flagging, not blocking: the static-asset test pins the exact string `data.status === 'ok'`. A future refactor that destructures (`const { status, llm } = data; if (status === 'ok' ...)`) would falsely fail the test. That's the inherent trade-off with HTML-string smoke tests. Not worth changing today; a future contributor who hits the false positive can broaden the pattern.

All 6 CI checks green on `53e15de` (fmt, clippy, test, aarch64 cross-compile, `--no-default-features`, PR body checklist). Going in.

ai-hpc:
Merged at `b65fafef9c957763ae67c033100277a0eae9991c`.
Thanks to @galuis116!

PR #106, opened 2026-05-18
runtime: stabilize Jetson startup context

### LINKED ISSUE #107: GENIEPOD_AI_RUNTIME_CONTEXT=8192 destabilizes the stack under steady-state load (chat UI lag + swap engagement)
## Summary

`GENIEPOD_AI_RUNTIME_CONTEXT=8192` (the alpha.9 default landed in PR #76, with `--int8-kv`) is destabilizing the box under normal operation. Symptoms observed on Jetson Orin Nano 8 GB with the full stack running:

- **Chat UI on `:3000` becomes unstable** — slow / hanging requests, occasional empty responses or 5xx, requests that used to be <1 s now occasionally take many seconds.
- **Other operations stutter** — `:3080` dashboard polls intermittently fail or lag, `genie-ctl status` slow, voice cycle latency banner shows higher variance than the alpha.7 baseline.
- **Swap file is in active use** — `free -h` shows non-zero `Swap: used`, which is the unambiguous signal that the working set has overflowed RAM. With swap engaged, every memory access has a chance of paging from disk, which compounds latency for *everything* (CUDA dispatch, HTTP handlers, audio capture).

## Why this is a real regression vs PR #76's verification

PR #76 raised `GENIEPOD_AI_RUNTIME_CONTEXT` from `2048` → `8192` based on issue #75's evidence that the runtime could fit 8192 tokens with INT8 KV when `genie-ai-runtime` loaded **first** (`Before=` whisper / homeassistant / genie-core). That measurement was a **cold-boot best case** — drop_caches, no resident services, then start genie-ai-runtime alone. Issue #75's table confirmed this explicitly:

| ctx | Loaded | KV cache | Runtime budget free | Note |
|---|---|---|---|---|
| 4096 | 4096 | 288 MB | 1903 MB | (genie-ai-runtime first) |
| 6144 | 6144 | 432 MB | 992 MB | (genie-ai-runtime first) |
| 8192 | 8192 | 576 MB | 1616 MB | (genie-ai-runtime first) |
| 4096 | **1735** | (clamped) | (~tight) | **after full stack was already running** |

The 8192-with-1616-MB-free measurement assumed nothing else was holding RAM at runtime startup. In steady-state operation — whisper resident + genie-core's HTTP server + piper + deep-filter + Home Assistant container + dashboards + Telegram adapter + tokio runtimes — the actual free RAM after genie-ai-runtime claims its full 8192-ctx KV pool is materially less than that 1616 MB. Once free RAM drops below the kernel's reclaim watermark, swap engages, and everything degrades.

## Reproduction (representative)

On a Jetson Orin Nano 8 GB running alpha.9 main (post-PR #76), after `make deploy` + `bash /opt/geniepod/setup-jetson.sh` + reboot to ensure boot order is what PR #76 designed:

```bash
ssh aihpc@192.168.55.1
free -h        # baseline; check Swap: used == 0 immediately after boot
sudo systemctl start geniepod.target
sleep 60       # let warmup land, dashboard register, etc.
free -h        # check Swap: used after the stack is fully resident
```

Reproduces non-zero `Swap: used` and audible / measurable lag in the chat UI for me.

## Acceptance criteria

- Steady-state operation with the full default stack (genie-ai-runtime + genie-whisper + genie-core + genie-api + genie-health + genie-governor + genie-mqtt + homeassistant) on Orin Nano 8 GB does **not** engage swap during normal use.
- Chat UI at `:3000` returns within the alpha.7-baseline latency band on every request, not just the first one.
- The first-reply latency banner from #19 shows the same ±100 ms median variance bound that PR #48 (#25) targets after CPU pinning lands.

### A. Drop the default context back to a safer steady-state value

Easiest fix: lower `GENIEPOD_AI_RUNTIME_CONTEXT` from `8192` to `4096` (or `5120` if benchmarking shows it fits). #75's data has `4096`-context-cold-boot fitting in `1903 MB` of runtime budget free, which has more headroom for the rest of the stack to share. Combined with PR #74's body-compaction threshold (24 KB for typical English chat), 4096 tokens is still plenty for the web-chat path that motivated raising the context in the first place.

Trade-off: voice-loop conversations that span many turns get less LLM context. For a home appliance with command-style voice this is fine; for a long discursive chat it's a regression on multi-turn memory.

### B. Make the context size auto-shrink under memory pressure

Probe `free` (or `/proc/meminfo`) at `genie-ai-runtime.service` start and pick the largest context that leaves a safety margin (e.g., 1.5 GB free post-load). `setup-jetson.sh` could compute a per-host `Environment=GENIEPOD_AI_RUNTIME_CONTEXT=<N>` drop-in based on `MemTotal` and the resident size of the rest of the enabled stack.

Trade-off: more complexity, more failure modes. Probably right long-term but heavier to land.

### C. Detect swap engagement and report it on the dashboard

Doesn't fix the cause but surfaces the symptom. `genie-health` could probe `/proc/swaps` and `/proc/meminfo` periodically and surface `swap_used_kb > 0` as a yellow dashboard warning. Operators would at least see why their box is slow.

Probably worth doing **alongside** A or B — even after dropping the context, swap-engagement is a useful health signal for future tuning regressions.

## Cross-references

- #75 — original benchmark that justified `8192` (cold-boot measurement only).
- PR #76 — landed `8192` + `--int8-kv` + `Before=` ordering as the new default. My review comment on that PR flagged "Smaller-Jetson safety: an Orin Nano 4 GB or earlier Jetson without `--int8-kv` support could fail to allocate 8192 ctx at boot." — this issue is the **same shape on the 8 GB Jetson** under steady-state load.
- PR #74 — `GENIE_RUNTIME_MAX_BODY_BYTES = 24 KB` body-compaction threshold on the client side. Already tracks the post-#76 context size; would need to be revisited if option A drops the context.
- #69 / PR #70 — page-cache preservation across runtime restart. Relevant because swap engagement may evict warm GGUF pages, which would re-amplify the latency cost of any LLM call.
- PR #48 — CPU pinning. Reduces *scheduler* jitter but doesn't help when the real bottleneck is paging from disk.

## Notes for the operator

If swap is already engaged on your box and you want to test the hypothesis before any code change:

```bash
ssh aihpc@192.168.55.1
sudo systemctl stop geniepod.target
sudo swapoff -a && sudo sync && sudo sh -c 'echo 3 > /proc/sys/vm/drop_caches' && sudo swapon -a
# Edit the drop-in:
sudo mkdir -p /etc/systemd/system/genie-ai-runtime.service.d
echo -e '[Service]\nEnvironment=GENIEPOD_AI_RUNTIME_CONTEXT=4096' \
  | sudo tee /etc/systemd/system/genie-ai-runtime.service.d/local-ctx.conf
sudo systemctl daemon-reload
sudo systemctl start geniepod.target
sleep 60
free -h
# If Swap: used stays at 0 and the chat UI is responsive, this issue is confirmed.
```

`/opt/geniepod/bin/genie-restart-all.sh` (PR #94) does the stop+drop_caches+swapoff+swapon+start sequence as a single command — useful for repeating the experiment cleanly.

bittoby:
Fixes #107

## Summary
- lower the default genie-ai-runtime context to 4096 with INT8 KV as the default service configuration
- make `start_all.sh` wait for the configured LLM health endpoint before starting memory-heavy services
- compact Genie AI Runtime requests to a small Jetson-friendly prompt while preserving tool rules and household context
- route identity/memory recall questions directly to `memory_recall`

## Testing
- [x] `cargo fmt --check`
- [x] `cargo test -p genie-core tools::quick::tests`
- [x] `cargo test -p genie-core llm::openai_compat::tests`
- [x] `cargo test -p genie-core --test tool_dispatch_test`
- [x] Jetson deploy + hard restart: runtime started with `-c 4096 --int8-kv`, `Context: 4096 tokens`, `KV cache: 288 MB`, `FREE: 1104 MB`
- [x] Jetson port 3000 chat: normal chat returned, streaming completed with token events and `done`
- [x] Jetson memory recall: `What is my name?` returned `Your name is Jared` with `tool: memory_recall` in 21 ms

## Real Behavior Proof

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,9 +175,10 @@ GenieClaw is intentionally narrower than a broad general-agent stack.
 That is a hardware decision as much as a product decision. In practical Jetson
 Orin Nano 8 GB testing, heavier agent shells can require very large context
 windows just to stay coherent, which drives up KV cache size, first-token
-latency, and overall memory pressure. Even `8192` context can already be tight
-on this class of device, and the result is often slower replies and worse
-appliance behavior.
+latency, and overall memory pressure. GenieClaw defaults to a 4096-token
+runtime context on this class of device because larger contexts can be too
+tight across full-stack restarts, causing slower replies or worse appliance
+behavior.
 
 For GenieClaw, that means:
 
--- a/crates/genie-core/src/llm/openai_compat.rs
+++ b/crates/genie-core/src/llm/openai_compat.rs
@@ -172,10 +172,12 @@ impl RequestProfile {
     }
 }
 
-const GENIE_RUNTIME_MAX_BODY_BYTES: usize = 24 * 1024;
-const GENIE_RUNTIME_BODY_OVERHEAD_BYTES: usize = 768;
+const GENIE_RUNTIME_MAX_BODY_BYTES: usize = 4 * 1024;
+const GENIE_RUNTIME_BODY_OVERHEAD_BYTES: usize = 512;
+const GENIE_RUNTIME_CONTEXT_MAX_BYTES: usize = 900;
 const GENIE_RUNTIME_COMPACT_SYSTEM: &str =
     "You are GeniePod Home. Answer the user's latest request directly and concisely.";
+const GENIE_RUNTIME_COMPACT_SYSTEM_PREFIX: &str = "You are GeniePod Home. Reply briefly for voice. Use a tool only when required. Tool calls must be ONLY JSON: {\"tool\":\"tool_name\",\"arguments\":{}}. No markdown.";
 
 impl OpenAiCompatClient {
     pub fn new(backend_name: &'static str, host: &str, port: u16) -> Self {
@@ -509,19 +511,29 @@ fn parse_status_line(line: &str) -> u16 {
 }
 
 fn compact_genie_runtime_messages(messages: &[Message], max_body_bytes: usize) -> Vec<Message> {
-    let system_messages = messages
+    let system_text = messages
         .iter()
         .filter(|m| m.role == "system")
-        .cloned()
-        .collect::<Vec<_>>();
-    let has_system_context = !system_messages.is_empty();
+        .map(|m| m.content.trim())
+        .filter(|m| !m.is_empty())
+        .collect::<Vec<_>>()
+        .join("\n\n");
+    let mut compacted = if system_text.is_empty() {
+        Vec::new()
+    } else {
+        vec![Message {
+            role: "system".into(),
+            content: compact_genie_runtime_system(&system_text),
+        }]
+    };
+    let has_system_context = !compacted.is_empty();
 
     let Some(latest_idx) = messages
         .iter()
         .rposition(|m| m.role == "user")
         .or_else(|| messages.iter().rposition(|m| m.role != "system"))
     else {
-        return system_messages;
+        return compacted;
     };
 
     let latest_source = &messages[latest_idx];
@@ -537,7 +549,6 @@ fn compact_genie_runtime_messages(messages: &[Message], max_body_bytes: usize) -
         },
     };
 
-    let mut compacted = system_messages;
     let body_budget = max_body_bytes.saturating_sub(GENIE_RUNTIME_BODY_OVERHEAD_BYTES);
     let mut estimated_bytes = estimate_messages_bytes(&compacted) + estimate_message_bytes(&latest);
     let mut retained_history = Vec::new();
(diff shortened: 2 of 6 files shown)


ai-hpc (approved):
Closes #107 with option A, plus three structurally-right improvements that make the whole stack more robust at the new tighter context. Going in.

**1. Context drop `8192 → 4096` with `--int8-kv`** (`deploy/systemd/genie-ai-runtime.service`). Honest comment update on the unit explains why we're stepping back ("Keep the default at 4k because 8k can still fail to start on memory-fragmented full-stack restarts") rather than just silently changing the number. README narrative updated to match — the prior "Even 8192 context can already be tight" wording explicitly becomes "GenieClaw defaults to a 4096-token runtime context on this class of device because larger contexts can be too tight across full-stack restarts". Doc and config now agree.

**2. `start_all.sh` waits for LLM `/health` before starting memory-heavy services** — new `wait_for_http_health` helper polls `curl -fsS --max-time 2 "$url"` up to 180 times (3 min budget) before letting whisper / core / Home Assistant start. Right shape: the existing `Before=` systemd ordering from PR #76 makes systemd start them in order, but doesn't wait for the LLM unit to actually be *ready to serve* — it just waits for the unit to be `active`, which fires the moment `ExecStart` returns (and the LLM is still loading its model into iGPU memory at that point). The health-gate closes that gap, so the rest of the stack starts only after the runtime is actually answering. `[services.llm].url` is read from config with a fallback to `http://127.0.0.1:8080/health` if absent.

**3. Compaction tuned for the 4k runtime** (`crates/genie-core/src/llm/openai_compat.rs`). Three meaningful changes:

- `GENIE_RUNTIME_MAX_BODY_BYTES` lowered `24KB → 4KB` (and overhead `768 → 512`). The 24KB threshold from PR #74 (later expanded by PR #87) was sized for the 8192-token runtime; at 4096 tokens it would overrun.
- Compaction shape changed from "pass system messages through verbatim + retain N older user/assistant pairs" to **structured rebuild**: `compact_genie_runtime_system` emits a minimal system prompt with a generated tool-list (`compact_genie_runtime_tool_lines`), a generated rules block (`compact_genie_runtime_rules`), and a household-context tail (`compact_household_context`, capped at 900 bytes via `truncate_utf8`). The tool list is filtered to only the tools actually referenced in the source system prompt, so a chat-only deploy doesn't carry `home_control` text it can't act on. The rules block similarly adapts to what's available (`Home control is unavailable; say Home Assistant is not connected if asked` when HA is off).
- The hardcoded `GENIE_RUNTIME_COMPACT_SYSTEM` blurb gets replaced by a richer prefix that explicitly tells the LLM the tool-call JSON contract (`{"tool":"tool_name","arguments":{}}, no markdown`). That eliminates a class of "model emits prose mentioning a tool name instead of a structured call" failures.

`truncate_utf8` is a careful helper — drops to the previous char boundary if `max_bytes` lands mid-codepoint, then `trim_end`s. CJK / accented-char household contexts won't get sliced into invalid UTF-8.

Test updates correctly reflect the new shape — `genie_runtime_profile_compacts_runtime_prompt_under_4k_budget` (renamed from `_under_expanded_budget`) asserts `prepared.compacted == true` instead of `false`, asserts the new "GeniePod Home" + "memory_recall" + "What is my name?" survive while `"tool manifest tool manifest"` (the noise the compaction now drops) doesn't. The previous test pinned the old "do not compact under 24KB" behavior; the new test pins the new "compact under 4KB" behavior. Right reversal.

**4. Identity-recall fast-path** in `crates/genie-core/src/tools/quick.rs`. New `memory_recall_query` recognizes `"what is my name"`, `"do you remember my name"`, `"who am i"`, `"what do you remember about X"`, `"search memory for X"`, and similar patterns, then dispatches to `memory_recall` *without* going through the LLM at all. Per the PR body, returns "Your name is Jared" in 21 ms on Jetson — bypasses an entire LLM round-trip for the most common identity questions. Two new unit tests pin both the name-form ("what is my name" → `memory_recall { query: "name" }`) and the search-form ("search memory for Jared" → `memory_recall { query: "jared" }`). The old `does_not_route_memory_search_to_web` test is correctly *deleted* since the new behavior actively does route it (just to `memory_recall`, not `web_search`).

Knock-on: this fix at the routing layer is also a robust answer to issue #85's "the LLM hallucinated 'Your name is GeniePod'" failure. Even on a runtime that doesn't see the tool manifest correctly, the identity question never reaches the LLM in the first place.

`tool_dispatch_test.rs` extended to pin the new `start_all.sh` health-wait shape (`read_llm_url`, `wait_for_http_health`, `Configured LLM health`) and the new context default (`GENIEPOD_AI_RUNTIME_CONTEXT=4096`). Both invariants are now CI-enforced.

End-to-end on Jetson per the PR body: runtime started with `Context: 4096 tokens`, `KV cache: 288 MB`, `FREE: 1104 MB` (vs the 1616 MB the 8k cold-boot measurement promised — but that benchmark was misleading for the steady-state case anyway, exactly the point of #107). Chat returns normally with streaming and `done`. Identity recall fires `memory_recall` in 21 ms.

All 8 CI checks green on `18bad34` (fmt, clippy, test, aarch64 cross-compile, `--no-default-features`, shellcheck, ruff, PR body checklist). Going in.

ai-hpc:
Merged at `73b838bdff0427039daa618983a1a2773302931d`.
Thanks to @bittoby!

Issue #24, opened 2026-05-12
beta: evidence trails (tamper-evident append-only audit log of every agent session, tool decision, actuation)


ai-hpc:
## Motivation

There's a partial audit trail today — `/opt/geniepod/data/runtime/contracts.jsonl` records prompt/tool/policy contract hashes at startup, conversations are persisted, and tool dispatch emits journal lines. But none of these are a coherent **evidence trail** for the operator: there's no single append-only log of "who asked what, which tool ran, which actuation fired, what the result was, when" — and no integrity guarantee that the log hasn't been tampered with after the fact.

For beta we want forensic-grade trails: tamper-evident, structured, queryable, exportable. This is the system of record when someone asks "why did the assistant unlock the door at 3 AM?"

### In scope
- **Append-only audit log.** New file `/opt/geniepod/data/audit/events.jsonl` (rotated daily, retention configurable). Every entry has: monotonic sequence number, ISO-8601 timestamp, event kind, request origin, conversation id, speaker identity (when known), payload, and a hash chain anchor.
- **Event kinds covered.**
  - `voice_cycle` — user speech transcript + assistant response (already partially logged).
  - `tool_call_decision` — gate decision (allow / deny / pending-confirmation / rate-limited) with reason.
  - `tool_call_executed` — tool name, args (with sensitive fields redacted), result, latency.
  - `actuation` — every home-control action that crossed `actuation_safety` (device id, state before/after, confidence).
  - `memory_write` — every fact stored to long-term memory, with the source utterance.
  - `skill_load` — every skill loaded, manifest hash, policy verdict.
  - `config_change` — when `/etc/geniepod/geniepod.toml` or its drop-ins change, log the diff hash.
- **Hash chain.** Each entry includes `prev_hash` = `sha256(prev_entry_bytes)`. Tampering with any historical entry breaks the chain. Daily roll-over signs the previous day's tail hash with a key stored in `/opt/geniepod/data/audit/keys/`.
- **Redaction policy.** Sensitive args (HA tokens, MQTT passwords, plaintext credentials passed to tools) are redacted before write. Redaction list is configurable.
- **Operator tooling.**
  - `genie-ctl audit tail [--follow]` — like `journalctl`, but specifically over the evidence log.
  - `genie-ctl audit verify` — walks the chain, reports first broken link or "ok all N entries".
  - `genie-ctl audit export --since <date> --format jsonl|csv` — ships the trail off-box.
[...]

PR #132, opened 2026-05-21
fix(audit): surface IO failures in AuditLogger and ToolAuditLogger

### LINKED ISSUE #131: [bug] audit: AuditLogger and ToolAuditLogger silently drop events on IO failure
### Summary

Both `AuditLogger::append` (crates/genie-core/src/tools/actuation.rs:305) and `ToolAuditLogger::append` (crates/genie-core/src/tools/dispatch.rs:106) return `()` and silently swallow four distinct IO failure modes each (mkdir, open, serialize, write). A disk-full, permission-denied, or invalid-path condition produces gaps in the audit log with no log line, no metric, and no return signal — defeating the audit log's forensic purpose.

### Steps to reproduce

1. Read `crates/genie-core/src/tools/actuation.rs:305-320` — body of `AuditLogger::append`. Four silent-discard sites:
   - `let _ = std::fs::create_dir_all(parent);` (L311)
   - `let Ok(mut file) = OpenOptions::new().create(true).append(true).open(path) else { return; };` (L313)
   - `let Ok(line) = serde_json::to_string(&event) else { return; };` (L316)
   - `let _ = writeln!(file, "{line}");` (L319)
2. Read `crates/genie-core/src/tools/dispatch.rs:106-121` — body of `ToolAuditLogger::append`. Same four sites at L112 / L114 / L117 / L120.
3. Force a failure at runtime:
   ```
   sudo install -d -m 0000 /var/lib/genie-core
   # exercise any tool call via chat or voice — both audit paths will silently no-op
   ```
4. Observe: the action completes normally, no `tracing::error!` or `warn!` is emitted, and both audit files are unchanged. `read_recent_executed_actions` returns an empty `Vec`.

### Expected behavior

Both `append` methods should surface IO failures so that:
- A `tracing::error!` is emitted with the path and the underlying `io::Error` kind, visible in `journalctl -u genie-core`.
- A counter (e.g. `audit_writes_failed_total{logger="actuation|tool",kind="open|write|..."}`) increments, observable via `/api/health` or the dashboard.
- The method signature returns `Result<(), AuditError>` so a security-critical caller can opt into a "no audit, no action" posture if desired. Default caller behavior may stay log-and-continue; the choice should be the caller's, not the logger's.

### Actual behavior

All eight failure modes (four per logger) produce a silent no-op. The audit files can be made permanently zero-byte by a disk-full condition or a permissions misconfiguration, and nothing in the runtime notices. Replay / undo logic that depends on `read_recent_executed_actions` silently degrades to "no recent actions".

### Hardware

### JetPack / L4T version

_No response_

59c7e49

### Relevant logs

```shell
N/A — failure is silent, which is the bug.
```

### Additional context

- This is the underlying substrate for #24 (tamper-evident evidence trails). A hash chain over events you didn't actually write provides no integrity guarantee.
- Proposed PR sketch (will be in the linked PR; summarised here for scope review):
  - Introduce `pub enum AuditError { CreateDir(io::Error), Open(io::Error), Serialize(serde_json::Error), Write(io::Error) }` with `Display` + `std::error::Error`.
  - Change both signatures to `-> Result<(), AuditError>`.
  - Extract a shared helper (e.g. `append_json_line(path, lock, payload)`) so both loggers share one implementation and one set of tests.
  - At call sites in `actuation.rs` and `dispatch.rs`, wrap with a helper that emits `tracing::error!(path, ?kind, "audit write failed")` and increments the counter. Preserves today's log-and-continue default behaviour.
  - Unit tests using `tempfile` + `set_permissions` to force `PermissionDenied`, asserting the error variant and counter increment for each logger.
- Not a security disclosure under SECURITY.md: no RCE, no data exposure, no privilege escalation. Reliability + observability gap.

Khaostica:
## Summary

Fixes #131. `AuditLogger::append` and `ToolAuditLogger::append` previously returned `()` and silently swallowed four IO failure modes each (mkdir, open, serialize, write). A disk-full or permission-denied condition produced gaps in the audit log with no log line, no metric, and no return signal — defeating the log's forensic purpose. Both methods now return `Result<(), AuditError>`, and a new `append_or_log` wrapper preserves today's log-and-continue caller behaviour while routing failures through `tracing::error!`.

## Changes

- Added `pub enum AuditError { CreateDir, Open, Serialize, Write }` with `Display` + `std::error::Error` impls in `crates/genie-core/src/tools/actuation.rs`. Variants are deliberately narrow (one per failure step) so structured-log filters and future metric labels can distinguish them.
- Extracted `pub(crate) fn append_json_line<T: Serialize>(path, payload) -> Result<(), AuditError>` so both loggers share a single, tested IO path.
- Changed `AuditLogger::append` and `ToolAuditLogger::append` to return `Result<(), AuditError>`. Disabled loggers continue to return `Ok(())` without doing IO.
- Added `append_or_log` on both loggers. Emits `tracing::error!` on failure with the path and underlying error, then continues. Callers use the bare `append` when they need to react to the failure (e.g. a future "no audit, no action" code path).
- Converted the six fire-and-forget call sites in `tools/dispatch.rs` to `append_or_log`, preserving the prior log-and-continue behaviour while making failures visible in `journalctl -u genie-core`.
- Added unit tests covering: disabled-logger Ok path, JSONL round-trip, and a cross-platform `AuditError::CreateDir | AuditError::Open` surfacing test (uses a regular file at the parent path to force a deterministic IO failure on both Unix and Windows). Updated `audit_logger_reads_recent_executed_actions` to consume the new `Result`.

No new dependencies. A metrics counter (`audit_writes_failed_total`) would be a natural follow-up once the workspace introduces a metrics infrastructure — deliberately out of scope here per CONTRIBUTING.md's stance on new runtime deps.

## Real Behavior Proof

### What I ran

No Jetson hardware available. Dev host is Windows 11 + WSL2 Ubuntu 26.04 LTS (Rust 1.95.0 / cargo 1.95.0, matching the repo's pinned toolchain). The workspace doesn't compile directly on the Windows host because `genie-core` uses unconditional Unix-only APIs (`libc::gmtime_r`, `tokio::net::UnixStream`), so I ran every gate inside WSL Linux against the same on-disk worktree via `/mnt/c`:

```
cargo fmt --all -- --check
cargo test --workspace --locked --all-targets
cargo test --workspace --locked --doc
```

I also re-read each of the six caller sites in `tools/dispatch.rs` to confirm none of them depend on `append` returning early on a soft-failure path — they're all fire-and-forget, so the `append_or_log` rename is behaviour-preserving for the happy path and behaviour-improving for the failure path.

### What I observed

`cargo fmt --all -- --check`: clean (no output).

`cargo clippy --workspace --all-targets --locked -- -D warnings`:

```
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 53.62s
```

`cargo clippy -p genie-core -p genie-ctl --no-default-features --all-targets --locked -- -D warnings`:

```
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 39.98s
```

`cargo test --workspace --locked --all-targets` — the new and updated tests that exercise the changed code paths:

```
test tools::actuation::tests::append_returns_ok_when_logger_is_disabled ... ok
test tools::actuation::tests::append_returns_error_when_parent_path_is_a_file ... ok
test tools::actuation::tests::append_writes_jsonl_line_with_event_fields ... ok
test tools::actuation::tests::audit_logger_reads_recent_executed_actions ... ok
test tools::dispatch::tests::tool_audit_logger_disabled_appends_ok ... ok
test tools::dispatch::tests::tool_audit_logger_surfaces_blocked_parent_error ... ok
```

Workspace-wide summary across the four test binaries that ran in this run: `6 passed; 33 passed; 423 passed; 2 passed`, plus `17 passed; 1 failed` in the `tool_dispatch_test` integration target.

The single failure is **environment-only and unrelated to this change** — `jetson_lifecycle_scripts_are_valid_shell`, which does `bash -n` against `deploy/scripts/*.sh`. The files on `origin/main` are LF in the git index (verified via `git show origin/main:deploy/scripts/start_all.sh | od -c`) but Git-for-Windows' `core.autocrlf` rewrites them to CRLF on checkout, and bash on Linux chokes on `\r`. The Linux CI runners check out with LF and this test passes on every PR — and `git diff --stat origin/main` confirms this branch touches zero shell scripts (only `actuation.rs` and `dispatch.rs`).

`cargo test --workspace --locked --doc`:

```
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s   (genie_common)
test result: ok. 1 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 0.00s   (genie_core)
test result: ok. 0 passed; 0 failed; 2 ignored; 0 measured; 0 filtered out; finished in 0.00s   (genie_skill_sdk)
```

## Test plan

For a reviewer with Jetson hardware who wants to confirm the runtime behaviour end-to-end: point `[actuation].audit_path` at a read-only directory (e.g. `sudo install -d -m 0555 /var/lib/geniepod/audit-readonly`, then set `audit_path = "/var/lib/geniepod/audit-readonly/actuation.jsonl"`) and issue any actuation via the dashboard or voice. Without this fix the audit file silently stays empty; with this fix `journalctl -u genie-core` shows `audit event dropped due to IO failure error="audit log: open file for append: Permission denied" path=Some("/var/lib/geniepod/audit-readonly/actuation.jsonl")` per attempt, and the actuation itself still completes (log-and-continue behaviour preserved).

## Notes for reviewers

- `pub enum AuditError` is the only new public API surface in `genie-core`.
- I left `append` itself `pub` (returning `Result`) rather than crate-private, because a future evidence-trail caller (#24) may want a "no audit, no action" posture, which requires the error to be visible. Easy to demote if you'd prefer.
- `append_json_line` is `pub(crate)` only — purely a dedup helper between the two loggers in the same crate.
- I did not change behaviour at the six existing call sites beyond "now logs when it fails". Open to feedback if any of them should instead refuse to proceed when the audit cannot be recorded.

### DIFF
--- a/crates/genie-core/src/tools/actuation.rs
+++ b/crates/genie-core/src/tools/actuation.rs
@@ -1,6 +1,8 @@
 use serde::{Deserialize, Serialize};
 use std::collections::{HashMap, HashSet, VecDeque};
+use std::fmt;
 use std::fs::{File, OpenOptions};
+use std::io;
 use std::io::{BufRead, BufReader, Write};
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, Mutex};
@@ -284,6 +286,57 @@ pub struct AuditEvent {
     pub undo_of: Option<u64>,
 }
 
+/// Errors that can occur while appending an event to an on-disk audit log.
+///
+/// Each variant identifies which underlying step failed so callers can
+/// distinguish e.g. a misconfigured path (`CreateDir` / `Open`) from a
+/// disk-pressure failure (`Write`). The inner errors are preserved so the
+/// `io::ErrorKind` and serde detail remain available for structured logging.
+#[derive(Debug)]
+pub enum AuditError {
+    CreateDir(io::Error),
+    Open(io::Error),
+    Serialize(serde_json::Error),
+    Write(io::Error),
+}
+
+impl fmt::Display for AuditError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::CreateDir(err) => write!(f, "audit log: create parent directory: {err}"),
+            Self::Open(err) => write!(f, "audit log: open file for append: {err}"),
+            Self::Serialize(err) => write!(f, "audit log: serialize event: {err}"),
+            Self::Write(err) => write!(f, "audit log: write line: {err}"),
+        }
+    }
+}
+
+impl std::error::Error for AuditError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        match self {
+            Self::CreateDir(err) | Self::Open(err) | Self::Write(err) => Some(err),
+            Self::Serialize(err) => Some(err),
+        }
+    }
+}
+
+/// Append a single JSON-encoded line to `path`, creating parent directories
+/// as needed. Returns the specific `AuditError` for the failed step so the
+/// caller can surface it to logs / metrics. Shared by `AuditLogger` and
+/// `ToolAuditLogger` so both have identical behavior under IO failure.
+pub(crate) fn append_json_line<T: Serialize>(path: &Path, payload: &T) -> Result<(), AuditError> {
+    if let Some(parent) = path.parent() {
+        std::fs::create_dir_all(parent).map_err(AuditError::CreateDir)?;
+    }
+    let mut file = OpenOptions::new()
+        .create(true)
+        .append(true)
+        .open(path)
+        .map_err(AuditError::Open)?;
+    let line = serde_json::to_string(payload).map_err(AuditError::Serialize)?;
+    writeln!(file, "{line}").map_err(AuditError::Write)
+}
+
 #[derive(Debug, Clone, Default)]
 pub struct AuditLogger {
     path: Option<PathBuf>,
@@ -302,21 +355,31 @@ impl AuditLogger {
         }
     }
 
-    pub fn append(&self, event: AuditEvent) {
+    /// Append an audit event. Returns the specific failure kind on IO or
+    /// serialization error so callers can log structured detail (or refuse
+    /// to proceed if their security posture is "no audit, no action").
+    ///
+    /// When the logger is `disabled()`, returns `Ok(())` without doing any IO.
+    pub fn append(&self, event: AuditEvent) -> Result<(), AuditError> {
         let Some(path) = &self.path else {
-            return;
+            return Ok(());
         };
         let _guard = self.lock.lock().expect("audit logger lock");
-        if let Some(parent) = path.parent() {
-            let _ = std::fs::create_dir_all(parent);
+        append_json_line(path, &event)
+    }
+
+    /// Convenience wrapper for callers that have no recovery path: appends
+    /// the event and, on failure, emits a `tracing::error!` with the path and
+    /// underlying error. The error is intentionally swallowed — use [`append`]
+    /// directly if you need to react to the failure.
+    pub fn append_or_log(&self, event: AuditEvent) {
+        if let Err(err) = self.append(event) {
+            tracing::error!(
+                error = %err,
+                path = ?self.path,
+                "audit event dropped due to IO failure"
+            );
         }
-        let Ok(mut file) = OpenOptions::new().create(true).append(true).open(path) else {
-            return;
-        };
-        let Ok(line) = serde_json::to_string(&event) else {
-            return;
-        };
-        let _ = writeln!(file, "{line}");
     }
 
     pub fn path(&self) -> Option<&Path> {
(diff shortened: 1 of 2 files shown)


ai-hpc:
Closing this because the PR author and the single commit author do not match.

PR author: @Khaostica
Commit author: @Johnb1109

Please resubmit from the commit author account or clarify the authorship before reopening.

Khaostica:
@ai-hpc
@Johnb1109 Is the GitHub account my partner uses on a different device, sorry about the confusion.
Do you need any other information or could you reopen the PR?

Khaostica:
@ai-hpc Would you consider reopening this PR? The authorship should be clarified.

Khaostica:
@ai-hpc
Can this PR get re-opened? The authorship was stated a week ago.

PR #144, opened 2026-05-22
fix(memory): correct FTS5 column indices for bm25_rank and evergreen (#143)

### LINKED ISSUE #143: [bug] memory: FTS5 column index off-by-one breaks BM25 ranking and evergreen protection
### Summary

In the FTS5 search path, `bm25_rank` and `evergreen` are read from the wrong column indices, causing all relevance ranking to collapse to 0/1 and silently bypassing decay protection for evergreen memories.

### Steps to reproduce

1. Store at least one evergreen memory and several regular memories via `Memory::store_evergreen` / `Memory::store`.
2. Call `Memory::search(query, limit)` with a query that matches multiple entries.
3. Observe that results are not ranked by relevance — all BM25 scores are 0.0 or 1.0.
4. Observe that the evergreen entry is subject to decay multiplication despite being marked evergreen.

The root cause is in `crates/genie-core/src/memory/mod.rs` lines 332–333.

The SELECT in the FTS search query returns 13 columns (indices 0–12):

```
0  m.id
1  m.kind
2  m.content
3  m.created_ms
4  m.accessed_ms
5  m.recall_count
6  m.max_score
7  m.promoted
8  m.scope
9  m.sensitivity
10 m.spoken_policy
11 m.evergreen
12 bm25(memories_fts) AS bm25_rank
```

`read_entry` consumes indices 0–10. The two remaining columns are then read:

```rust
// WRONG — reads m.evergreen (INTEGER 0/1), not the BM25 float
let bm25_rank: f64 = row.get::<_, f64>(11).unwrap_or(0.0);
// WRONG — reads m.spoken_policy (TEXT "allow"), not the evergreen flag
let evergreen: bool = row.get::<_, i64>(10).unwrap_or(0) != 0;
```

The fix is:

```rust
let bm25_rank: f64 = row.get::<_, f64>(12).unwrap_or(0.0);
let evergreen: bool = row.get::<_, i64>(11).unwrap_or(0) != 0;
```

### Expected behavior

- `bm25_rank` holds the real BM25 score (a negative float from FTS5), so `decay::bm25_rank_to_score` produces meaningful relevance weights and results are ranked correctly.
- `evergreen` is `true` for memories stored via `store_evergreen`, so the decay multiplier is `1.0` and those memories are never penalised by age.

### Actual behavior

- `bm25_rank` reads `m.evergreen` (0 or 1), so all BM25 scores collapse to `0.0` or `1.0`. Relevance ranking is completely broken for every FTS search.
- `evergreen` reads `m.spoken_policy` (TEXT `"allow"`), which `rusqlite` cannot parse as `i64` and falls back to `0`. Every memory — including permanently-stored evergreen facts — is treated as decayable. The evergreen protection mechanism is entirely bypassed for all search results.

### Hardware

### JetPack / L4T version

_No response_

`main` — first observed at commit range touching `crates/genie-core/src/memory/mod.rs`

### Relevant logs

```shell
No log output — the failures are silent. `unwrap_or` defaults suppress all errors; incorrect values flow through scoring without panicking.
```

### Additional context

- The `ORDER BY bm25_rank` clause in the FTS query also returns results in the wrong order because `bm25_rank` is `0` or `1` rather than a meaningful float.
- `prune_decayed` and `prune_stale` are unaffected — they query `evergreen` directly in SQL and are correct.
- The existing `evergreen_memories_dont_decay` unit test exercises `prune_decayed`, not the search path, so it does not catch this regression.

JPette1783:
## Summary

Fixes two off-by-one column index errors in the FTS5 memory search path that silently broke BM25 relevance ranking and bypassed evergreen decay protection for every search result. Closes #143.

## Changes

- `crates/genie-core/src/memory/mod.rs:332` — corrected `bm25_rank` read from index `11` → `12` (was reading `m.evergreen` INTEGER instead of the BM25 float)
- `crates/genie-core/src/memory/mod.rs:333` — corrected `evergreen` read from index `10` → `11` (was reading `m.spoken_policy` TEXT instead of the evergreen flag)

## Real Behavior Proof

No Jetson available — verified on x86_64 dev host. The bug is pure Rust logic with no hardware dependency: it is a wrong array index on a rusqlite `Row`, reproducible on any platform that can run `cargo test`.

### What I ran

```
# Environment: x86_64 Linux, rustc 1.78, SQLite 3.45
cargo test -p genie-core memory 2>&1
```

### What I observed

All existing memory tests pass on both the broken and fixed code because the only search-path test (`evergreen_memories_dont_decay`) exercises `prune_decayed`, which queries `evergreen` directly in SQL and is unaffected by these indices.

To confirm the indices are now correct, the column order was traced manually against the SELECT in `Memory::search`:

```
Index  Column
-----  ------
0      m.id
1      m.kind
2      m.content
3      m.created_ms
4      m.accessed_ms
5      m.recall_count
6      m.max_score
7      m.promoted
8      m.scope
9      m.sensitivity
10     m.spoken_policy        ← last field consumed by read_entry
11     m.evergreen            ← evergreen now reads here (was 10)
12     bm25(memories_fts)     ← bm25_rank now reads here (was 11)
```

Before the fix: `row.get::<_, f64>(11)` returned `0.0` or `1.0` (the evergreen INTEGER cast to f64). After the fix: it returns a real negative BM25 float (e.g. `-1.847`), which `bm25_rank_to_score` maps to a meaningful relevance weight.

Before the fix: `row.get::<_, i64>(10)` on `spoken_policy = "allow"` failed to parse, fell back to `0`, so `evergreen` was always `false`. After the fix: `row.get::<_, i64>(11)` on `evergreen = 1` correctly returns `true`.

## Test plan

1. Checkout `fix/memory-fts5-column-index-143`.
2. Run `cargo test -p genie-core memory` — all existing tests should pass.
3. To exercise the fixed path directly:
   ```rust
   let mem = Memory::open_in_memory().unwrap();
   mem.store_evergreen("fact", "permanent knowledge").unwrap();
   mem.store("fact", "regular memory one").unwrap();
   mem.store("fact", "regular memory two").unwrap();
   let results = mem.search("knowledge", 10).unwrap();
   // Verify results[0] is "permanent knowledge" and is not decay-penalised.
   ```
4. Add a `println!` on the raw `bm25_rank` value to confirm it is a real negative float, not `0.0`/`1.0`.

## Notes for reviewers

- `prune_decayed` and `prune_stale` are unaffected — they filter on `evergreen` in SQL directly and never touch these indices.
- The existing `evergreen_memories_dont_decay` unit test does not cover the FTS search path and therefore did not catch this regression. A follow-up issue to add a search-path evergreen test would be worthwhile.
- `ORDER BY bm25_rank` in the SELECT was also producing arbitrary ordering before this fix, since sorting integers `0`/`1` is not equivalent to sorting real BM25 scores.

### DIFF
--- a/crates/genie-core/src/memory/mod.rs
+++ b/crates/genie-core/src/memory/mod.rs
@@ -329,8 +329,8 @@ impl Memory {
         let raw_entries: Vec<(MemoryEntry, f64, bool)> = stmt
             .query_map(rusqlite::params![fts_query, limit * 3], |row| {
                 let entry = read_entry(row)?;
-                let bm25_rank: f64 = row.get::<_, f64>(11).unwrap_or(0.0);
-                let evergreen: bool = row.get::<_, i64>(10).unwrap_or(0) != 0;
+                let bm25_rank: f64 = row.get::<_, f64>(12).unwrap_or(0.0);
+                let evergreen: bool = row.get::<_, i64>(11).unwrap_or(0) != 0;
                 Ok((entry, bm25_rank, evergreen))
             })?
             .filter_map(|r| r.ok())


ai-hpc (approved):

ai-hpc:
Reviewed and merged: full CI passed, and the FTS column index fix restores BM25 ranking plus evergreen protection. Merged at 5f8d1f7af504ae7968ed4bf8cb51db1a67d821a5; thanks @web-dev0521.

PR #149, opened 2026-05-22
fix(governor): check systemctl exit status when swapping LLM model

### LINKED ISSUE #148: [bug] governor: swap_llm_model ignores systemctl exit status — memory-relief model swap silently no-ops (OOM risk)
### Summary

`ServiceCtl::swap_llm_model` runs `systemctl daemon-reload` and `systemctl restart <unit>` via `.output().await?` but never checks `output.status.success()` — unlike every other state-changing method in `service_ctl.rs`. If the restart is denied (polkit policy, masked unit, bad override), the governor's mode-driven LLM model swap silently no-ops while reporting success, so the heavier model keeps running and the device can OOM.

### Steps to reproduce

1. Run `genie-governor` on the device with mode transitions enabled (Day / NightA / NightB / Media).
2. Arrange for the LLM unit restart to fail — e.g. the governor process lacks polkit rights to `systemctl restart`, the unit is masked, or the systemd override the function just wrote is rejected on reload.
3. Trigger a mode transition that swaps the model, e.g. `Day/NightA -> NightB` (the memory-relief transition to a smaller model), `Media -> *`, or `NightB -> Day` (`governor.rs:204-223`).
4. Observe: `swap_llm_model` returns `Ok(())`. The governor logs only the pre-action `"swapping LLM model"` info line. The model never actually changes.

### Expected behavior

`swap_llm_model` should check `status.success()` on both the `daemon-reload` and `restart` commands, log the captured `stderr` on failure, and return `Err(...)` — matching `start`, `docker_start`, and `enable_zram` in the same file. A failed model swap during a memory-pressure transition must be observable (logged / surfaced), not reported as success.

### Actual behavior

```rust
// crates/genie-governor/src/service_ctl.rs:109-120
// Reload systemd and restart the LLM service.
Command::new("systemctl")
    .args(["daemon-reload"])
    .output()
    .await?;          // <- exit status discarded

Command::new("systemctl")
    .args(["restart", &unit])
    .output()
    .await?;          // <- exit status discarded

Ok(())                // <- always reports success
```

`.output().await?` only propagates an error if the *process fails to spawn*. A non-zero `systemctl` exit (permission denied, masked unit, reload error) is swallowed. Compare with the sibling methods, which all branch on `output.status.success()`:

- `start` (`service_ctl.rs:19-23`) — checks status, logs stderr, bails.
- `docker_start` (`service_ctl.rs:81-85`) — checks status, logs stderr, bails.
- `enable_zram` (`service_ctl.rs:136-139`) — checks status, logs stderr.

`swap_llm_model` is the only state-changing method that ignores it. The impact is worst on the `-> NightB` memory-relief swap: if it silently fails, the larger daytime model stays resident on the 8 GB Orin overnight, defeating the governor's purpose and risking OOM. (Callers in `governor.rs:207/215/222` additionally discard the result with `let _ = ...await`, so even after this fix the return value should be logged at the call site — but the function must first be *capable* of reporting failure.)

### Hardware

Jetson Orin Nano Super 8 GB

### JetPack / L4T version

_No response_

`main` — `crates/genie-governor/src/service_ctl.rs:91-121`

### Relevant logs

```shell
# All you ever see is the pre-action line; no error even when the restart was denied:
INFO genie_governor::service_ctl: swapping LLM model unit=genie-ai-runtime.service model=/opt/geniepod/models/<nightb-model>
# (no failure log, no error returned — model never actually swapped)
```

### Additional context

- Not a duplicate. Searched open + closed issues for `governor` / `swap_llm_model` / `daemon-reload` / `model swap` / `service_ctl` — the matches are config-resolution / default-model / deploy-pipeline topics (#40, #52, #54, #107, …); none describe the dropped exit status. Related but distinct from #107 (context size destabilizes the stack).
- Fix is small and local: capture `output`, branch on `status.success()`, log `stderr`, and `bail!` on failure for both commands — then log the result at the `governor.rs` call sites instead of `let _ =`.

enjoyandlove:
## Summary

Make the governor's LLM model swap report failures instead of swallowing them. `ServiceCtl::swap_llm_model` now checks the exit status of `systemctl daemon-reload` and `systemctl restart`, and the three call sites in `governor.rs` log the result instead of discarding it with `let _ =`. Fixes #148.

## Changes

- `service_ctl.rs`: `swap_llm_model` checks `status.success()` on both `systemctl daemon-reload` and `systemctl restart <unit>`, logs the captured `stderr` (`tracing::warn!`), and `bail!`s on a non-zero exit — matching the behavior of `start` / `docker_start` / `enable_zram`.
- `governor.rs`: the three swap call sites (`Day → NightA`, `NightA → NightB`, `NightB → Day`) now log a `tracing::error!` with the unit, model path, and error on failure instead of dropping the `Result` with `let _ =`.
- `CHANGELOG.md`: added an Unreleased entry for #148.

## Real Behavior Proof

### What I ran

```
cargo test -p genie-governor
cargo clippy -p genie-governor
```

## Test plan

- Build the governor crate and run its tests: `cargo test -p genie-governor`.
- Simulate a denied restart (mask the LLM unit or revoke the polkit policy), trigger a mode transition that swaps the model, and confirm:
  - `swap_llm_model` returns `Err`,
  - the captured `stderr` is logged at `warn`,
  - the call site logs `"LLM model swap failed"` at `error` with the unit and model path.

## Notes for reviewers

- Behavior on success is unchanged; this only adds error surfacing on the previously silent failure path.
- The previous `.output().await?` only surfaced spawn failures, so a denied restart (polkit policy, masked unit, rejected override) silently no-op'd while reporting success — leaving the heavier model resident during a memory-relief transition and risking OOM on the 8 GB Orin.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## Unreleased
 
+- **Governor LLM model swap reports failures** (#148): `ServiceCtl::swap_llm_model`
+  now checks `status.success()` on both `systemctl daemon-reload` and
+  `systemctl restart <unit>`, logs the captured `stderr`, and `bail!`s on a
+  non-zero exit — matching `start` / `docker_start` / `enable_zram`. Previously
+  `.output().await?` only surfaced spawn failures, so a denied restart (polkit
+  policy, masked unit, rejected override) silently no-op'd while reporting
+  success, leaving the heavier model resident during a memory-relief transition
+  and risking OOM on the 8 GB Orin. The three `governor.rs` call sites now log
+  the swap result (`tracing::error!`) instead of discarding it with `let _ =`.
 - **Real streaming TTS** (#26): the voice loop now detects sentence
   boundaries inside the LLM streaming callback and forwards completed
   sentences to a concurrent TTS task immediately, instead of waiting
--- a/crates/genie-governor/src/governor.rs
+++ b/crates/genie-governor/src/governor.rs
@@ -204,7 +204,9 @@ impl Governor {
                 if let Some(model) = target.llm_model() {
                     let path = format!("/opt/geniepod/models/{}", model);
                     let unit = self.llm_service_unit();
-                    let _ = ServiceCtl::swap_llm_model(&unit, &path).await;
+                    if let Err(e) = ServiceCtl::swap_llm_model(&unit, &path).await {
+                        tracing::error!(error = %e, unit = %unit, model = %path, "LLM model swap failed");
+                    }
                 }
                 let _ = tokio::fs::remove_file("/run/geniepod/media_mode").await;
             }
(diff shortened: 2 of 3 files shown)


ai-hpc (approved):
Reviewed and merged: CI passed, and the governor now reports systemctl failures instead of silently continuing. Merged at f56746d0edfc7486188bafdbb211cb2f28f86b49; thanks @enjoyandlove.

PR #150, opened 2026-05-22
fix(llm/openai_compat): slice error bodies on UTF-8 char boundary

### LINKED ISSUE #147: [bug] llm: non-ASCII backend error body panics & aborts the whole genie-core daemon (UTF-8 boundary slice)
### Summary

`truncate_body` in the OpenAI-compatible LLM client slices the response body at a raw byte offset (`&trimmed[..240]`). When byte 240 lands inside a multi-byte UTF-8 character — trivially common for any backend that returns CJK / emoji / accented text — Rust panics, and because the release profile sets `panic = "abort"`, the panic takes down the entire `genie-core` process (and with it `/api/health` and the voice loop), not just the one request.

### Steps to reproduce

1. Point `genie-core` at any OpenAI-compatible backend (e.g. the local `genie-ai-runtime`).
2. Make the backend return a response that is **not** clean JSON with an `error.message` field, where the body is **longer than 240 bytes** and contains a non-ASCII (multi-byte UTF-8) character near offset 240. Real-world triggers:
   - A non-200 status whose body is an HTML/plaintext proxy error page in a non-English locale (`openai_compat.rs:292`).
   - A 200 body that fails `ChatResponse` parsing and contains non-ASCII text (`openai_compat.rs:301`).
   - A streaming error body in the SSE path (`openai_compat.rs:369`).
3. Observe `genie-core` panic with `byte index 240 is not a char boundary` and the process **abort**.

Minimal reproduction of the faulty helper:

```rust
// crates/genie-core/src/llm/openai_compat.rs:768
fn truncate_body(body: &str) -> String {
    const MAX_LEN: usize = 240;
    let trimmed = body.trim();
    if trimmed.len() <= MAX_LEN {
        trimmed.to_string()
    } else {
        format!("{}...", &trimmed[..MAX_LEN]) // panics if byte 240 splits a char
    }
}

// e.g. a 300-byte body of repeated "é" (2 bytes each) panics at &trimmed[..240]
```

### Expected behavior

A malformed or non-ASCII error body from the backend should produce a *truncated error message* and a normal `Err(...)` that propagates up the call stack — exactly like the sibling helper `truncate_utf8` (`openai_compat.rs:682`) already does, by walking back to a char boundary with `is_char_boundary`. One bad response should affect at most one request; the daemon must stay up.

### Actual behavior

`&trimmed[..MAX_LEN]` panics when `MAX_LEN` (240) is not a char boundary. The workspace `[profile.release]` in the root `Cargo.toml` sets `panic = "abort"`, so the panic aborts the whole `genie-core` process rather than unwinding one task. This kills `/api/health` and, because the voice loop awaits the LLM call on the main task, the voice service dies with it. A single malformed/non-ASCII backend response is enough to take the daemon down (very plausible when `genie-ai-runtime` is under memory pressure and emits a partial/localized body).

Suggested fix — make `truncate_body` char-boundary-safe, reusing the same walk-back `truncate_utf8` already implements:

```rust
fn truncate_body(body: &str) -> String {
    const MAX_LEN: usize = 240;
    let trimmed = body.trim();
    if trimmed.len() <= MAX_LEN {
        trimmed.to_string()
    } else {
        let mut end = MAX_LEN;
        while end > 0 && !trimmed.is_char_boundary(end) {
            end -= 1;
        }
        format!("{}...", &trimmed[..end])
    }
}
```

### Hardware

### JetPack / L4T version

_No response_

`main` — `crates/genie-core/src/llm/openai_compat.rs:768-776`

### Relevant logs

```shell
thread 'main' panicked at 'byte index 240 is not a char boundary; it is inside 'é' (bytes 239..241) of `...`'
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
# panic = "abort" -> process exits; systemd shows genie-core dead, /api/health unreachable
```

### Additional context

- Distinct from #124 (stream cancellation) and #107 (context size). No existing open/closed issue covers this.
- The correct pattern already exists in the same file: `truncate_utf8` (`openai_compat.rs:682-692`) does the `is_char_boundary` walk-back. `truncate_body` is the only byte-sliced truncation and the clear oversight.
- All three call sites route through `backend_error_message` -> `truncate_body` on the failure paths listed above, so this is reachable on the common "backend returned something unexpected" path, not a rare edge.

enjoyandlove:
## Summary

`truncate_body` in `llm/openai_compat.rs` sliced backend error responses at a
fixed 240-byte offset. When that offset landed inside a multi-byte UTF-8
character, the slice panicked — and since release builds use `panic = "abort"`,
it took down the entire `genie-core` daemon instead of failing one request.
This walks the cutoff back to a char boundary so malformed responses surface as
an ordinary error string. Fixes #147.

## Changes

- `truncate_body` now truncates via the existing `truncate_utf8` helper instead
  of the raw `&trimmed[..MAX_LEN]` byte slice, so the cutoff always lands on a
  UTF-8 char boundary.
- Added regression tests covering a multi-byte char straddling the cutoff,
  short bodies passing through untouched, non-JSON Unicode error bodies hitting
  the truncation fallback, and nested/top-level JSON error message extraction.
- Added a CHANGELOG entry under Unreleased.

## Real Behavior Proof

### What I ran

```
cargo test -p genie-core --lib llm::openai_compat
```

### What I observed

The previous `&trimmed[..240]` path panicked when the 240-byte boundary fell
inside a multi-byte char (e.g. a localized HTML error page or malformed JSON),
aborting the daemon. With the fix, the same input is truncated to a valid UTF-8
string ending in `...`, verified by
`truncate_body_does_not_panic_on_multibyte_boundary` and
`backend_error_message_truncates_non_json_unicode_body`.

## Test plan

- `cargo test -p genie-core --lib llm::openai_compat`
- Point a backend at a URL returning a non-ASCII error body longer than 240
  bytes (e.g. a localized HTML 5xx page), trigger a request, and confirm the
  daemon logs a truncated error string instead of crashing.

## Notes for reviewers

- The fix reuses the existing `truncate_utf8` helper rather than introducing a
  new truncation path.
- Behavior for ASCII bodies and short bodies is unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 ## Unreleased
 
+- **Crash fix: non-ASCII backend error bodies** (#147): `truncate_body`
+  in `llm/openai_compat.rs` sliced the response body at a fixed 240-byte
+  offset (`&trimmed[..240]`). When that offset landed inside a multi-byte
+  UTF-8 character — common for localized HTML error pages or malformed
+  JSON — the slice panicked, and because release builds use
+  `panic = "abort"` it took down the whole `genie-core` daemon (health
+  checks and voice service) rather than failing the single request. The
+  truncation now walks back to a char boundary via the existing
+  `truncate_utf8` helper, so a malformed response surfaces as an ordinary
+  error string.
 - **Real streaming TTS** (#26): the voice loop now detects sentence
   boundaries inside the LLM streaming callback and forwards completed
   sentences to a concurrent TTS task immediately, instead of waiting
--- a/crates/genie-core/src/llm/openai_compat.rs
+++ b/crates/genie-core/src/llm/openai_compat.rs
@@ -771,7 +771,9 @@ fn truncate_body(body: &str) -> String {
     if trimmed.len() <= MAX_LEN {
         trimmed.to_string()
     } else {
-        format!("{}...", &trimmed[..MAX_LEN])
+        // Slice on a UTF-8 char boundary; a fixed byte offset can land mid-character
+        // and panic, which aborts the whole daemon under `panic = "abort"`.
+        format!("{}...", truncate_utf8(trimmed, MAX_LEN))
     }
 }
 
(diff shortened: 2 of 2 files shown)


ai-hpc (approved):

ai-hpc:
Reviewed and merged: CI passed, and the UTF-8 truncation fix prevents backend error bodies from aborting genie-core. Merged at a2c9fc5f16f41f394921c45d11fe99eb3fb873ea; thanks @enjoyandlove.

PR #151, opened 2026-05-22
feat(agent): add limited-context architecture harness

bittoby:
## Summary

Combines the five urgent direction changes into one PR:

- clarifies GenieClaw as the agent layer, not the voice/audio/runtime/device-control owner
- adds explicit AI, voice, and home runtime boundary contracts in `genie_core::runtime_boundary`
- adds a limited-context agent harness in `genie_core::agent_harness` for prompt, tool manifest, memory hydration, response reserve, and optional provider checks
- adds disabled-by-default optional API-key provider config plus `llm::OptionalProviderPlan` readiness checks
- adds portable `[agent].runtime_profile` config for Jetson, Raspberry Pi, generic SBC, laptop, and Mac profiles while keeping Jetson + 4096 tokens as the default production baseline

## Real Behavior Proof

### What I ran

```bash
git diff --check
which cargo || true
which rustc || true
which rustfmt || true
```

### What I observed

- `git diff --check` passed.
- This host does not have `cargo`, `rustc`, or `rustfmt`, so local Rust fmt/clippy/tests could not be run here.
- The change is structured to be validated by GitHub Actions: Rust fmt, clippy, workspace tests, and aarch64 cross-compile.

## Test plan

Expected CI coverage:

- `cargo fmt --all -- --check`
- `cargo clippy --workspace --all-targets --locked -- -D warnings`
- `cargo test --workspace --locked --all-targets`
- `cargo clippy --no-default-features`
- `cargo test --no-default-features`
- aarch64/Jetson cross-compile

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -19,6 +19,16 @@ It sits below:
 
 The purpose of this repo is to make the local home AI useful, private, safe, and understandable. It should not grow into the OS, the full home automation engine, the CUDA inference runtime, or the product app.
 
+The architectural invariant is now enforced in code as well as docs:
+
+- `genie_core::runtime_boundary` names the AI, voice, and home runtime owners.
+- `genie_core::agent_harness` validates prompt, tool, memory, response reserve,
+  and optional-provider context against the Jetson 4096-token baseline.
+- `[agent]` selects the deployment profile while keeping `jetson` as the
+  flagship default.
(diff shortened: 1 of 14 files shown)


ai-hpc (approved):
Reviewed: this matches the agent-layer redesign, keeps optional providers opt-in and limited-context gated, and CI is green across fmt, clippy, tests, no-default-features, and aarch64 cross-compile. No blocking issues found.

PR #163, opened 2026-05-23
test(home): regression for confirmed sensitive action at policy gate (closes #162)

kiannidev:
## Summary

- Add `control_confirmation_origin_executes_sensitive_action_when_confirmed` — covers the real `confirm_pending_home_action` re-dispatch path (`RequestOrigin::Confirmation`, `confirmed: true`).
- Add `control_confirmation_origin_still_requires_confirm_when_unconfirmed` — policy gate still returns `ConfirmationRequired` when `confirmed: false`.
- Complements existing `control_executes_sensitive_action_once_confirmed` (Telegram + confirmed), merged with #141 / #138 fix.

Closes #162

## Real Behavior Proof

- [x] I have built and run the affected code locally (or noted why I could not).
- [ ] I have verified the change end-to-end on Jetson hardware **OR** explained the equivalent verification path I used.

### What I ran

```bash
cargo test -p genie-core control_confirmation_origin
cargo fmt --all -- --check
```

### What I observed

Both new tests passed. No production code changes — tests only.

## Test plan

- [x] `cargo test -p genie-core control_confirmation_origin`
- [x] `cargo test -p genie-core control_executes_sensitive_action_once_confirmed`
- [x] `cargo fmt --all -- --check`

PR #141, opened 2026-05-22
fix(home/policy): execute confirmable action once confirmed

greatjourney589:
## Summary

Fixes #138. A confirmable home action (e.g. unlocking a door) was looping forever:
even after the caller supplied a confirmation, the policy gate re-issued a fresh
`ConfirmationRequired` on every retry and never executed the action. This makes the
gate honor `confirmed` so a satisfied confirmation falls through to execution, while
genuine hard denies still bail.

## Changes

- `control()` in `crates/genie-core/src/tools/home.rs` now branches on `confirmed`
  when a policy is not allowed but `requires_confirmation`: only re-prompt when the
  action is **not** yet confirmed; otherwise fall through to the runtime safety gate.
- Hard denies (not confirmable) still `bail!` with the policy reason — behavior
  unchanged for those.
- Added regression test `control_executes_sensitive_action_once_confirmed` covering a
  sensitive `lock`/`unlock` action that must execute (not re-prompt) when `confirmed = true`.

## Real Behavior Proof

- [x] I have built and run the affected code locally (or noted why I could not).
- [ ] I have verified the change end-to-end on Jetson hardware **OR** explained the equivalent verification path I used.

### What I ran

No Jetson available — verified via the crate's unit/integration tests on the dev host:

```
cargo test -p genie-core tools::home
```

### What I observed

The new `control_executes_sensitive_action_once_confirmed` test passes: a confirmed
`unlock` on "Front door" returns `ControlOutcome::Executed` (output contains "Unlock")
instead of looping back into `ConfirmationRequired`. Existing home-policy tests still pass.

## Test plan

- `cargo test -p genie-core tools::home`
- Trigger a sensitive action (e.g. unlock a door) via Telegram, observe the confirmation
  prompt, then confirm — the action should now execute once rather than re-prompting.

## Notes for reviewers

- Scope is limited to the confirmable-vs-deny branch in `control()`; the runtime safety
  gate downstream is unchanged.

PR #167, opened 2026-05-23
fix(actuation): preserve original origin through confirmation re-entry

### LINKED ISSUE #166: [bug] [bug] actuation: confirm_pending_home_action overrides request_origin with `Confirmation` — bypasses per-origin rate limit, ACL, and audit trail
### Summary

`ToolDispatcher::confirm_pending_home_action()` in `crates/genie-core/src/tools/dispatch.rs:786-814` discards `PendingConfirmation.requested_by` and re-enters `execute_with_context` with `request_origin: RequestOrigin::Confirmation`. That single override breaks three downstream contracts in `exec_home_control_inner`: the per-origin ACL check (`actuation_origin_allowed`), the per-origin rate limiter (`ActuationRateLimiter::check_and_record`), and the audit log (`AuditEvent.origin`). The net effect is that **per-origin actuation limits configured in `[core.actuation_safety].max_actions_per_minute_by_origin` can be bypassed via the confirmation flow**, sensitive-action audit rows lose the originating channel, and operator overrides of `allowed_origins` can silently lock out confirmation entirely (or, conversely, can never restrict confirmation per-source).

### Steps to reproduce

1. Config with a tight per-origin rate limit and the default `allowed_origins`:

   ```toml
   [core.actuation_safety]
   enabled = true
   allowed_origins = ["voice", "dashboard", "api", "telegram", "repl", "confirmation"]
   max_actions_per_minute = 12
   max_actions_per_minute_by_origin = { telegram = 1, voice = 8 }
   ```

2. Trigger a sensitive home action (lock / alarm panel / camera / garage / door cover — any entity that fails `home::voice_safe_target`) from `RequestOrigin::Telegram`:

   ```bash
   # via the internal tool dispatcher, RequestOrigin::Telegram
   home_control { "entity": "front_door_lock", "action": "lock" }
   # → returns "Confirmation required ... Pending token: act-..."
   ```

   The Telegram bucket is charged once even though nothing actuated.

3. Confirm the token:

   ```bash
   curl -s -X POST "http://127.0.0.1:3000/api/actuation/confirm?token=act-..."
   ```

4. Inspect `safety/actuation-audit.jsonl`:

   ```bash
   tail -n 2 ./data/safety/actuation-audit.jsonl
   ```

   - Both rows (the `ConfirmationIssued` and the `Executed`) show origin fields with the wrong combination: the issuing row has the original Telegram origin, but the **executed row reports `origin: "confirmation"`** — so a later "who unlocked the door?" search returns "confirmation", not "telegram".

5. Repeat: trigger another sensitive action from Telegram and confirm. The Telegram bucket is at 2/1 (limit exceeded) on the *request* side, but **the confirmation re-execution charges the `Confirmation` bucket**, which falls back to the global `max_actions_per_minute = 12`. So Telegram-channel callers can execute up to 12 sensitive actions/min via the confirm endpoint, despite the configured `telegram = 1`.

### Expected behavior

- `confirm_pending_home_action()` re-enters the dispatcher with the original origin (`pending.requested_by`), not `RequestOrigin::Confirmation`. The `confirmed: true` flag (already plumbed through after PR #141) is what signals the policy gate that the action is pre-approved — origin should not be overridden.
- The ACL check at `exec_home_control_inner` uses the original origin, so operators with `allowed_origins = ["voice", "dashboard"]` (no `"confirmation"` listed) can still complete confirmations triggered from those channels.
- The rate limiter does NOT double-charge: the original request already pushed a timestamp into the original-origin bucket on step 2. When the same logical action lands via `confirm`, the limiter should short-circuit on `exec_ctx.confirmed` rather than push a second timestamp.
- `AuditEvent.origin` for the executed/confirmed row reflects the channel that requested the action (`Telegram`, `Voice`, `Dashboard`, …). `exec_ctx.confirmed` is already known to the audit path; if downstream consumers need to distinguish "confirmed vs direct", surface that separately rather than by overloading the origin enum.
- Per-origin caps configured in `max_actions_per_minute_by_origin` are respected end-to-end, including across confirmation re-entry.

### Actual behavior

- `confirm_pending_home_action()` (dispatch.rs:786-814) hardcodes `RequestOrigin::Confirmation` on the re-enter call.
- `exec_home_control_inner` at dispatch.rs:589 then calls `actuation_origin_allowed(&self.actuation_safety, RequestOrigin::Confirmation)` — the original channel is no longer in the check.
- dispatch.rs:609-611 calls `self.actuation_rate_limiter.check_and_record(&self.actuation_safety, RequestOrigin::Confirmation)`. `actuation_rate_limit` (dispatch.rs:1133-1140) looks for `"confirmation"` in `max_actions_per_minute_by_origin` and otherwise returns the global `max_actions_per_minute` — defeating the per-origin limit the operator configured for the channel that actually issued the request.
- Every `AuditEvent` emitted on the confirm path (dispatch.rs:617, 664, 687, 714) stores `origin: exec_ctx.request_origin` = `Confirmation`, so `read_recent_executed_actions` returns `origin: "confirmation"` for any confirmed action and the originating channel is permanently lost in the audit trail.

### Hardware

Jetson Orin Nano Super 8 GB

### JetPack / L4T version

_No response_

v1.0.0-alpha.9 

### Relevant logs

```shell

```

### Additional context

_No response_

galuis116:
## Summary

Preserve the originating channel (`Voice`, `Telegram`, `Dashboard`, `Api`, `Repl`) when a pending sensitive home action is executed via the confirmation flow. Previously, `ToolDispatcher::confirm_pending_home_action()` rebuilt the `ToolExecutionContext` with `request_origin: RequestOrigin::Confirmation`, which silently bypassed `[core.actuation_safety].max_actions_per_minute_by_origin`, hid the originating channel in `safety/actuation-audit.jsonl`, and made `[core.actuation_safety].allowed_origins` overrides all-or-nothing for the entire confirmation flow. The `confirmed: true` flag (added in PR #141) is now the only signal the policy gate needs to know the action is pre-approved.

Fixes #166 .

## Changes

- `crates/genie-core/src/tools/dispatch.rs::confirm_pending_home_action` (L786-823): pass `pending.requested_by` as the re-entry `request_origin` instead of the hardcoded `RequestOrigin::Confirmation`. Added a multi-line comment explaining the invariant so future readers don't re-introduce the override.
- `crates/genie-core/src/tools/dispatch.rs::exec_home_control_inner` (L609-615): when `exec_ctx.confirmed == true`, skip `ActuationRateLimiter::check_and_record`. The original-origin bucket already paid one slot on the request that returned `ConfirmationRequired`; the confirmation re-entry must not double-charge the same logical action.
- `crates/genie-core/src/tools/dispatch.rs` tests: added a `SensitiveHomeProvider` fixture (returns `voice_safe: false`, `domain: "lock"`) plus two helpers (`extract_confirmation_token`, `audit_events_matching`) so the confirmation path can be exercised end-to-end inside `#[tokio::test]`.
- Three new regression tests in the `tests` module:
  - `confirm_preserves_original_origin_in_audit_log` — reads back the actuation audit JSONL and asserts the `Executed` row's `origin == "telegram"` (was `"confirmation"`).
  - `confirm_does_not_bypass_per_origin_rate_limit` — with `max_actions_per_minute_by_origin = { telegram = 1 }`, the second Telegram-initiated sensitive request inside the window is rejected with a "rate limit" message; only the first action reaches the HA provider.
  - `confirm_does_not_double_charge_when_already_paid` — with `max_actions_per_minute_by_origin = { telegram = 2 }`, request → confirm → request succeeds, proving the confirm did not consume the second slot.

No production-config defaults change. The behaviour of direct (non-confirmation) actuations is unchanged.

## Real Behavior Proof

- [x] I have NOT verified on Jetson hardware. The change is in pure dispatcher logic (no audio, voice, ALSA, CUDA, or Home Assistant runtime path), exercised end-to-end by `#[tokio::test]` against a mock `HomeAutomationProvider`. The equivalent verification path is the new in-process tests below plus the existing `home_control_rate_limits_by_origin`, `home_control_respects_configured_allowed_origins`, and `action_history_hydrates_from_audit_log` tests, all running green.

### What I ran

```bash
# Build the workspace from a clean main + this branch:
cargo build
# → Finished `dev` profile [unoptimized + debuginfo] target(s) in 1m 26s
# (5 binaries produced under target/debug/)

# Targeted: the three new regression tests
cargo test -p genie-core --lib tools::dispatch::tests::confirm
# → running 3 tests
# → test tools::dispatch::tests::confirm_does_not_bypass_per_origin_rate_limit ... ok
# → test tools::dispatch::tests::confirm_does_not_double_charge_when_already_paid ... ok
# → test tools::dispatch::tests::confirm_preserves_original_origin_in_audit_log ... ok
# → test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 448 filtered out

# Targeted: the full dispatch test module (29 tests, includes the existing
# direct-path rate-limit, ACL, and history-hydration tests this PR must not regress)
cargo test -p genie-core --lib tools::dispatch::tests::
# → test result: ok. 29 passed; 0 failed; 0 ignored; 0 measured; 422 filtered out

# Full workspace test suite
cargo test
# → TOTAL passed: 608  failed: 0  ignored: 3
# (same as main baseline + the 3 new tests added by this PR)
```

### What I observed

1. **Audit row now carries the originating channel.** Inside `confirm_preserves_original_origin_in_audit_log`, a Telegram-initiated sensitive `home_control { entity: "front door", action: "lock" }` is issued, then confirmed via `confirm_pending_home_action(token)`. The actuation audit JSONL contains exactly one `"status": "executed"` row and its `origin` is `"telegram"` — the value the test asserts. Before this patch the same setup produced `origin: "confirmation"`, which is what the bug report attached to this PR documents.

2. **Per-origin rate limit is now honoured through the confirmation flow.** Inside `confirm_does_not_bypass_per_origin_rate_limit`, with `max_actions_per_minute_by_origin = { telegram: 1 }`:
   - First Telegram request → `ConfirmationRequired` (charges slot 1).
   - Confirm → executes (no recharge, per the `confirmed`-guard in `exec_home_control_inner`).
   - Second Telegram request → rejected by the rate limiter with `output.to_lowercase().contains("rate limit")`. Only one action reached the `SensitiveHomeProvider::execute`. Before this patch the second request succeeded because it would have been routed through the `Confirmation` bucket (default global 12/min).

3. **No double-charge on confirm.** Inside `confirm_does_not_double_charge_when_already_paid`, with `telegram = 2`: request → confirm → request succeeds, proving the confirm did not push a second timestamp into the Telegram bucket. Before this patch the second request would have been rejected (bucket would be at 2/2 after the confirm: 1 for the issue, 1 for the re-entry).

4. **No direct-path regression.** The pre-existing `home_control_rate_limits_by_origin` (two `RequestOrigin::Dashboard` calls with `max_actions_per_minute_by_origin = { dashboard: 1 }`, second must be rate-limited) still passes, as does `home_control_respects_configured_allowed_origins` and `home_control_records_action_history`. Full `cargo test` matches the main-baseline 608 / 0 / 3.

## Test plan

A reviewer can re-verify on any Rust 1.85+ host (no Jetson, no Home Assistant, no audio needed):

- Check out this branch and run `cargo test -p genie-core --lib tools::dispatch::tests::confirm` — three tests, ~1s, all green.
- Run the broader `cargo test -p genie-core --lib tools::dispatch::tests::` — 29 tests, all green.
- Optional end-to-end manual proof against a real `genie-core`:
  1. Add a non-voice-safe entity in your dev Home Assistant (any `lock.*`, `cover.*` garage, or `alarm_control_panel.*`).
  2. In `geniepod.toml`, set `[core.actuation_safety]` with `enabled = true`, `max_actions_per_minute_by_origin = { telegram = 1 }`, default `allowed_origins`.
  3. Start `genie-core` against your dev HA (`HA_TOKEN=… cargo run --bin genie-core`).
  4. Trigger the sensitive action via the Telegram bot (or whichever channel you wired); you'll get a `Pending token: act-…` back.
  5. `curl -X POST "http://127.0.0.1:3000/api/actuation/confirm?token=act-…"` — the action executes.
  6. `tail -n 2 data/safety/actuation-audit.jsonl` — the `"status":"executed"` row carries `"origin":"telegram"` (was `"confirmation"` on main).
  7. Trigger the same Telegram action again inside the 60s window. It's now rate-limited (was previously accepted because the confirmation re-entry recharged the `Confirmation` bucket, not the `Telegram` bucket).

## Notes for reviewers

- **`RequestOrigin::Confirmation` is now unwritten in production code** — the only writer was the hardcoded re-entry this PR removes. The variant is left in the enum for serde-backward-compat with any historical entries that may already be on disk in `safety/actuation-audit.jsonl` from operators running pre-fix builds. A follow-up PR to delete the variant + its `Display`/`as_policy_key` branches + the `"confirmation"` entry in default `allowed_origins` can land cleanly once we're comfortable that no one needs to read those historical rows back through the enum. Kept out of this PR to minimise rebase churn against #163 and to keep the diff focused on the behavioural fix.
- **Timing.** Best to land *after* PR #163 (`test(home): regression for confirmed sensitive action at policy gate`) merges. That PR scaffolds the gate-side test for the same flow; rebasing this PR on top of it is cheaper than the reverse and keeps the new dispatcher tests adjacent to #163's gate test in `git log`.
- **Audit-trail back-compat.** This PR does NOT rewrite historical audit rows. Operators upgrading from a previous build will continue to see `origin: "confirmation"` on rows their old build wrote; only NEW confirmed executions carry the original channel. If a "rewrite old rows" migration is wanted, it should be its own PR.
- **No config schema change, no default behavioural change for direct (non-confirmation) calls.** Operators who never used `[core.actuation_safety]` overrides see no difference. Operators who did configure `max_actions_per_minute_by_origin` now have those limits actually enforced through the confirmation flow — that *is* the bug fix, and is the only operator-visible behavioural change.

### DIFF
--- a/crates/genie-core/src/tools/dispatch.rs
+++ b/crates/genie-core/src/tools/dispatch.rs
@@ -606,9 +606,14 @@ impl ToolDispatcher {
             });
             anyhow::bail!("Home action blocked by channel policy: {}", reason);
         }
-        if let Err(err) = self
-            .actuation_rate_limiter
-            .check_and_record(&self.actuation_safety, exec_ctx.request_origin)
+        // Skip the rate-limit recharge for pre-confirmed actions. The
+        // origin's bucket already paid one slot on the original request that
+        // returned `ConfirmationRequired`; counting the confirmation re-entry
+        // as a second hit would double-charge the same logical action.
+        if !exec_ctx.confirmed
+            && let Err(err) = self
+                .actuation_rate_limiter
+                .check_and_record(&self.actuation_safety, exec_ctx.request_origin)
         {
             let reason = err.to_string();
             self.audit_logger.append(AuditEvent {
@@ -796,11 +801,21 @@ impl ToolDispatcher {
                 "value": pending.value,
             }),
         };
+        // Re-enter with the channel that ORIGINALLY requested the action, not a
+        // synthetic `Confirmation` origin. The `confirmed: true` flag is what
+        // tells the policy gate the action is pre-approved; overriding origin
+        // would (a) hide the originating channel in `AuditEvent.origin`,
+        // (b) bypass `max_actions_per_minute_by_origin` for the requesting
+        // channel by charging the `Confirmation` bucket instead, and
+        // (c) break ACL setups whose `allowed_origins` exclude
+        // `"confirmation"`. The original-bucket already paid one slot when
+        // the request returned `ConfirmationRequired`, so the limiter skips
+        // re-charging here (see `confirmed`-guard in `exec_home_control_inner`).
         let result = self
             .execute_with_context(
                 &call,
                 ToolExecutionContext {
-                    request_origin: RequestOrigin::Confirmation,
+                    request_origin: pending.requested_by,
                     confirmed: true,
                     ..ToolExecutionContext::default()
                 },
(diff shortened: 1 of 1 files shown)


ai-hpc:
Reviewed and merged: #166 is valid, and this preserves the original actuation origin through confirmation while avoiding double-charge regressions covered by dispatcher tests. Merged at a46222b59a741165ff2d049021db6f4ea609b384; thanks @galuis116.

PR #169, opened 2026-05-23
fix(conversation): clip first-message title at UTF-8 char boundary (closes #168)

### LINKED ISSUE #168: [bug] conversation: first-message title byte slice panics on emoji / non-ASCII content — aborts genie-core daemon (UTF-8 boundary slice)
### Summary

`ConversationStore::append()` in `crates/genie-core/src/conversation.rs:82-119` formats the first user message of every new conversation into a 60-byte title via the byte slice `&content[..57]`. When byte 57 of the message lands inside a multi-byte UTF-8 codepoint, Rust panics with `byte index 57 is not a char boundary`. With `panic = "abort"` set workspace-wide ([Cargo.toml:39](Cargo.toml#L39)), this **aborts the whole `genie-core` daemon** in release builds. An emoji-laden English first message, or any sufficiently long Cyrillic / Greek / Hebrew / Arabic / mixed-language first message, reliably trips the panic on the user-facing `POST /api/chat` and `POST /api/chat/stream` endpoints, as well as on the voice transcript path in `voice_loop.rs`. This is the **same bug class as #147 / PR #150** (UTF-8 byte slice in `llm/openai_compat.rs::truncate_body`); the fix from that PR added a `truncate_utf8` helper but only patched the LLM error-body site — the conversation-title site still uses the naive byte slice.

### Steps to reproduce

1. Build and start `genie-core` against any working LLM backend (or even against a non-working one — the panic happens before the LLM call is dispatched, in the message-persistence step):

   ```bash
   GENIEPOD_CONFIG=deploy/config/geniepod.dev.toml ./target/release/genie-core
   ```

2. From a fresh conversation (any of: post-startup auto-created conversation, dashboard "new conversation" button, `POST /api/chat` without a `conv_id`), send an emoji-containing message whose byte length exceeds 60:

   ```bash
     -d '{"message":"I love coding! 🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉"}'
   ```

3. Observe `genie-core`'s log:

   ```
   thread 'tokio-runtime-worker' panicked at crates/genie-core/src/conversation.rs:103:35:
   byte index 57 is not a char boundary; it is inside '🎉' (bytes 55..59) of `I love coding! 🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉`
   ```

4. In release (`panic = "abort"`), the process exits and systemd respawns it. Inspect:

   ```bash
   systemctl status genie-core
   # → Active: active (running) since … (just restarted)
   journalctl -u genie-core -n 20
   # → "byte index 57 is not a char boundary" panic immediately before the unit's restart
   ```

5. The user's HTTP request fails with a dropped connection / 500. The user's POSTed message IS durably persisted in `data/conversations.db` (the INSERT at conversation.rs:90 ran to completion before the title-format panic at line 103), but the conversation has no title and the chat reply never lands.

### Expected behavior

- `ConversationStore::append()` truncates the first user message to a title at a UTF-8 char boundary, never panicking on a multi-byte codepoint.
- The character budget for the title is honoured by characters, not bytes — a 60-byte budget collapses to ~15 emoji or ~30 Cyrillic letters, which is less useful for the dashboard conversation-list rail than a per-character budget of similar magnitude (~60 chars + `…`).
- All other behaviours of `append()` are unchanged: the INSERT into `messages` still happens, the `count == 1 && role == "user"` branch still sets the title, the `else` branch still bumps `updated_ms`, and the title string format remains `"{truncated}..."` for compatibility with the dashboard rendering.
- In release (`panic = "abort"`), no input on `/api/chat`, `/api/chat/stream`, or the voice loop can crash the daemon by virtue of being non-ASCII.

### Actual behavior

- `conversation.rs:103` `format!("{}...", &content[..57])` uses byte indexing on a `&str`. Byte 57 is a char boundary only when the prefix up to byte 57 happens to align with codepoint boundaries:
  - ASCII content (1 byte/char): byte 57 is always a boundary. Safe.
  - Chinese / CJK (3 bytes/char): byte 57 is a multiple of 3, a boundary. Safe by coincidence.
  - Cyrillic / Greek / Hebrew / Latin-Extended (2 bytes/char): byte 57 is odd, falls inside the 29th char. **PANIC**.
  - Emoji (4 bytes/char): byte 57 = 4×14 + 1, falls inside the 15th emoji. **PANIC**.
  - Mixed: depends on the exact prefix.
- The panic happens AFTER the message INSERT succeeds, so the broken-state row is persisted before the daemon dies. Restarting `genie-core` re-loads the row from disk (title retrieval doesn't reformat), so re-opening the dashboard shows the broken conversation, but the original chat request still failed.
- Reachable from `crates/genie-core/src/server.rs:504` (`/api/chat`), `crates/genie-core/src/server.rs:728` (`/api/chat/stream`), and `crates/genie-core/src/voice_loop.rs:381` / `:1032` (voice transcripts that exceed 60 bytes). The voice path uses `let _ = conversations.append(...)` which suppresses the `Result` but **does NOT catch the panic** — the panic still propagates and aborts the process.

### Hardware

Non-Jetson (x86_64 dev / cross-build host) for the panic itself, since the bug is in pure persistence logic with no hardware path. Reproduces identically on Jetson — the panic semantics under `panic = "abort"` are the same, and the practical user impact (process abort + systemd respawn) is the worst on Jetson where restarts are slow (model weights need to be re-loaded, voice cold-start adds extra seconds).

### JetPack / L4T version

v1.0.0-alpha.9 / `main` @ `crates/genie-core/src/conversation.rs:102-106` (the title-format branch of `ConversationStore::append`).

### Relevant logs

```shell
# release build under systemd, emoji-laden first message on a fresh conversation
$ journalctl -u genie-core --since '1 minute ago'
… genie-core[12345]: thread 'tokio-runtime-worker' panicked at crates/genie-core/src/conversation.rs:103:35:
… genie-core[12345]: byte index 57 is not a char boundary; it is inside '🎉' (bytes 55..59) of `I love coding! 🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉`
… systemd[1]: genie-core.service: Main process exited, code=killed, status=6/ABRT
… systemd[1]: genie-core.service: Failed with result 'signal'.
… systemd[1]: genie-core.service: Scheduled restart job, restart counter is at N.
… genie-core[12378]: INFO  GeniePod core starting
```

```shell
# standalone reproducer that mirrors the buggy line, no build of genie-core needed:
$ cat > /tmp/title_panic.rs <<'EOF'
fn main() {
    let content = "I love coding! 🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉";
    println!("bytes={} chars={}", content.len(), content.chars().count());
    let _title = format!("{}...", &content[..57]);
}
EOF
$ rustc /tmp/title_panic.rs -o /tmp/title_panic && /tmp/title_panic
bytes=67 chars=28
thread 'main' panicked at /tmp/title_panic.rs:4:31:
byte index 57 is not a char boundary; it is inside '🎉' (bytes 55..59) of `I love coding! 🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉`
```

### Additional context

**Symptom users see:**
1. *Daemon abort on emoji greeting.* In release, the genie-core process exits with `SIGABRT` the first time a user sends an emoji-heavy first message in a new conversation. The user's request hangs / 500s; their message is silently persisted with no title; systemd respawns the daemon. The user has no obvious way to know the daemon crashed, only that "the assistant doesn't reply."
2. *Reachable from voice too.* `voice_loop.rs:381` and `voice_loop.rs:1032` use `let _ = conversations.append(...)` to suppress IO errors, but `let _` only ignores a `Result` — it does NOT catch a panic. A 60-byte+ voice transcript with multi-byte STT output (Cyrillic, Hebrew, Arabic, mixed-language households) on the first user turn of a new conversation has the same failure mode.
3. *Cross-language regression.* This bites every operator whose household uses Cyrillic, Greek, Hebrew, Arabic, or emoji as a default. ASCII-only and CJK-only households happen to be safe (byte 57 aligns by coincidence). Anything else rolls dice.

- `crates/genie-core/src/conversation.rs` has no test that calls `append()` with a non-ASCII message exceeding 60 bytes. The only `append` callers in the test suite (`voice_loop_integration.rs`, `tool_dispatch_test.rs`, etc.) use short ASCII fixtures.
- PR #150's regression test (`truncate_body_does_not_panic_on_multibyte_boundary` in `openai_compat.rs:1227`) lives next to the fixed `truncate_body`, but its coverage doesn't extend to other naive byte slices elsewhere in the workspace. A grep confirms `conversation.rs:103` is the only remaining naive-byte-slice on a `&str` content value in production code (audio/WAV `[..44]` slices in `voice/` are on `&[u8]` so they're safe).

**Files to touch:**
- `crates/genie-core/src/conversation.rs`:
  - Replace `&content[..57]` (line 103) with a char-boundary-aware truncation. Two reasonable patterns:
    - (a) Walk back to the nearest char boundary with a small helper:

      ```rust
      fn truncate_at_byte(text: &str, mut max_bytes: usize) -> &str {
          if text.len() <= max_bytes { return text; }
          while max_bytes > 0 && !text.is_char_boundary(max_bytes) { max_bytes -= 1; }
          &text[..max_bytes]
      }
      ```

    - (b) Truncate by CHARACTER count instead of byte count — closer to the operator-visible budget (the dashboard rail renders chars, not bytes):

      ```rust
      fn truncate_chars(text: &str, max_chars: usize) -> Cow<'_, str> {
          if text.chars().count() <= max_chars { Cow::Borrowed(text) }
          else { Cow::Owned(text.chars().take(max_chars).collect()) }
      }
      ```

    Pick (b) if you're willing to slightly change the displayed title for non-ASCII content; pick (a) if you want byte-for-byte parity with the current ASCII behaviour. Either is fine.
  - Optional consolidation: lift `truncate_utf8` from `crates/genie-core/src/llm/openai_compat.rs:872-880` into a small `crate::util` module so both call sites share a single implementation. Worth it only if the maintainers want the dedup; otherwise keep the helper local to `conversation.rs` to minimise PR surface.
- `crates/genie-core/src/conversation.rs` tests:
  - `append_title_truncates_emoji_first_message_without_panic` — 16 × `🎂` (64 bytes), `append(conv, "user", …, None)` must not panic; resulting title contains `…`-suffix (or `"..."` if matching legacy format) and is itself valid UTF-8.
  - `append_title_handles_cyrillic_first_message_at_odd_byte_boundary` — 31 × Cyrillic char (62 bytes, byte 57 is inside char 29); same expectation.
  - `append_title_short_message_used_verbatim` — ≤60-byte ASCII message; title is the exact content (regression on the else branch).
  - `append_title_long_ascii_truncates_with_ellipsis` — direct regression on the ASCII path that already works today.

**Acceptance (PR):**
- `cargo test -p genie-core --lib conversation` — existing tests pass, four new tests pass.
- Building `genie-core` in `--release` (which compiles under `panic = "abort"` per Cargo.toml) and sending the curl payload from "Steps to reproduce" no longer aborts the daemon; the conversation gets a sensible title; `/api/chat` returns a normal reply.
- The voice-path repro (a transcribed 60-byte+ Cyrillic / emoji first message via `voice_loop`) no longer aborts the daemon.
- The dashboard conversation-list rail (`GET /api/conversations`) shows the new conversation's truncated title rendering correctly — no broken UTF-8 in the JSON response.

**Timing:** Standalone — no upstream PRs touch `conversation.rs` (last commits on this file: `5871183 Restore actuation history from audit log` and earlier). Land independently of #163. If the maintainers want PR #150's `truncate_utf8` helper lifted to a shared util crate as part of this work, that becomes the natural follow-up scope discussion.

**Real Behavior Proof:** `cargo test -p genie-core --lib conversation` showing the four new tests pass. Plus optional `cargo run --release -p genie-core` against the dev TOML, the curl payload from "Steps to reproduce" returning a 200 with a normal chat reply, and `tail data/conversations.db | sqlite3 -cmd '.headers on' 'SELECT id, title FROM conversations ORDER BY created_ms DESC LIMIT 1;'` showing the truncated title with a valid suffix.

**Related:** #147 / PR #150 fixed the same UTF-8 byte-slice bug class in `llm/openai_compat.rs::truncate_body` (LLM backend error-body slicing) and added a `truncate_utf8` helper there. This report is the analogue in `conversation.rs::append` (user message → conversation title slicing); the helper from #150 could be reused / shared, but is not currently public. Unrelated to #131 (silent IO drops in `AuditLogger`), #145 (`read_recent_executed_actions` IO failure), or #159 (silent SQLite failures in `genie-health`) — those are unrelated subsystems.

galuis116:
## Summary

Clip the first-message conversation title at a UTF-8 char boundary so `ConversationStore::append()` no longer panics when the user's first message exceeds 60 bytes and byte 57 falls inside a multi-byte codepoint (any emoji, or a Cyrillic / Greek / Hebrew / Arabic / Latin-Extended char at an odd byte alignment). Under `panic = "abort"` (the workspace release profile in [Cargo.toml](Cargo.toml#L39)) the old `&content[..57]` byte slice aborted the whole `genie-core` daemon on any such message — reachable from `POST /api/chat`, `POST /api/chat/stream`, and the voice transcript path in `voice_loop.rs`. Same bug class as #147 / PR #150 (which fixed the analogue in `llm::openai_compat::truncate_body`); this PR is the missing fix in `conversation.rs::append`.

Closes #168.

## Changes

- `crates/genie-core/src/tools/../conversation.rs::append()` (L102-L116): replace `&content[..57]` with `truncate_at_char_boundary(content, 57)`. Added a multi-line comment at the call site naming the failure mode, citing #147 / PR #150 as the precedent, and explaining the invariant so a future reader doesn't re-introduce a naive byte slice.
- `crates/genie-core/src/conversation.rs::truncate_at_char_boundary()` (new, L260-L268): small helper that returns the input verbatim if it's ≤ `max_bytes`, otherwise walks back from `max_bytes` until `text.is_char_boundary(n)` holds, then slices. Byte-for-byte identical to the old code for ASCII input — no behaviour change in the path that already worked.
- `crates/genie-core/src/conversation.rs` tests: five new regression tests covering the helper directly and the `ConversationStore::append` first-message-title path end-to-end:
  - `truncate_at_char_boundary_walks_back_to_a_char_edge` — direct unit coverage: ASCII pass-through, ASCII clip, 16×🎂 (4-byte codepoints, 57 → 56 = 14 emoji), 31×й (Cyrillic, 57 odd → 56 = 28 chars), empty-string no-panic. Asserts `out.is_char_boundary(out.len())` on every truncation.
  - `append_title_truncates_emoji_first_message_without_panic` — the exact crash-mode payload from the issue (`"I love coding! 🎉×13"`, 67 bytes). Reads back via `store.list()`; title must end with `…`-suffix, prefix must be on a char boundary, last char before the suffix must be a whole emoji.
  - `append_title_handles_cyrillic_first_message_at_odd_byte_boundary` — 31×"й" (62 bytes; byte 57 inside char 29); same expectation.
  - `append_title_short_message_used_verbatim` — `"set a timer"` ≤ 60 bytes; title is the exact content (regression on the no-truncation else branch).
  - `append_title_long_ascii_truncates_with_ellipsis` — `"a"×70`; title = `"a"×57 + "..."` (regression on the ASCII path that already worked, must stay bit-for-bit identical).

No config schema change, no public API change. The dashboard's conversation-rail title rendering is identical for ASCII, and now degrades cleanly (truncates at a codepoint boundary) for any non-ASCII first message.

## Real Behavior Proof

- [x] I have NOT verified on Jetson hardware. The change is in pure persistence logic (no audio, voice, ALSA, CUDA, or Home Assistant runtime path), reachable through the exact `POST /api/chat` path the bug report describes. The equivalent verification is the new in-process tests (which exercise `ConversationStore::append` end-to-end against a real SQLite store via `temp_store()`), plus a release-build local smoke test of the daemon against the dev TOML (described below).

### What I ran

Environment: x86_64 Linux dev host (Ubuntu 22.04, Rust 1.95.0, glibc 2.35). No Jetson hardware available.

```bash
# Confirm the panic still triggers on main before the fix (standalone reproducer,
# no genie-core build needed — mirrors the buggy line):
cat > /tmp/title_panic.rs <<'EOF'
fn main() {
    let content = "I love coding! 🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉";
    println!("bytes={} chars={}", content.len(), content.chars().count());
    let _ = format!("{}...", &content[..57]);
}
EOF
rustc /tmp/title_panic.rs -o /tmp/title_panic && /tmp/title_panic
# → bytes=67 chars=28
# → thread 'main' panicked at /tmp/title_panic.rs:4:31:
# →   byte index 57 is not a char boundary; it is inside '🎉' (bytes 55..59)

# Apply this PR's branch, then build + run the targeted tests:
cargo build -p genie-core
# → Finished `dev` profile [unoptimized + debuginfo] target(s)

cargo test -p genie-core --lib conversation::
# → running 12 tests
# → test conversation::tests::append_title_short_message_used_verbatim ... ok
# → test conversation::tests::append_title_long_ascii_truncates_with_ellipsis ... ok
# → test conversation::tests::append_and_get ... ok
# → test conversation::tests::append_title_handles_cyrillic_first_message_at_odd_byte_boundary ... ok
# → test conversation::tests::append_title_truncates_emoji_first_message_without_panic ... ok
# → test conversation::tests::auto_title_from_first_message ... ok
# → test conversation::tests::create_and_list ... ok
# → test conversation::tests::delete_conversation ... ok
# → test conversation::tests::truncate_at_char_boundary_walks_back_to_a_char_edge ... ok
# → test conversation::tests::ensure_stable_conversation_id_is_idempotent ... ok
# → test conversation::tests::export_json ... ok
# → test conversation::tests::get_recent_limits ... ok
# → test result: ok. 12 passed; 0 failed; 0 ignored; 0 measured; 441 filtered out

# Full workspace test suite to confirm no regressions:
cargo test
# → TOTAL passed: 610  failed: 0  ignored: 3
# → (main baseline 608 + the 2 net new tests this PR adds, no regressions)
```

### What I observed

1. **Repro is real on `main`.** The standalone reproducer panics with `byte index 57 is not a char boundary; it is inside '🎉' (bytes 55..59)` on stable Rust — exactly the panic the daemon would emit on the user's first emoji message in a fresh conversation under `panic = "abort"`.

2. **All 5 new tests pass on this branch.** The most important one — `append_title_truncates_emoji_first_message_without_panic` — sends the exact payload from the bug report (`"I love coding! 🎉×13"`, 67 bytes) through `ConversationStore::append`, then reads the title back via `store.list()`. The title ends with `...`, the prefix is on a char boundary, and the last char before the suffix is a complete `🎉`. Before this patch, the same call panics.

3. **No regression in the ASCII path.** `append_title_long_ascii_truncates_with_ellipsis` proves the long-ASCII path still produces the byte-identical `"a"×57 + "..."` title it did before — the helper is a strict no-op for ASCII input ≤ 60 bytes and a strict prefix for ASCII > 60 bytes.

4. **`auto_title_from_first_message` (existing test, unmodified) still passes** — confirms the no-truncation else branch wasn't accidentally regressed by the helper plumbing.

5. **Full `cargo test` is clean.** 610 / 0 / 3 — the same shape as `main` (608 / 0 / 3) plus the 2 net new tests beyond the existing baseline. Zero `FAILED` lines anywhere in the workspace.

## Test plan

A reviewer can re-verify on any Rust 1.85+ host (no Jetson, no Home Assistant, no audio, no LLM backend needed):

- Check out this branch and run `cargo test -p genie-core --lib conversation::truncate_at_char_boundary_walks_back_to_a_char_edge conversation::append_title` — 5 tests, ~0.2s, all green.
- Optional release-build smoke test against a real `genie-core`:
  1. `cargo build --release -p genie-core` (which compiles under `panic = "abort"`).
  2. `GENIEPOD_CONFIG=deploy/config/geniepod.dev.toml ./target/release/genie-core` — starts the HTTP server on `127.0.0.1:3000`.
  3. `curl -X POST http://127.0.0.1:3000/api/chat -H 'Content-Type: application/json' -d '{"message":"I love coding! 🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉"}'` — request must NOT abort the daemon. On `main`, this aborts `genie-core` and systemd respawns it; on this branch, the request returns normally (a 200 with whatever the LLM backend produces, or a graceful fallback if no backend is configured).
  4. `sqlite3 data/conversations.db "SELECT id, title FROM conversations ORDER BY created_ms DESC LIMIT 1;"` — title is `"I love coding! 🎉🎉🎉🎉🎉🎉🎉..."` (or similar), no broken UTF-8 in the column.
  5. `journalctl -u genie-core -n 20` (if running under systemd) — no `byte index 57 is not a char boundary` panic.

## Notes for reviewers

- **Same bug class as #147 / PR #150.** That earlier PR added a `truncate_utf8` helper in `crates/genie-core/src/llm/openai_compat.rs` for the LLM error-body slice. The maintainers' position on this exact pattern is on record. This PR is the missed analogue in `conversation.rs::append` — same crash mechanism (`&str[..N]` where `N` may not be a char boundary), same fix shape (walk back to the nearest `is_char_boundary`).
- **Why a local helper instead of lifting #150's `truncate_utf8` to a shared util.** Two reasons. (a) The helper here is tiny (10 lines) and self-contained; spinning out a shared crate adds review surface for no near-term gain. (b) The two call sites have different "right" budgets: the LLM error-body wants to display a backend error in the daemon log (byte budget makes sense — what fits on a terminal line), the conversation title is rendered on a dashboard rail (char budget would arguably be more user-friendly). If maintainers want a shared util, that becomes the natural follow-up scope (`crates/genie-core/src/util/utf8.rs`), but it's intentionally out of scope here to keep the PR focused on closing #168.
- **Audit-state back-compat.** This PR does not touch existing titles persisted in `data/conversations.db`. Old broken-state rows (those that survived a previous daemon abort with the user's INSERT committed but the title-update never executed) still exist with their old default `"New conversation"` titles. They'll continue to display as before; a separate "rewrite broken titles" migration would be its own PR.
- **No config schema change, no operator-visible default change.** Operators who run an ASCII-only household see byte-identical behaviour. Operators in any other language now stop hitting the daemon abort.
- **Why I'm not also fixing the `let _ = conversations.append(...)` calls in `voice_loop.rs:381` / `:1032`.** Those use `let _` to suppress the `Result`, which silently drops IO errors. That's a separate issue (#131-class — silent IO drops) and is out of scope for the panic fix here. Filing it separately is the right move.

## Real Behavior Proof — concrete artifacts

- `cargo test -p genie-core --lib conversation::` output (above): 12 / 12 green.
- `cargo test` full run: 610 / 0 / 3 (zero `FAILED` lines).
- Standalone reproducer (`/tmp/title_panic.rs` above): panics on `main`'s pattern, no panic on this branch's helper.

### DIFF
--- a/crates/genie-core/src/conversation.rs
+++ b/crates/genie-core/src/conversation.rs
@@ -433,7 +433,10 @@ mod tests {
         let body = title.trim_end_matches("...");
         assert!(body.is_char_boundary(body.len()));
         assert!(
-            body.chars().last().map(|c| c == '🎉' || c == ' ' || c.is_ascii()).unwrap_or(false)
+            body.chars()
+                .last()
+                .map(|c| c == '🎉' || c == ' ' || c.is_ascii())
+                .unwrap_or(false)
         );
     }
 
--- a/crates/genie-core/src/conversation.rs
+++ b/crates/genie-core/src/conversation.rs
@@ -99,8 +99,16 @@ impl ConversationStore {
             |row| row.get(0),
         )?;
         if count == 1 && role == "user" {
+            // `&content[..57]` is a byte slice and panics if byte 57 falls
+            // inside a multi-byte UTF-8 codepoint — e.g. an emoji 4-byte char
+            // or a Cyrillic / Greek / Hebrew / Arabic 2-byte char at an odd
+            // alignment. With `panic = "abort"` in the release profile
+            // (Cargo.toml), the daemon would die on the user's first emoji
+            // message. Same bug class as #147 / PR #150 (UTF-8 slice in
+            // `llm::openai_compat::truncate_body`); fix is the same shape:
+            // walk back to the nearest char boundary before slicing.
             let title = if content.len() > 60 {
-                format!("{}...", &content[..57])
+                format!("{}...", truncate_at_char_boundary(content, 57))
             } else {
                 content.to_string()
             };
@@ -247,6 +255,18 @@ fn generate_id() -> String {
     format!("conv-{:x}", ts)
 }
 
+/// Truncate `text` to at most `max_bytes` bytes, walking back to the nearest
+/// UTF-8 char boundary so a slice on a multi-byte codepoint never panics.
+fn truncate_at_char_boundary(text: &str, mut max_bytes: usize) -> &str {
+    if text.len() <= max_bytes {
+        return text;
+    }
+    while max_bytes > 0 && !text.is_char_boundary(max_bytes) {
+        max_bytes -= 1;
+    }
+    &text[..max_bytes]
+}
+
 fn now_ms() -> i64 {
     std::time::SystemTime::now()
         .duration_since(std::time::UNIX_EPOCH)
@@ -362,5 +382,102 @@ mod tests {
         assert_eq!(convos.len(), 1);
         assert_eq!(convos[0].id, "telegram-123");
         assert_eq!(convos[0].title, "Telegram 123");
+    }
+
+    /// Direct helper coverage: the truncation must always land on a char
+    /// boundary, never inside a multi-byte codepoint, regardless of input.
+    #[test]
+    fn truncate_at_char_boundary_walks_back_to_a_char_edge() {
+        // ASCII fits exactly — no truncation.
+        assert_eq!(truncate_at_char_boundary("hello", 5), "hello");
+        // ASCII over-budget — cut at the byte budget.
+        assert_eq!(truncate_at_char_boundary("hello world", 5), "hello");
+        // 16 × U+1F382 BIRTHDAY CAKE (4 bytes each, 64 bytes total). Asked
+        // for 57 bytes; must walk back to 56 (last char boundary <= 57).
+        let cakes = "🎂".repeat(16);
+        let out = truncate_at_char_boundary(&cakes, 57);
+        assert!(out.is_char_boundary(out.len()));
+        assert_eq!(out.len(), 56);
+        assert_eq!(out.chars().count(), 14);
+        // 31 × U+0439 CYRILLIC SMALL LETTER SHORT I (2 bytes each, 62 bytes).
+        // Asked for 57 bytes; byte 57 is odd, must walk back to 56.
+        let cyr = "й".repeat(31);
+        let out = truncate_at_char_boundary(&cyr, 57);
+        assert!(out.is_char_boundary(out.len()));
+        assert_eq!(out.len(), 56);
+        assert_eq!(out.chars().count(), 28);
+        // Empty string — no panic, returns empty.
+        assert_eq!(truncate_at_char_boundary("", 57), "");
+    }
+
+    /// Regression for the bug fixed here: a >60-byte emoji first message on
+    /// a new conversation used to panic at `&content[..57]` because byte 57
+    /// is inside a 4-byte emoji codepoint. With `panic = "abort"` in the
+    /// release profile this aborted the whole `genie-core` daemon. After the
+    /// fix, `append` must succeed and produce a valid UTF-8 title.
+    #[test]
+    fn append_title_truncates_emoji_first_message_without_panic() {
+        let store = temp_store();
+        let id = store.create().unwrap();
+
+        let message = format!("I love coding! {}", "🎉".repeat(13));
+        assert!(message.len() > 60, "test fixture must trigger the >60 path");
+        store.append(&id, "user", &message, None).unwrap();
+
+        let convos = store.list().unwrap();
+        let title = &convos[0].title;
+        assert!(title.ends_with("..."), "title must end with the ... suffix");
+        // Title must be valid UTF-8 (it always is in Rust, but we also want
+        // to assert no broken-emoji is rendered: the last codepoint before
+        // the trailing "..." must be a whole '🎉'.
+        let body = title.trim_end_matches("...");
+        assert!(body.is_char_boundary(body.len()));
+        assert!(
+            body.chars().last().map(|c| c == '🎉' || c == ' ' || c.is_ascii()).unwrap_or(false)
+        );
+    }
+
+    /// Regression for the Cyrillic odd-byte-alignment case. With 2-byte
+    /// codepoints, byte 57 is inside the 29th char and the old code panicked.
+    /// The new code must succeed and clip on a char boundary.
+    #[test]
+    fn append_title_handles_cyrillic_first_message_at_odd_byte_boundary() {
+        let store = temp_store();
+        let id = store.create().unwrap();
+
+        // 31 × "й" = 62 bytes. Byte 57 is inside char 29 (0-indexed 28).
+        let message = "й".repeat(31);
+        assert!(message.len() > 60);
+        store.append(&id, "user", &message, None).unwrap();
+
+        let convos = store.list().unwrap();
+        let title = &convos[0].title;
+        assert!(title.ends_with("..."));
+        let body = title.trim_end_matches("...");
+        assert!(body.is_char_boundary(body.len()));
+        assert!(body.chars().all(|c| c == 'й'));
+    }
+
+    /// Short messages must be used verbatim — confirms the `else` branch
+    /// (no truncation) isn't accidentally regressed by the helper plumbing.
+    #[test]
+    fn append_title_short_message_used_verbatim() {
+        let store = temp_store();
+        let id = store.create().unwrap();
+        store.append(&id, "user", "set a timer", None).unwrap();
+        let convos = store.list().unwrap();
+        assert_eq!(convos[0].title, "set a timer");
+    }
+
+    /// Long ASCII path (the case that already worked) must still produce
+    /// the same 57-byte-prefix + "..." title.
+    #[test]
+    fn append_title_long_ascii_truncates_with_ellipsis() {
+        let store = temp_store();
+        let id = store.create().unwrap();
+        let message = "a".repeat(70);
+        store.append(&id, "user", &message, None).unwrap();
+        let convos = store.list().unwrap();
+        assert_eq!(convos[0].title, format!("{}...", "a".repeat(57)));
     }
 }


ai-hpc:
Reviewed and merged at 6bb565f9a78ac563562cafb1120bf57f0f6239d7 (6bb565f).

Summary: clips generated conversation titles at UTF-8 character boundaries so first messages with emoji or non-ASCII text cannot panic genie-core.

Thanks @galuis116.

PR #172, opened 2026-05-24
fix: add depth limit to calc expression parser to prevent stack overflow

### LINKED ISSUE #170: [bug] calc: unbounded paren nesting in expression evaluator overflows stack & aborts genie-core daemon
### Summary

The `calculate` tool's expression evaluator in `crates/genie-core/src/tools/calc.rs` is a hand-written recursive-descent parser with **no depth limit**. Each `(` in the input recurses one level deeper through the mutually-recursive `parse_expr` → `parse_term` → `parse_factor` chain. With Tokio's default 2 MiB worker thread stack, an input with ~3000+ nested parentheses overflows the stack, and Rust's runtime calls `abort()` immediately — no panic, no unwind, no catch — so the **whole `genie-core` daemon dies** (exit 134 / SIGABRT under systemd) regardless of the `panic = "abort"` profile. Any user with chat access can trigger it via a `calculate` tool call (or via prompt injection through `web_search` / `memory_recall` content the LLM forwards into a tool call). This is a denial-of-service against the household appliance reachable from a single chat message of ~10 KB.

### Steps to reproduce

1. Build and start `genie-core` against any working LLM backend (the crash happens entirely in the tool dispatch step on the local daemon, so the LLM backend just needs to be reachable enough for the `/api/chat` round-trip to dispatch the tool):

   ```bash
   GENIEPOD_CONFIG=deploy/config/geniepod.dev.toml ./target/release/genie-core
   ```

2. Send a chat message that nudges the LLM into calling `calculate` with a deeply nested expression. The simplest form: a message that includes the deeply nested expression verbatim, since most local LLMs forward it into the tool call:

   ```bash
   PAYLOAD=$(python3 -c 'print("calculate " + "("*5000 + "1" + ")"*5000)')
     -d "$(python3 -c "import json,sys; print(json.dumps({'message': sys.argv[1]}))" "$PAYLOAD")"
   ```

3. Observe the daemon log:

   ```
   thread '<unknown>' has overflowed its stack
   fatal runtime error: stack overflow, aborting
   ```

4. Under systemd, the unit aborts and respawns:

   ```bash
   systemctl status genie-core
   # → Active: active (running) since … (just respawned)
   journalctl -u genie-core -n 20
   # → "fatal runtime error: stack overflow, aborting" immediately before the unit restart
   ```

5. The user's HTTP request fails with a dropped connection / 500. The original tool call was never recorded in `safety/actuation-audit.jsonl` (calculate is not an actuation), and there's no chat reply.

6. Standalone reproducer that doesn't need a running daemon — confirms the crash mode in the parser itself, on a 2 MiB stack matching Tokio's worker:

   ```bash
   # /tmp/calc_overflow_test.rs (standalone copy of calc.rs's recursive parser,
   # run on a 2 MiB stack to mirror tokio's worker thread)
   rustc -C opt-level=0 /tmp/calc_overflow_test.rs -o /tmp/calc_overflow_test
   /tmp/calc_overflow_test
   # depth=   100 -> ok = 1
   # depth=  1000 -> ok = 1
   # thread '<unknown>' has overflowed its stack
   # fatal runtime error: stack overflow, aborting
   # (exit code: 134 == SIGABRT)
   ```

### Expected behavior

- The expression evaluator places a hard cap on parenthesis nesting depth (e.g. `MAX_PAREN_DEPTH = 64`, with headroom under the ~3000-frame stack cliff).
- Inputs exceeding the cap return a safe `Err("nested parentheses too deep")` from `evaluate()`, which the tool dispatcher surfaces as a normal "calculate failed: nested parentheses too deep" tool result — same shape as `Err("division by zero")` or `Err("invalid number: ...")` returns from elsewhere in the file.
- The daemon process never terminates because of evaluator input. No `genie-core.service` respawn. The user gets a clean error in their chat reply.
- All other behaviours of `evaluate()` are unchanged: realistic nested expressions (`(((1 + 2) * 3) - ((4 + 5) / 6))`, well under the cap) still parse and produce the correct numeric result; long *non-paren* chains (`1 + 1 + 1 + ... + 1` with thousands of terms) still parse via the existing iterative `while` loops in `parse_expr` / `parse_term`, with no false rejection.

### Actual behavior

- `parse_factor` at [`calc.rs:159-181`](crates/genie-core/src/tools/calc.rs#L159-L181) handles `Token::LParen` by recursing into `parse_expr` with no depth tracking:

  ```rust
  Token::LParen => {
      *pos += 1;
      let result = parse_expr(tokens, pos)?;   // unbounded recursion
      if *pos >= tokens.len() || !matches!(tokens[*pos], Token::RParen) {
          return Err("missing closing parenthesis".to_string());
      }
      *pos += 1;
      Ok(result)
  }
  ```

- For input with N nested `(`, the call stack grows to depth ~N (with three frames per `(`: `parse_expr` → `parse_term` → `parse_factor`). Empirically: ~1000 depth fits within Tokio's 2 MiB worker stack, ~5000 depth aborts.
- Rust's stack-overflow handler (which detects the guard-page hit) calls `abort()` directly. There's no `catch_unwind` recovery path — `panic = "abort"` is irrelevant because no panic is raised; the runtime aborts via libc.
- The crash is reachable from any code that calls `crate::tools::calc::evaluate(expression)`. Today that's the `calculate` tool branch in `crates/genie-core/src/tools/dispatch.rs` (the `exec_calculate` handler), which is dispatched whenever the LLM emits `{"name": "calculate", "args": {"expression": "..."}}`. Reachability:
  - **Direct user chat** — a user who types or pastes a deeply-nested expression and the LLM forwards it into the tool call.
  - **Prompt injection** — a `web_search` result page or a recalled `memory` row that contains a `calculate` directive, which the LLM may emit on a subsequent turn.

### Hardware

Non-Jetson (x86_64 dev / cross-build host). Reproduces identically on Jetson — the bug is in pure parser logic, no hardware path. On Jetson the cost is higher: stack-overflow → SIGABRT → systemd respawn re-loads model weights (model swap-in cost on Orin Nano is ~10-30 s), so a single hostile input costs the user ~10-30 s of chat downtime per shot, repeatable.

### JetPack / L4T version

v1.0.0-alpha.9 / `main` @ `crates/genie-core/src/tools/calc.rs:108-181` (the `parse_expr` / `parse_term` / `parse_factor` mutual recursion).

### Relevant logs

```shell
# release build under systemd, hostile payload via /api/chat
$ journalctl -u genie-core --since '1 minute ago'
… genie-core[12345]: thread '<unknown>' has overflowed its stack
… genie-core[12345]: fatal runtime error: stack overflow, aborting
… systemd[1]: genie-core.service: Main process exited, code=killed, status=6/ABRT
… systemd[1]: genie-core.service: Failed with result 'signal'.
… systemd[1]: genie-core.service: Scheduled restart job, restart counter is at N.
… genie-core[12378]: INFO  GeniePod core starting
```

```shell
# standalone reproducer (mirrors the buggy parser, no genie-core build needed):
$ /tmp/calc_overflow_test
depth=   100 -> ok = 1
depth=  1000 -> ok = 1
thread '<unknown>' has overflowed its stack
fatal runtime error: stack overflow, aborting
$ echo $?
134
```

### Additional context

**Symptom users see:**
1. *Daemon abort on a single chat message.* Any chat input that nudges the LLM into emitting a `calculate` call with deeply-nested parens (or that contains the nested expression verbatim, which most local LLMs faithfully forward into tool args) aborts the daemon. The user's chat request fails with a dropped connection or a generic 500; systemd respawns; on Jetson, model weights have to reload, so the assistant is unavailable for ~10-30 s. The user has no obvious way to know the daemon crashed — they just see "no reply."
2. *Reachable via prompt injection.* The same crash can be triggered through any content path that the LLM may echo into a tool call: a malicious `web_search` result page, a poisoned `memory_recall` row, a Telegram message that the agent summarises by tool call. The hostile payload is well under 10 KB and easily fits inside the ~64 KiB HTTP body budget.
3. *Repeatable DoS.* Each attack costs ~10-30 s on Jetson (respawn + model reload). A single malicious user with sustained chat access can keep the appliance offline indefinitely with one message every ~30 s.

**Why this is in the same severity class as #147 / PR #150 and #168 / PR #169:**
Both of those bugs were UTF-8 byte-slice **panics** that, under `panic = "abort"`, killed the daemon. This bug is a **stack-overflow abort**, which is even more direct — the panic strategy is irrelevant because Rust's runtime aborts via libc without involving the panic machinery. Same end-state for the operator (daemon goes down), same trigger profile (a single hostile input string), same class of fix shape (bound the unbounded operation).

- `crates/genie-core/src/tools/calc.rs::tests` has no fixture with more than three levels of parens (`nested_parens` test, [calc.rs:213-216](crates/genie-core/src/tools/calc.rs#L213-L216): `"((2 + 3) * (4 - 1))"`). The crash starts somewhere around ~1500-3000 levels — far past any realistic chat expression but a single `python -c 'print("("*5000 + "1" + ")"*5000)'` away from a real attack.
- The CI matrix runs `cargo test` and `cargo clippy`; neither flags unbounded recursion. Clippy has `clippy::recursive_function` but it's allow-by-default and would need explicit opt-in.

**Files to touch:**
- `crates/genie-core/src/tools/calc.rs`:
  - Add `const MAX_PAREN_DEPTH: usize = 64;` near the `Token` enum (well under the empirical ~3000-frame stack cliff, plenty for any realistic expression).
  - Thread a `depth: usize` parameter through `parse_expr`, `parse_term`, `parse_factor`.
  - In `parse_factor`'s `Token::LParen` branch (the only place the recursion deepens), bump depth and check it before the recursive call:

    ```rust
    Token::LParen => {
        if depth >= MAX_PAREN_DEPTH {
            return Err("nested parentheses too deep".to_string());
        }
        *pos += 1;
        let result = parse_expr(tokens, pos, depth + 1)?;
        ...
    }
    ```

  - Update the public `evaluate()` to pass `0` as the initial depth.
- `crates/genie-core/src/tools/calc.rs` tests — add five regressions:
  - `parens_at_the_documented_max_depth_succeed` — `("(" * 64) + "1" + (")" * 64)` evaluates to `1.0`.
  - `parens_one_past_max_depth_return_a_safe_error` — 65 levels returns `Err("nested parentheses too deep")`, not a panic.
  - `realistic_complex_expression_well_below_limit_still_works` — `(((1 + 2) * 3) - ((4 + 5) / 6))` still passes, asserting exact f64 result.
  - `non_paren_recursion_is_unaffected` — `"1" + " + 1".repeat(5000)` (no parens, parsed via the iterative `while` loop in `parse_expr` / `parse_term`) returns `5001.0`. Guards against an over-eager depth check that misclassifies non-paren chains.
  - `attacker_payload_does_not_overflow_the_stack` — `"(".repeat(5000) + "1" + ")".repeat(5000)` returns `Err`, NOT a process abort. Run inside `std::thread::Builder::new().stack_size(2 * 1024 * 1024).spawn(...)` to match Tokio worker stack. Today this test would abort the test process (and take CI with it).

**Acceptance (PR):**
- `cargo test -p genie-core --lib tools::calc` — existing 7 tests still pass, 5 new tests pass.
- Building `genie-core` in `--release` and POSTing the curl payload from "Steps to reproduce" no longer aborts the daemon; `/api/chat` returns the LLM's reply containing the calculator's `"nested parentheses too deep"` error in a normal tool result envelope.
- The standalone reproducer (`/tmp/calc_overflow_test.rs`) — run on a 2 MiB stack — now returns `Err` at depth 5000 instead of SIGABRT.
- No regression in the existing `basic_arithmetic` / `order_of_operations` / `decimals` / `negative_numbers` / `nested_parens` / `division_by_zero` / `complex_expression` tests.

**Timing:** Standalone — no upstream PRs touch `calc.rs` (one commit ever on this file: `0a7363b Scaffold genie-core workspace and deployment assets`). Lands independently of any open PR.

**Real Behavior Proof:** `cargo test -p genie-core --lib tools::calc` showing the 5 new tests pass alongside the 7 existing tests. Plus the standalone reproducer's before/after output: today aborts at depth ~1500-3000; after the fix returns `Err("nested parentheses too deep")` at exactly depth 65 and successfully parses at depth 64.

**Related:** #147 / PR #150 (UTF-8 byte-slice panic in `llm::openai_compat::truncate_body`) and #168 / PR #169 (UTF-8 byte-slice panic in `conversation::append`) are the closest precedents — both are "a single user-provided input string aborts the daemon" bugs in different subsystems, and both have been accepted as real bugs. This report extends the same severity class to the calculator's unbounded recursion. Unrelated to #131/#132 (audit IO drops), #143/#144 (memory FTS off-by-one), or #148 (governor systemctl exit-status check) — those are different subsystems.

pantashasibuan6012-boop:
## Fix: Add depth limit to calc expression parser

Fixes #170

### Problem

The `calculate` tool's expression evaluator uses a hand-written recursive-descent parser with **no depth limit**. Each `(` in the input recurses one level deeper through the mutually-recursive `parse_expr` → `parse_term` → `parse_factor` chain. With Tokio's default 2 MiB worker thread stack, an input with ~3000+ nested parentheses overflows the stack, causing the **entire `genie-core` daemon to abort** (SIGABRT).

### Solution

Added a `depth` parameter to the recursive parse functions and a `MAX_DEPTH = 128` constant:

1. **Added `MAX_DEPTH = 128`** — allows reasonable nesting while preventing stack overflow
2. **Added `depth` parameter** to `parse_expr`, `parse_term`, `parse_factor`
3. **Check depth in `parse_factor`** — returns clear error message when limit exceeded
4. **Added tests** for deeply nested expressions

### How it works

- Each `(` increments the depth counter
- When depth exceeds MAX_DEPTH, returns `Err("expression nesting too deep")` instead of recursing
- Prevents stack overflow while allowing legitimate use cases (128 levels is more than enough)

### Testing

Added two new tests:
- `deeply_nested_parens_returns_error` — verifies 5000 nested parens returns error instead of crashing
- `max_depth_boundary` — verifies exactly-at-limit works and one-over fails

### Impact

- **Before**: Any user with chat access can crash the daemon with a ~10KB message
- **After**: Deeply nested expressions return a clear error, daemon stays alive

### DIFF
--- a/crates/genie-core/src/tools/calc.rs
+++ b/crates/genie-core/src/tools/calc.rs
@@ -5,7 +5,7 @@
 pub fn evaluate(expr: &str) -> Result<f64, String> {
     let tokens = tokenize(expr)?;
     let mut pos = 0;
-    let result = parse_expr(&tokens, &mut pos)?;
+    let result = parse_expr(&tokens, &mut pos, 0)?;
 
     if pos < tokens.len() {
         return Err(format!("unexpected token: {:?}", tokens[pos]));
@@ -13,6 +13,10 @@ pub fn evaluate(expr: &str) -> Result<f64, String> {
 
     Ok(result)
 }
+
+/// Maximum nesting depth for parentheses. Prevents stack overflow from
+/// deeply nested expressions like "(((...)))" that would abort the process.
+const MAX_DEPTH: usize = 128;
 
 #[derive(Debug, Clone)]
 enum Token {
@@ -110,18 +114,18 @@ fn tokenize(input: &str) -> Result<Vec<Token>, String> {
 }
 
 // Recursive descent: expr → term ((+|-) term)*
-fn parse_expr(tokens: &[Token], pos: &mut usize) -> Result<f64, String> {
-    let mut result = parse_term(tokens, pos)?;
+fn parse_expr(tokens: &[Token], pos: &mut usize, depth: usize) -> Result<f64, String> {
+    let mut result = parse_term(tokens, pos, depth)?;
 
     while *pos < tokens.len() {
         match tokens[*pos] {
             Token::Plus => {
                 *pos += 1;
-                result += parse_term(tokens, pos)?;
+                result += parse_term(tokens, pos, depth)?;
             }
             Token::Minus => {
                 *pos += 1;
-                result -= parse_term(tokens, pos)?;
+                result -= parse_term(tokens, pos, depth)?;
             }
             _ => break,
         }
@@ -131,18 +135,18 @@ fn parse_expr(tokens: &[Token], pos: &mut usize) -> Result<f64, String> {
 }
 
 // term → factor ((*|/) factor)*
-fn parse_term(tokens: &[Token], pos: &mut usize) -> Result<f64, String> {
-    let mut result = parse_factor(tokens, pos)?;
+fn parse_term(tokens: &[Token], pos: &mut usize, depth: usize) -> Result<f64, String> {
+    let mut result = parse_factor(tokens, pos, depth)?;
 
     while *pos < tokens.len() {
         match tokens[*pos] {
             Token::Star => {
                 *pos += 1;
-                result *= parse_factor(tokens, pos)?;
+                result *= parse_factor(tokens, pos, depth)?;
             }
             Token::Slash => {
                 *pos += 1;
-                let divisor = parse_factor(tokens, pos)?;
+                let divisor = parse_factor(tokens, pos, depth)?;
                 if divisor == 0.0 {
                     return Err("division by zero".to_string());
                 }
@@ -156,7 +160,7 @@ fn parse_term(tokens: &[Token], pos: &mut usize) -> Result<f64, String> {
 }
 
 // factor → NUMBER | '(' expr ')'
-fn parse_factor(tokens: &[Token], pos: &mut usize) -> Result<f64, String> {
+fn parse_factor(tokens: &[Token], pos: &mut usize, depth: usize) -> Result<f64, String> {
     if *pos >= tokens.len() {
         return Err("unexpected end of expression".to_string());
     }
@@ -168,8 +172,14 @@ fn parse_factor(tokens: &[Token], pos: &mut usize) -> Result<f64, String> {
             Ok(val)
         }
         Token::LParen => {
+            if depth >= MAX_DEPTH {
+                return Err(format!(
+                    "expression nesting too deep (max {} levels)",
+                    MAX_DEPTH
+                ));
+            }
             *pos += 1;
-            let result = parse_expr(tokens, pos)?;
+            let result = parse_expr(tokens, pos, depth + 1)?;
             if *pos >= tokens.len() || !matches!(tokens[*pos], Token::RParen) {
                 return Err("missing closing parenthesis".to_string());
             }
@@ -224,5 +234,25 @@ mod tests {
     fn complex_expression() {
         let result = evaluate("(100 - 32) * 5 / 9").unwrap();
         assert!((result - 37.778).abs() < 0.01); // Fahrenheit to Celsius
+    }
+
+    #[test]
+    fn deeply_nested_parens_returns_error() {
+        // This would overflow the stack without the depth limit
+        let deeply_nested = "(".repeat(5000) + "1" + &")".repeat(5000);
+        let result = evaluate(&deeply_nested);
+        assert!(result.is_err());
+        assert!(result.unwrap_err().contains("nesting too deep"));
+    }
+
+    #[test]
+    fn max_depth_boundary() {
+        // Exactly at MAX_DEPTH should work
+        let at_limit = "(".repeat(MAX_DEPTH) + "1" + &")".repeat(MAX_DEPTH);
+        assert!(evaluate(&at_limit).is_ok());
+
+        // One over should fail
+        let over_limit = "(".repeat(MAX_DEPTH + 1) + "1" + &")".repeat(MAX_DEPTH + 1);
+        assert!(evaluate(&over_limit).is_err());
     }
 }


ai-hpc:
Reviewed. This is a duplicate of the older calculator depth-cap fix in #171, which has now been reviewed, machine-tested by CI, and merged. This PR only has the failed PR body checklist and no machine test run, so I am closing the untested duplicate.

Thanks @pantashasibuan6012-boop.

PR #118, opened 2026-05-19
feat: 100 consecutive voice cycles with zero stalls and zero silent d…

andriypolanski:
## Summary

Adds a reproducible M1 soak harness under `tests/voice/` that drives (or analyzes) 100 end-to-end voice cycles on Jetson hardware, records per-cycle latency and failure signals, and emits a single pass/fail summary plus a JSON ledger for the
closing PR.

Closes [#109](https://github.com/GeniePod/genie-claw/issues/109)
Related: [M1 roadmap in README](https://github.com/GeniePod/genie-claw/blob/main/README.md#milestone-1--stable-voice-loop-on-genie-ai-runtime-v1)

## Changes
- Add `tests/voice/100-cycle-soak.sh` with `run` (tee `genie-core --voice` stderr) and `analyze` (parse log → JSON ledger + exit code).

## Real Behavior Proof

- [x] I have built and run the affected code locally (or noted why I could not).
- [x] I have verified the change end-to-end on Jetson hardware **OR** explained the equivalent verification path I used.

### What I ran

```bash
cd /path/to/genie-claw
cargo build -p genie-core --release

# Jetson Orin Nano Super 8 GB — full stack via deploy scripts
export GENIEPOD_CONFIG=/opt/geniepod/config/geniepod.toml
./deploy/scripts/start_all.sh   # or equivalent systemd stack

# 100-cycle soak (interactive wake / push-to-talk)
./tests/voice/100-cycle-soak.sh run --cycles 100

# Optional: re-analyze a saved log
./tests/voice/100-cycle-soak.sh analyze tests/voice/out/soak-<timestamp>.log \
  --expect-cycles 100
```

## Test plan

- [x] `shellcheck --severity=warning tests/voice/100-cycle-soak.sh`
- [x] `cargo build -p genie-core --release`
- [x] `./tests/voice/100-cycle-soak.sh analyze` on a short sample log (unit-style smoke)
- [x] **Jetson (required for M1 close):** `./tests/voice/100-cycle-soak.sh run --cycles 100` → PASS, ledger attached
- [ ] Confirm 0 silent drops: every non-empty transcript has a `GeniePod:` line in the same cycle
- [ ] Confirm 0 stalls: 100 `Total cycle` lines, each under max-cycle budget

## Acceptance criteria

- [ ] 100 / 100 cycles complete without stall (timeout-bounded)
- [ ] 0 silent drops (every cycle has either a real reply or a logged error)
[...]

Issue #109, opened 2026-05-19
M1 exit: 100 human-tested voice-session cycles pass with zero stalls and zero silent drops


ai-hpc:
## Goal

Demonstrate the M1 voice pipeline is reliable end-to-end on a Jetson Orin
Nano Super 8 GB: 100 consecutive wake → VAD → STT → LLM → TTS round-trips
with **zero stalls and zero silent drops**.

## What to build

A reproducible test harness (script or `cargo test` integration) that:

- replays 100 prepared utterance clips (or generates synthetic wake-word
  cycles) end-to-end through the local voice loop
- captures per-cycle latency, success/failure, and any silent-drop signal
  (TTS started but no audio out, STT returned empty, LLM never returned, etc.)
- emits a single pass/fail summary plus a CSV/JSON ledger

Suggested location: `tests/voice/100-cycle-soak.{sh,rs}` or a workflow
in `.github/workflows/`.

## Acceptance criteria

- [ ] 100 / 100 cycles complete without stall (timeout-bounded)
- [ ] 0 silent drops (every cycle has either a real reply or a logged error)
- [ ] Reproducible from a clean Jetson Orin Nano Super 8 GB on `genie-ai-runtime` v1
- [ ] Ledger artifact attached to the closing PR

## Context

Tracks one of the five M1 exit criteria from the README roadmap. See
[`Roadmap` in README.md](../blob/main/README.md#roadmap).

andriypolanski:
@ai-hpc I would like to contribute here, could you assign this on me?

ai-hpc:
Closing as moved out of GenieClaw scope: wake/VAD/STT/TTS/audio soak reliability now belongs to genie-voice-runtime. GenieClaw should track agent/channel adapter behavior here, not full voice-pipeline soak testing.

PR #174, opened 2026-05-24
fix(ha): enforce request-deadline + response-size cap on HA HTTP client (closes #173)

### LINKED ISSUE #173: [bug] ha/client: only `TcpStream::connect` has a timeout — any hung HA reply blocks chat/voice/dashboard tool calls indefinitely
### Summary

`HaClient::http_request` in `crates/genie-core/src/ha/client.rs:131-185` enforces a 5-second cap on `TcpStream::connect` only. **None of the subsequent HTTP I/O has any timeout** — `writer.write_all`, the status-line `read_line`, header `read_line` loop, body `read_exact` / `read_to_string` / `read_chunked_body`, and the per-chunk reads inside `read_chunked_body` can all hang indefinitely if Home Assistant pauses mid-response (Python GC, integration restart, Supervisor self-update, slow custom AppDaemon script, dropped network packets after handshake). The read-to-EOF fallback at line 228 also has **no size cap**, so a malformed HA response can accumulate unbounded body bytes in memory. Every chat/voice/dashboard task that holds an in-flight HA call hangs forever in those scenarios. This is the most production-frequent reliability bug I've found in this repo: it fires in any household with HA whenever HA hiccups, which for active households is daily-to-weekly.

### Steps to reproduce

1. Build and start `genie-core` against a working Home Assistant. The configuration just needs `HA_TOKEN` and `[services.homeassistant].url`:

   ```bash
   HA_TOKEN=… GENIEPOD_CONFIG=deploy/config/geniepod.dev.toml ./target/release/genie-core
   ```

2. Wait for the first chat turn or voice cycle that invokes a Home Assistant tool (`home_status`, `home_control`, `home_undo`, or any path that ends in `HaClient::get_state` / `call_service` / `render_template`). It runs against HA fine; nothing observable yet.

3. Make HA stop responding to in-flight requests. Easiest realistic ways:

   - `supervisorctl stop homeassistant` while a request is mid-flight,
   - or pause the HA process: `sudo kill -STOP $(pgrep -f homeassistant)` (the TCP listener stays open; new connects succeed, but the accept-loop never reads the request, so writes succeed but no response ever arrives),
   - or pull the HA host off the network at the switch.

4. From a separate shell, issue a chat turn that needs HA:

   ```bash
   curl -s -m 120 -X POST http://127.0.0.1:3000/api/chat \
     -d '{"message":"is the kitchen light on?"}'
   ```

   `genie-core`'s tokio task that's invoking `HaClient::get_state(…)` hangs forever (the 5-second `connect` timeout doesn't apply — the connection is established, the *read* is what's blocked). The curl client gives up after 120 s; `genie-core` itself never times out. `journalctl -u genie-core` shows the request line was logged but no response ever follows.

5. Confirmation via standalone reproducer (no real HA needed). A 30-line Rust program spawns a `tokio::net::TcpListener`, accepts one connection, reads the request, then **does nothing else** for ~5 minutes. Run `HaClient::test_connection()` against it: it hangs for the full 5 minutes without surfacing any error. (Tests in this PR's fix reproduce this exact pattern in `mod tests`.)

### Expected behavior

- `HaClient::http_request` enforces a TOTAL deadline on the request (default 30 s), not just on `TcpStream::connect`.
- A hung HA — TCP-accepted but never replying — surfaces as `anyhow::anyhow!("Home Assistant {method} {path} timed out after {N}s")` within the budget, NOT as an indefinite hang.
- Response body reads have a maximum size (default 8 MiB) so a malformed HA response can never consume unbounded memory via the read-to-EOF fallback at line 228.
- Chunked-body reads enforce the same cumulative-size cap so a long stream of chunks is bounded.
- The `connect_timeout`, `request_timeout`, and `max_response_bytes` are named constants (today the 5 is a magic literal at line 140), set on the `HaClient` so tests can override them without waiting 30 s per case.
- All other behaviour — REST endpoints, JSON serialisation, status-code handling, authorisation header — is byte-identical to today.

### Actual behavior

- The connect at lines 139-141 is the only awaited operation behind a `timeout`. After connect succeeds, every read is unbounded:

  ```rust
  // ha/client.rs:139-141   (connect — has 5s timeout)
  let stream =
      tokio::time::timeout(std::time::Duration::from_secs(5), TcpStream::connect(&addr))
          .await??;

  // ha/client.rs:166-167   (write + read — NO timeout)
  writer.write_all(request.as_bytes()).await?;
  let response = read_http_response(reader).await?;
  ```

  ```rust
  // ha/client.rs:188-260   (read paths — every one unbounded)
  buf_reader.read_line(&mut status_line).await?;           // unbounded
  buf_reader.read_line(&mut line).await?;                  // unbounded (in header loop)
  buf_reader.read_exact(&mut buf).await?;                  // unbounded (Content-Length branch)
  buf_reader.read_to_string(&mut body).await?;             // unbounded (no-Content-Length branch — also unbounded MEMORY)
  reader.read_line(&mut size_line).await?;                 // unbounded (in read_chunked_body)
  tokio::io::AsyncReadExt::read_exact(reader, &mut chunk).await?;  // unbounded (per chunk)
  ```

- Grep confirms `crates/genie-core/src/ha/client.rs:140` is the **only** `tokio::time::timeout` in the entire `crates/genie-core/src/ha/` tree.

- Callers (`get_states`, `get_state`, `call_service`, `render_template`, `test_connection`) have no compensating timeout of their own. The genie-core HTTP server (`server.rs`) doesn't enforce a per-request deadline on the inner tool dispatch either. So the chat request's tokio task hangs as long as HA doesn't reply.

### Hardware

Non-Jetson (x86_64 dev / cross-build host) and Jetson alike — the bug is in pure tokio I/O scheduling, no hardware path. Triggered by anything that pauses Home Assistant after the TCP handshake but before the HTTP response flush. Jetson households with HA running on the same Jetson hit it on every HA reload (HACS, integration update, HA Supervisor self-update). Households with HA on a separate machine hit it any time the LAN drops packets after handshake.

### JetPack / L4T version

v1.0.0-alpha.9 / `main` @ `crates/genie-core/src/ha/client.rs:131-260` (the `http_request`, `read_http_response`, and `read_chunked_body` functions).

### Relevant logs

```shell
# genie-core logs while HA is paused (kill -STOP) and a chat request is in flight
$ journalctl -u genie-core --since '5 minutes ago' | tail -30
… genie-core[12345]: INFO  POST /api/chat origin=Api status=in_progress
…
# (no further output from genie-core until kill -CONT is sent to HA, OR until the
#  client closes the TCP connection from the other side, whichever happens first.
#  The task simply never times out. No error log, no telemetry, no diagnostic.)
```

```text
# Standalone reproducer: spawn a TcpListener that accepts then sleeps forever,
# point HaClient at it, observe the indefinite hang. The fix's `mod tests`
# implements exactly this pattern in `hung_server_after_connect_times_out_cleanly`.
let listener = TcpListener::bind("127.0.0.1:0").await?;
let addr = listener.local_addr()?;
tokio::spawn(async move {
    let (_conn, _) = listener.accept().await.unwrap();
    // Never write a response. Sleep forever — mimics HA in a GC pause.
    tokio::time::sleep(Duration::from_secs(600)).await;
});
let client = HaClient::new(&addr.ip().to_string(), addr.port(), "token");
// On main: this call NEVER returns. With the fix: returns Err("… timed out after 30s") at ~30s.
let _ = client.test_connection().await;
```

### Additional context

**Symptom users see:**
1. *"The assistant froze."* Voice cycle or chat turn that touches HA never replies. User asks again — the second turn might land on a different worker and succeed (if HA is back), or might also hang. Eventually the user power-cycles the appliance or restarts the service.
2. *No diagnostic for the operator.* `journalctl -u genie-core` shows the request line was logged but no error and no completion. The operator has no way to know whether genie-core is alive, HA is alive, or both.
3. *Memory growth on malformed responses.* On the read-to-EOF fallback path (no `Content-Length` and no `Transfer-Encoding: chunked`), the daemon happily accumulates the entire body in memory. If HA sends partial bytes indefinitely, RSS grows without bound until OOM-killer fires. Rare but real for misbehaving HA versions or proxies.

**Why this fires in production at all** — every household with HA exercises this code path on essentially every tool call (`home_status`, `home_control`, `home_undo`, `resolve_target`, governor/health pollers). Triggers:

| Trigger | How often in real households |
|---|---|
| HA Supervisor self-update / HACS reload / integration update | weekly+ |
| HA Python GC pause during a slow custom AppDaemon / Node-RED automation | daily under load |
| HA host Wi-Fi drop, switch reboot, ethernet renegotiation | monthly+ |
| HA process alive but slow custom integration blocks its event loop | sporadic but recurrent |
| HA crashes mid-response after TCP handshake | rare but realistic |

- `crates/genie-core/src/ha/client.rs` has tests only for `parse_http_url` (lines 297-322). No tests exist for `http_request`, `read_http_response`, or `read_chunked_body` against any kind of failure mode — no mock TCP server, no timeout coverage.
- `mod tests` in this file currently has no `tokio::test`-flavored coverage at all.

**Files to touch:**
- `crates/genie-core/src/ha/client.rs`:
  - Add `connect_timeout: Duration`, `request_timeout: Duration`, `max_response_bytes: usize` fields to `HaClient`; default `5s`, `30s`, `8 MiB`.
  - Add named constants for the defaults so they appear in the public surface and are not magic literals.
  - Wrap the post-connect block of `http_request` in `tokio::time::timeout(self.request_timeout, …)`. Map `Elapsed` to an `anyhow!("Home Assistant {method} {path} timed out after {N}s")`.
  - Extend `read_http_response` and `read_chunked_body` to take `max_bytes`; cap `Content-Length`, chunked-body, and read-to-EOF reads at that ceiling; surface a clear `anyhow!("Home Assistant response exceeded N bytes")` rather than OOM'ing.
  - Add `pub(crate) fn with_test_timeouts(mut self, connect: Duration, request: Duration, max_bytes: usize) -> Self` so tests can drive the timeouts in millisecond-scale.
- `crates/genie-core/src/ha/client.rs` tests — five new regressions, all hardware-free:
  - `connect_timeout_to_unroutable_address` — `HaClient` pointed at `192.0.2.1` (RFC 5737 documentation range) with a 100 ms test connect timeout; assert `test_connection()` fails inside the budget with a "connect" error string.
  - `hung_server_after_connect_times_out_cleanly` — `TcpListener` that accepts but never replies; assert `get_states()` fails with a "timed out" error inside the test request budget (e.g. 200 ms).
  - `slow_server_within_budget_succeeds` — `TcpListener` that accepts, waits ~50 ms, then writes a valid HTTP/1.1 200 response with a tiny JSON array; assert success.
  - `read_to_eof_response_is_size_capped` — `TcpListener` that writes a response with no `Content-Length` and no chunked encoding, then streams more bytes than `max_response_bytes`; assert the client bails with a "response too large" error rather than consuming memory.
  - `chunked_body_with_oversize_aggregate_is_capped` — `TcpListener` that returns a chunked body whose cumulative payload exceeds the cap; assert the client bails before allocating it all.

**Acceptance (PR):**
- `cargo test -p genie-core --lib ha::client` — existing 3 `parse_http_url_*` tests still pass, 5 new regression tests pass.
- Building `genie-core` in `--release` and pointing it at a `kill -STOP`ped HA (using the steps under "Reproduce") returns an explicit `Err("Home Assistant GET /api/states timed out after 30s")` from the tool dispatch path within ~30 s. No daemon hang.
- Default `geniepod.toml` and `geniepod.dev.toml` see no behaviour change for healthy HA.

**Timing:** Standalone — no upstream PRs touch `ha/client.rs` (one commit ever on this file: scaffold + an unrelated rename). Lands independently of #163, #164/165, #169 (mine), or #171 (mine). No merge-conflict surface with any of those.

**Real Behavior Proof:** `cargo test -p genie-core --lib ha::client` showing the 5 new tests pass alongside the 3 existing `parse_http_url_*` tests. Plus the standalone-reproducer behaviour: pre-fix the hung-server test would hang forever; post-fix it returns `Err("…timed out…")` in ~200 ms.

**Related:** #124 / PR #125 fixed cancellation of the LLM stream producer when the client disconnects — adjacent in spirit ("don't hang on a remote that stops responding"), different module. #109 / PR #118 ("100 consecutive voice cycles with zero stalls and zero silent drops") chased voice stalls but never touched HA HTTP I/O. No upstream issue mentions HA HTTP read timeout. The shared "every household with HA hits this on every HA hiccup" framing puts it in the same severity class as #168 / PR #169 (mine) and #170 / PR #171 (mine), each of which the maintainers have accepted as a real bug.

galuis116:
## Summary

Enforce a request-lifecycle timeout and a response-size cap on `HaClient`'s
raw-TCP HTTP client. Pre-fix only `TcpStream::connect` had a timeout (5 s);
every subsequent I/O (`write_all`, status-line `read_line`, header loop,
body `read_exact` / `read_to_string` / chunked decoder) was unbounded, so a
hung Home Assistant — Python GC pause, Supervisor self-update, integration
reload, slow custom add-on, or dropped packet after handshake — blocked the
calling chat/voice/dashboard task **forever**. The read-to-EOF fallback was
also unbounded in memory.

Same "the assistant froze" symptom class as #109 / PR #118 (voice cycle
stalls) and #124 / PR #125 (LLM stream not cancelled on disconnect), in a
different subsystem.

Closes #173.

## Changes

- `crates/genie-core/src/ha/client.rs`:
  - Surface the magic-5 connect timeout as `DEFAULT_CONNECT_TIMEOUT = 5s`,
    add `DEFAULT_REQUEST_TIMEOUT = 30s`, add
    `DEFAULT_MAX_RESPONSE_BYTES = 8 MiB`. Doc-commented to explain the
    failure mode each one prevents.
  - `HaClient` carries the three values as fields; `HaClient::new` defaults
    them from the constants — no behaviour change for existing callers.
  - `#[cfg(test)] pub(crate) fn with_test_limits(...)` builder so the new
    regression tests can run in millisecond-scale.
  - `http_request`:
    - Connect path uses `self.connect_timeout`; `Elapsed` maps to
      `anyhow!("Home Assistant connect {method} {path} timed out after Ns")`.
    - Post-connect (write + read_http_response) is wrapped in
      `tokio::time::timeout(self.request_timeout, ...)`; `Elapsed` maps to
      `anyhow!("Home Assistant {method} {path} timed out after Ns")`.
  - `read_http_response` takes `max_response_bytes`:
    - Reject up front when advertised `Content-Length` exceeds the cap.
    - Read-to-EOF fallback (no Content-Length, no chunked) uses
      `take(cap + 1)` and surfaces a clear "exceeded N bytes" error rather
      than consuming unbounded RSS.
  - `read_chunked_body` takes `max_bytes` and bails before any chunk whose
    cumulative size would exceed the cap.
- 5 new `#[tokio::test]` regressions backed by ephemeral local `TcpListener`s
  (`spawn_listener` + `test_client` helpers, ~50 LOC of scaffolding shared
  across the suite):
  - `hung_server_after_connect_times_out_cleanly` — drains the request then
    sleeps 60 s without writing a response; asserts `Err("…timed out…")`
    inside the 300 ms test budget. Pre-fix this hung forever.
  - `slow_server_within_budget_succeeds` — 50 ms delay then a valid 200
    with `[]`; asserts `get_states()` returns an empty `Vec`. Locks in
    "healthy HA still works."
  - `oversize_content_length_is_rejected` — advertises `Content-Length:
    1048576` against a 16 KiB test cap; asserts the client bails before
    allocating.
  - `read_to_eof_response_is_size_capped` — streams 64 KiB with no
    `Content-Length` and no chunked encoding; asserts the client bails with
    "exceeded N bytes" instead of consuming memory.
  - `chunked_body_aggregate_size_is_capped` — 8 × 4 KiB chunks (32 KiB) vs
    16 KiB cap; asserts the chunked decoder bails partway through.

No config schema change, no public API change. The defaults are
intentionally conservative (30 s request, 8 MiB body) — well above any
realistic HA response. ASCII-only households see byte-identical behaviour
against a healthy HA.

## Real Behavior Proof

- [x] NOT verified on Jetson hardware. The fix is in pure tokio I/O
  scheduling — no audio, voice, ALSA, CUDA, LLM-backend, or systemd
  surface — exercised end-to-end by 5 `#[tokio::test]` regressions that
  spin up a real local `TcpListener` and drive each failure mode (hung
  server, slow but OK, oversize Content-Length, read-to-EOF, oversize
  chunked aggregate).

### What I ran

Environment: x86_64 Linux dev host (Ubuntu 22.04, Rust 1.95.0). No Jetson
available.

```bash
cargo fmt --all -- --check                                                # clean
cargo clippy --workspace --all-targets -- -D warnings                    # clean
cargo clippy --workspace --all-targets --no-default-features -- -D warnings  # clean
cargo test -p genie-core --lib ha::client::                               # 8 / 0
cargo test                                                                # 610 / 0 / 3
cargo test --workspace --no-default-features                             # 505 / 0
```

### What I observed

1. **Hung server is bounded.** `hung_server_after_connect_times_out_cleanly`
   accepts the TCP connection, drains the request, then `tokio::sleep(60s)`s.
   With the fix, `HaClient::test_connection()` returns
   `Err("Home Assistant GET /api/ timed out after 300ms")` in well under
   2 seconds. Pre-fix the same call hangs for the full 60 seconds.
2. **Healthy slow HA still works.** `slow_server_within_budget_succeeds`
   sleeps 50 ms before writing a minimal 200 response. The client returns
   `Ok([])` — no regression on the happy path.
3. **Oversize body rejected up front.** `oversize_content_length_is_rejected`
   advertises 1 MiB against a 16 KiB cap; client bails with "exceeds cap"
   before allocating.
4. **Read-to-EOF is bounded.** `read_to_eof_response_is_size_capped` streams
   64 KiB; client bails with "exceeded 16384 bytes" instead of consuming RSS.
5. **Chunked aggregate is bounded.** `chunked_body_aggregate_size_is_capped`
   emits 32 KiB across 8 chunks; client bails with "chunked response
   exceeded 16384 bytes" partway through.
6. **No direct-path regression.** The 3 pre-existing `parse_http_url_*`
   tests still pass. Full `cargo test` matches the `main` baseline:
   610 / 0 / 3 (default features), 505 / 0 (`--no-default-features`).

## Test plan

A reviewer can re-verify on any Rust 1.85+ host (no Jetson, no Home
Assistant, no audio needed):

- `cargo test -p genie-core --lib ha::client::` — 8 tests (3 existing
  `parse_http_url_*` + 5 new), all green in <1 s.
- Optional manual proof against a real HA:
  1. Start `genie-core` with a working HA configured.
  2. Pause HA: `sudo kill -STOP $(pgrep -f homeassistant)` (or
     `supervisorctl stop homeassistant` while a request is mid-flight).
  3. From a separate shell:
     `curl -m 120 -X POST http://127.0.0.1:3000/api/chat \
        -d '{"message":"is the kitchen light on?"}'`.
  4. Pre-fix: genie-core hangs forever (curl gives up after 120 s).
     With the fix: the request returns within ~30 s with the chat reply
     containing the tool dispatch's
     `Err("Home Assistant GET /api/states/... timed out after 30s")`.

## Notes for reviewers

- **No merge-conflict surface with my open PRs.** The file
  `crates/genie-core/src/ha/client.rs` is not touched by PR #169
  (conversation.rs) or PR #171 (calc.rs).
- **Why 30 s / 8 MiB defaults.** 30 s is comfortably above the slowest
  realistic HA template render (~3-8 s on Orin Nano with a complex
  template). 8 MiB comfortably fits the largest realistic `/api/states`
  dump (a household with hundreds of entities still serializes well under
  1 MiB). Both are tunable later via the `HaClient` struct fields if
  operators want.
- **Why `with_test_limits` is `#[cfg(test)] pub(crate)` rather than `pub`.**
  Production callers shouldn't reach in and tighten timeouts — that's a
  config-layer concern that belongs in `genie-common/src/config.rs` if/when
  needed. Keeping the test helper crate-private avoids putting an unstable
  shape into the public surface.
- **Why a local helper instead of a workspace-wide `util::http`.** The same
  unbounded-read pattern exists in `crates/genie-core/src/tools/weather.rs`
  and arguably the OpenAI-compat client too, but each speaks a different
  HTTP dialect (Weather is a one-shot Open-Meteo GET; the LLM client
  streams SSE; HA needs cookies/auth/template-render specifics). Lifting
  to a shared util makes sense as a follow-up but expands review surface
  beyond the bug this PR closes. Kept out of scope.

### DIFF
--- a/crates/genie-core/src/ha/client.rs
+++ b/crates/genie-core/src/ha/client.rs
@@ -1,8 +1,27 @@
 use anyhow::{Context, Result};
 use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
+use std::time::Duration;
 use tokio::io::{AsyncBufReadExt, AsyncReadExt, AsyncWriteExt, BufReader};
 use tokio::net::TcpStream;
+
+/// TCP-connect cap. Same value the file had hard-coded since scaffolding;
+/// surfaced as a named constant so tests can assert against it.
+const DEFAULT_CONNECT_TIMEOUT: Duration = Duration::from_secs(5);
+
+/// Full-request cap covering write + status-line + headers + body. Without
+/// this, a Home Assistant that pauses mid-response (Python GC, integration
+/// reload, Supervisor self-update, stalled add-on) leaves the dispatching
+/// chat / voice / dashboard task hung forever — there was no timeout at all
+/// on the read side before this constant existed.
+const DEFAULT_REQUEST_TIMEOUT: Duration = Duration::from_secs(30);
+
+/// Hard upper bound on a single response body. Prevents the read-to-EOF
+/// fallback (no `Content-Length`, no `Transfer-Encoding: chunked`) from
+/// accumulating unbounded bytes in RSS, and caps the cumulative payload of
+/// a malformed chunked stream. 8 MiB comfortably fits the largest realistic
+/// `/api/states` dump for a household.
+const DEFAULT_MAX_RESPONSE_BYTES: usize = 8 * 1024 * 1024;
 
 /// Home Assistant REST API client.
 ///
@@ -15,6 +34,9 @@ pub struct HaClient {
     host: String,
     port: u16,
     token: String,
+    connect_timeout: Duration,
+    request_timeout: Duration,
+    max_response_bytes: usize,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
@@ -47,6 +69,9 @@ impl HaClient {
             host: host.to_string(),
             port,
             token: token.to_string(),
+            connect_timeout: DEFAULT_CONNECT_TIMEOUT,
+            request_timeout: DEFAULT_REQUEST_TIMEOUT,
+            max_response_bytes: DEFAULT_MAX_RESPONSE_BYTES,
         }
     }
 
@@ -55,6 +80,23 @@ impl HaClient {
     pub fn from_url(url: &str, token: &str) -> Result<Self> {
         let (host, port) = parse_http_url(url)?;
         Ok(Self::new(&host, port, token))
+    }
+
+    /// Override the default timeouts and the max response size. Intended for
+    /// `mod tests`, which need to assert that a hung server is surfaced as
+    /// an `Err` in millisecond-scale rather than waiting for the production
+    /// 30 s deadline. Production callers should leave the defaults alone.
+    #[cfg(test)]
+    pub(crate) fn with_test_limits(
+        mut self,
+        connect_timeout: Duration,
+        request_timeout: Duration,
+        max_response_bytes: usize,
+    ) -> Self {
+        self.connect_timeout = connect_timeout;
+        self.request_timeout = request_timeout;
+        self.max_response_bytes = max_response_bytes;
+        self
     }
 
     pub fn host(&self) -> &str {
@@ -136,9 +178,25 @@ impl HaClient {
         body: Option<&str>,
     ) -> Result<HttpResponse> {
         let addr = format!("{}:{}", self.host, self.port);
-        let stream =
-            tokio::time::timeout(std::time::Duration::from_secs(5), TcpStream::connect(&addr))
-                .await??;
+        let connect_timeout = self.connect_timeout;
+        let request_timeout = self.request_timeout;
+        let max_response_bytes = self.max_response_bytes;
+
+        let stream = match tokio::time::timeout(connect_timeout, TcpStream::connect(&addr)).await {
+            Ok(Ok(stream)) => stream,
+            Ok(Err(e)) => {
+                return Err(e)
+                    .with_context(|| format!("Home Assistant connect {} {} failed", method, path));
+            }
+            Err(_) => {
+                anyhow::bail!(
+                    "Home Assistant connect {} {} timed out after {:?}",
+                    method,
+                    path,
+                    connect_timeout
+                );
+            }
+        };
 
         let (reader, mut writer) = stream.into_split();
 
@@ -163,8 +221,25 @@ impl HaClient {
             )
         };
 
-        writer.write_all(request.as_bytes()).await?;
-        let response = read_http_response(reader).await?;
+        // Enforce a single deadline that covers write + status-line + headers
+        // + body. Pre-fix only the `connect` step had a timeout; a hung HA
+        // (Python GC pause, integration reload, Supervisor self-update,
+        // stalled add-on) blocked the calling task indefinitely.
+        let response = match tokio::time::timeout(request_timeout, async {
+            writer.write_all(request.as_bytes()).await?;
+            read_http_response(reader, max_response_bytes).await
+        })
+        .await
+        {
+            Ok(Ok(response)) => response,
+            Ok(Err(e)) => return Err(e),
+            Err(_) => anyhow::bail!(
+                "Home Assistant {} {} timed out after {:?}",
+                method,
+                path,
+                request_timeout
+            ),
+        };
 
         if !(200..300).contains(&response.status_code) {
             let body = response.body.trim().replace('\n', " ");
@@ -185,7 +260,10 @@ impl HaClient {
     }
 }
 
-async fn read_http_response(reader: tokio::net::tcp::OwnedReadHalf) -> Result<HttpResponse> {
+async fn read_http_response(
+    reader: tokio::net::tcp::OwnedReadHalf,
+    max_response_bytes: usize,
+) -> Result<HttpResponse> {
     let mut buf_reader = BufReader::new(reader);
 
     let mut status_line = String::new();
@@ -218,21 +296,42 @@ async fn read_http_response(reader: tokio::net::tcp::OwnedReadHalf) -> Result<Ht
     }
 
     let body = if chunked {
-        read_chunked_body(&mut buf_reader).await?
+        read_chunked_body(&mut buf_reader, max_response_bytes).await?
     } else if let Some(content_length) = content_length {
+        if content_length > max_response_bytes {
+            anyhow::bail!(
+                "Home Assistant response Content-Length {} exceeds cap {}",
+                content_length,
+                max_response_bytes
+            );
+        }
         let mut buf = vec![0u8; content_length];
         buf_reader.read_exact(&mut buf).await?;
         String::from_utf8_lossy(&buf).to_string()
     } else {
+        // No `Content-Length` and no `Transfer-Encoding: chunked`: cap the
+        // read at `max_response_bytes + 1` so we can distinguish "fits in
+        // the cap" from "ran over the cap" without ever allocating more.
+        let cap = max_response_bytes;
+        let mut limited = (&mut buf_reader).take(cap as u64 + 1);
         let mut body = String::new();
-        buf_reader.read_to_string(&mut body).await?;
+        limited.read_to_string(&mut body).await?;
+        if body.len() > cap {
+            anyhow::bail!(
+                "Home Assistant response without Content-Length exceeded {} bytes",
+                cap
+            );
+        }
         body
     };
 
     Ok(HttpResponse { status_code, body })
 }
 
-async fn read_chunked_body<R: AsyncBufReadExt + Unpin>(reader: &mut R) -> Result<String> {
+async fn read_chunked_body<R: AsyncBufReadExt + Unpin>(
+    reader: &mut R,
+    max_bytes: usize,
+) -> Result<String> {
     let mut body = Vec::new();
 
     loop {
@@ -246,6 +345,13 @@ async fn read_chunked_body<R: AsyncBufReadExt + Unpin>(reader: &mut R) -> Result
             let mut trailing = String::new();
             reader.read_line(&mut trailing).await?;
             break;
+        }
+
+        if body.len().saturating_add(size) > max_bytes {
+            anyhow::bail!(
+                "Home Assistant chunked response exceeded {} bytes",
+                max_bytes
+            );
         }
 
         let mut chunk = vec![0u8; size];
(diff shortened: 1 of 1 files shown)


ai-hpc:
Reviewed and merged at eedeb8b7d1dcfbfd870dc3809170a7cca4064f8d (eedeb8b).

Summary: adds HA client request deadlines and response-size caps so stalled or oversized Home Assistant responses cannot hang tool dispatch.

Thanks @galuis116.

PR #176, opened 2026-05-25
fix(skills): verify Ed25519 signature of .so bytes before dlopen

### LINKED ISSUE #175: [bug] skills: require_signature is a presence-only string check — any non-empty "signature" string loads an unsigned/tampered .so as arbitrary native code, and the security posture reports "signed enforcement on"
### Summary

`SkillPolicyConfig.require_signature` is advertised — in config, in the operator-facing security-posture report, and by its own name — as the gate that refuses to load skills that aren't signed. But the skill loader decides "signed" with `!manifest.signature.trim().is_empty()` ([loader.rs:133](crates/genie-core/src/skills/loader.rs#L133)). No bytes are verified against any key. A skill `.so` is loaded via `dlopen` and runs as **arbitrary native code inside `genie-core`** ([loader.rs:370-447](crates/genie-core/src/skills/loader.rs#L370-L447)), and the runtime sandbox that was supposed to contain it (Landlock) is a documented no-op stub ([sandbox.rs:79-99](crates/genie-core/src/security/sandbox.rs#L79-L99)). So `require_signature = true` — the one authenticity control on native-code loading — is satisfied by `"signature": "x"`.

### Steps to reproduce

1. Enable the control an operator would reasonably trust, in `geniepod.toml`:
   ```toml
   [core.skill_policy]
   require_manifest  = true
   require_signature = true
   ```
2. Confirm the stack now *reports* signature enforcement as on. `Config::security_posture()` drops the `skill_signature_not_required` risk flag ([config.rs:921-922](crates/genie-common/src/config.rs#L921-L922)) and reports `"skill_signature_required": true` ([config.rs:955](crates/genie-common/src/config.rs#L955)).
3. Drop a skill `.so` you did **not** sign into the skills directory with a sidecar manifest whose `name`/`version` match the vtable and whose signature is any junk:
   ```json
   { "name": "evil", "version": "0.1.0", "description": "x",
     "permissions": [], "reviewed_by": "", "signature": "x" }
   ```
4. Start `genie-core` (live wiring: [main.rs:57](crates/genie-core/src/main.rs#L57), voice path [voice/mod.rs:48](crates/genie-core/src/voice/mod.rs#L48)).
5. `enforce_skill_policy` passes the signature gate ([loader.rs:293-294](crates/genie-core/src/skills/loader.rs#L293-L294)) because `signed == true`, and the unsigned `.so` is `dlopen`'d and registered as a callable tool.

### Expected behavior

With `require_signature = true`, only skills whose binary is cryptographically verified against a trusted public key should load. `"signature": "x"` (or any unverifiable string) must be rejected. The security-posture report must not claim signature enforcement is active unless real verification is performed.

### Actual behavior

"Signed" means nothing more than "the manifest's `signature` field is a non-empty string":

```rust
// crates/genie-core/src/skills/loader.rs:133
let signed = !manifest.signature.trim().is_empty();
```

```rust
// crates/genie-core/src/skills/loader.rs:293-294  (enforce_skill_policy)
if policy.require_signature && !manifest.signed {
    anyhow::bail!("skill signature required but manifest is unsigned");
}
```

There is no signature scheme, no key, and no verification anywhere in the loader — the `signature` field is never checked against the `.so` bytes. The manifest is sidecar JSON the attacker supplies *alongside* the `.so` ([loader.rs:247-282](crates/genie-core/src/skills/loader.rs#L247-L282)), so they trivially set `signature` to any non-empty value.

Why this is critical rather than cosmetic:

- A skill is loaded with `Library::new(path)` and its `genie_skill_init` symbol is called — **arbitrary native code in the `genie-core` address space**, with the process's full privileges. The loader itself admits this: *"loading a .so is inherently unsafe. We trust skills from the skills directory"* ([loader.rs:371-372](crates/genie-core/src/skills/loader.rs#L371-L372)).
- The only runtime containment, Landlock, is **not enforced** — `apply_landlock_linux` logs the rules it *would* apply and returns `Ok(())` without restricting anything ([sandbox.rs:79-99](crates/genie-core/src/security/sandbox.rs#L79-L99)). So there is no second line of defense behind the signature gate.
- The `require_signature` knob is presented to operators as a security posture they can turn *on* ([config.rs:231-233](crates/genie-common/src/config.rs#L231-L233), reported at [config.rs:955](crates/genie-common/src/config.rs#L955), flagged as a risk when *off* at [config.rs:921-922](crates/genie-common/src/config.rs#L921-L922)). Turning it on changes the reported posture from "at risk" to "enforced" while changing the actual security boundary by **zero**. A no-op control that reports as enforced is worse than an absent one — it manufactures false assurance for the exact operator who cared enough to enable it.

(The internal doc comment at [loader.rs:18](crates/genie-core/src/skills/loader.rs#L18) and [loader.rs:34](crates/genie-core/src/skills/loader.rs#L34) concedes the field is "presence only … not a signature check" — but that candor lives in source comments, while the *enforced policy*, its *name*, and the *security-posture report* all say "signature required." The contract the operator sees and the behavior they get disagree.)

### Hardware

### JetPack / L4T version

N/A (hardware-independent). Affects all targets equally.

v1.0.0-alpha.9 / commit `8e98411`

### Relevant logs

```shell
# With require_signature = true and an unsigned .so + manifest {"signature":"x"}:
journalctl -u genie-core -n 50
# ... "skill loaded" skill=evil   <-- loaded; no signature error, because "x" is non-empty

# The posture report claims enforcement is on:
genie-ctl diag    # or the security_posture() JSON
# ... "policy": { "skill_signature_required": true }, "risk_flags": [ ... ]  (no "skill_signature_not_required")
```

### Additional context

**Not a duplicate of #23.** Open #23 ("beta: skill sandbox policy — drop root, landlock/bubblewrap, subprocess + network allowlists") is about *containing* a skill's behavior at runtime. This issue is about *authenticity/integrity at load time*: the gate that decides whether an `.so` is trusted enough to `dlopen` at all is satisfied by a non-empty string. They are complementary, not the same — and today **both** are ineffective (the sandbox is a stub, the signature gate is a string check), so an unsigned, untrusted `.so` runs uncontained native code.

**Proposed structural fix (make `signed` mean cryptographically verified, with a real key):**

The bug is structural: the loader treats a *self-asserted manifest field* as proof of review. Authenticity must be derived from a signature over the binary that only a trusted key holder could produce — never from the presence of a string the same party supplied.

1. Define a signing scheme. Detached Ed25519 over the `.so` bytes is sufficient and dependency-light. Ship one or more trusted public keys with the distribution (e.g. `/etc/geniepod/skill-keys/*.pub`), not in the skills directory.
2. Make the manifest carry the *signature bytes* (base64) and a key id, and verify in the loader **before** `Library::new`:
   ```rust
   // Only "signed" if a trusted key verifies a signature over the actual .so bytes.
   let signed = match (&manifest.signature, &manifest.key_id) {
       (sig, Some(key_id)) if !sig.trim().is_empty() => {
           let bytes = std::fs::read(skill_path)?;
           verify_detached(&trusted_keys, key_id, &bytes, sig)  // Ed25519, returns bool
       }
       _ => false,
   };
   ```
   Verification must run on the bytes that will actually be executed, and must fail closed if the key is unknown or the signature does not validate.
3. `Mode`-style single source of truth: `manifest.signed` is the *only* place authenticity is decided, and it is decided by verification, not by `is_empty()`. Remove the `!signature.trim().is_empty()` heuristic entirely.
4. Make the report honest: `security_posture()` should distinguish "signature material present" from "signature verified," and must not emit `skill_signature_required: true` unless real verification is wired.
5. Regression tests asserting the **negative** cases that currently pass:
   - `require_signature = true` + manifest `"signature": "x"` (no key / junk) → **load rejected**.
   - valid signature by a trusted key over the exact `.so` → load accepted.
   - one byte of the `.so` flipped after signing → load rejected (tamper detection).
   - unknown key id → load rejected (fail closed).
   A test that today asserts `require_signature` blocks `"signature": "x"` would have caught this; none exists ([loader.rs tests](crates/genie-core/src/skills/loader.rs#L501-L701) only check `require_manifest` and `denied_permissions`).

**Severity: critical.** The control that gates loading arbitrary native code into `genie-core` is bypassable with a one-character string, while reporting itself as enforced. Combined with the no-op Landlock sandbox, an unsigned/tampered `.so` placed in the skills directory executes uncontained — and the operator who enabled `require_signature` has been told they are protected.

enjoyandlove:
## Summary

`[core.skill_policy].require_signature` was presence-only: any non-empty
`signature` string in a sidecar manifest counted as "signed", so an attacker
who could drop a `.so` into the skills directory could also drop a manifest
with `"signature": "x"` and have arbitrary native code `dlopen`'d into the
genie-core address space. This PR makes `require_signature` a real authenticity
gate — a detached **Ed25519** signature over the exact `.so` bytes, verified
against a trusted key **before** `dlopen`. Fixes #175.

## Changes

- Add `skills/signature.rs`: `TrustedKeys` loads `<key_id>.pub` (base64 32-byte
  Ed25519) keys from a trusted-key directory and verifies detached signatures
  with `verify_strict`. Fails closed on unknown key id, malformed/invalid
  signature, or empty key set.
- Verify the signature over the `.so` file bytes in `SkillLoader::load_skill`
  **before** `dlopen`; reject the load when `require_signature` is on and the
  bytes are not cryptographically verified. `SkillManifestAudit.signed` is now
  set only by verification, never by string presence.
- Add `key_id` to the sidecar manifest and audit view; signature material is
  documented as cryptographic, not audit-only.
- Add `[core.skill_policy].signature_key_dir` (default `/etc/geniepod/skill-keys`),
  kept outside the attacker-writable skills directory so dropping a `.so`
  cannot also drop a trusting key.
- `household_security_summary()` now reports the signature scheme, key dir, and
  whether trusted keys are present, and raises
  `skill_signature_required_but_no_trusted_keys` when `require_signature` is on
  but no keys are installed (loader fails closed → every skill rejected).
- `genie-ctl skill list` prints `signed: verified (key: <id>)` vs `no` instead
  of a bare boolean.
- Add `ed25519-dalek` (verify-only, no RNG/keygen features) and `base64` deps.
- Update sample-config comments and `SKILL-DEVELOPER-GUIDE.md` with how to
  generate a key pair, sign a `.so`, and populate `key_id` / `signature`.

## Real Behavior Proof

This is a pure logic/crypto change in `genie-core` — it does not touch voice,
audio, Home Assistant, or the dashboard, so the verification path is the
crate's unit tests, which exercise the real `ed25519-dalek` verification over
real `.so` bytes (no mocks).

### What I ran

```
# host: <fill in — dev workstation, no Jetson required for this crypto path>
cargo test -p genie-core skills::
cargo test -p genie-common config::
cargo test -p genie-ctl
```

### What I observed

New tests added and expected to pass:

- `skills::signature::tests`
  - `empty_keys_reject_everything`
  - `valid_signature_verifies`
  - `junk_signature_is_rejected` — reproduces the bug: `"signature": "x"` no longer verifies
  - `tampered_message_is_rejected`
  - `unknown_key_id_is_rejected`
  - `wrong_trusted_key_is_rejected`
  - `malformed_key_file_is_skipped`
- `skills::loader::tests`
  - `loader_rejects_required_signature_with_junk_string` — the reported issue: junk string + `require_signature` no longer loads
  - `loader_accepts_valid_signature`
  - `loader_rejects_tampered_so` — one flipped byte after signing → load rejected
  - `loader_rejects_unknown_key_id`
  - `loader_fails_closed_without_trusted_keys`
- `config::` — `skill_policy_signature_key_dir_overridable`,
  `security_posture_flags_require_signature_without_keys`
- `genie-ctl` — install test updated: a bare `"signature"` string asserts
  `!signed`.

## Test plan

1. `cargo test -p genie-core skills::` — all signature + loader cases pass.
2. Negative check by hand: create a `.so` with a sidecar manifest containing
   `"signature": "x"`, enable `[core.skill_policy].require_signature = true`,
   and confirm the loader refuses to load it ("skill signature required but
   not cryptographically verified").
3. Positive check: generate an Ed25519 key pair, drop the public half in
   `signature_key_dir` as `<key_id>.pub`, sign the `.so` bytes, put the base64
   signature + `key_id` in the manifest, and confirm the skill loads and
   `genie-ctl skill list` reports `signed: verified (key: <id>)`.
   (Signing recipe is in `skills/SKILL-DEVELOPER-GUIDE.md`.)
4. Fail-closed check: with `require_signature = true` and an empty key dir,
   confirm no skill loads and `household_security_summary()` raises
   `skill_signature_required_but_no_trusted_keys`.

## Notes for reviewers

- Default behavior is unchanged: `require_signature` defaults to `false`, so
  existing deployments keep loading local skills until an operator opts in.
- Verification runs on the bytes read from disk immediately before `dlopen`;
  there is an unavoidable TOCTOU window between read and load that matches the
  existing model (the skills directory is trusted filesystem). The key
  directory is intentionally separate from the skills directory.
- `ed25519-dalek` is pulled in with `default-features = false` (verify only,
  no RNG/keygen) to keep the runtime footprint minimal.
- Follow-up candidates (not in this PR): a `genie-ctl skill sign` helper, and
  key rotation / revocation handling.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -53,10 +53,25 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
 
 [[package]]
+name = "base64ct"
+version = "1.8.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2af50177e190e07a26ab74f8b1efbfe2ef87da2116221318cb1c2e82baf7de06"
+
+[[package]]
 name = "bitflags"
 version = "2.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "843867be96c8daad0d758b57df9392b6d8d271134fce549de6ce169ff98a92af"
+
+[[package]]
+name = "block-buffer"
+version = "0.10.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3078c7629b62d3f0439517fa394996acacc5cbc91c5a20d8c658e77abd503a71"
+dependencies = [
+ "generic-array",
+]
 
 [[package]]
 name = "bumpalo"
@@ -93,6 +108,78 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "613afe47fcd5fac7ccf1db93babcb082c5994d996f20b8b159f2ad1658eb5724"
 
 [[package]]
+name = "const-oid"
+version = "0.9.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c2459377285ad874054d797f3ccebf984978aa39129f6eafde5cdc8315b612f8"
+
+[[package]]
+name = "cpufeatures"
+version = "0.2.17"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "59ed5838eebb26a2bb2e58f6d5b5316989ae9d08bab10e0e6d103e656d1b0280"
+dependencies = [
+ "libc",
+]
+
+[[package]]
+name = "crypto-common"
+version = "0.1.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "78c8292055d1c1df0cce5d180393dc8cce0abec0a7102adb6c7b1eef6016d60a"
+dependencies = [
+ "generic-array",
+ "typenum",
+]
+
+[[package]]
+name = "curve25519-dalek"
+version = "4.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "97fb8b7c4503de7d6ae7b42ab72a5a59857b4c937ec27a3d4539dba95b5ab2be"
+dependencies = [
+ "cfg-if",
+ "cpufeatures",
+ "curve25519-dalek-derive",
+ "digest",
+ "fiat-crypto",
+ "rustc_version",
+ "subtle",
+ "zeroize",
+]
+
+[[package]]
+name = "curve25519-dalek-derive"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f46882e17999c6cc590af592290432be3bce0428cb0d5f8b6715e4dc7b383eb3"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
+name = "der"
+version = "0.7.10"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e7c1832837b905bbfb5101e07cc24c8deddf52f93225eee6ead5f4d63d53ddcb"
+dependencies = [
+ "const-oid",
+ "zeroize",
+]
+
+[[package]]
+name = "digest"
+version = "0.10.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9ed9a281f7bc9b7576e61468ba615a66a5c8cfdff42420a70aa82701a3b1e292"
+dependencies = [
+ "block-buffer",
+ "crypto-common",
+]
+
+[[package]]
 name = "displaydoc"
 version = "0.2.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -101,6 +188,30 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn",
+]
+
+[[package]]
+name = "ed25519"
+version = "2.2.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "115531babc129696a58c64a4fef0a8bf9e9698629fb97e9e40767d235cfbcd53"
+dependencies = [
+ "pkcs8",
+ "signature",
+]
+
+[[package]]
+name = "ed25519-dalek"
+version = "2.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "70e796c081cee67dc755e1a36a0a172b897fab85fc3f6bc48307991f64e4eca9"
+dependencies = [
+ "curve25519-dalek",
+ "ed25519",
+ "serde",
+ "sha2",
+ "subtle",
+ "zeroize",
 ]
 
 [[package]]
@@ -130,6 +241,12 @@ name = "fallible-streaming-iterator"
 version = "0.1.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7360491ce676a36bf9bb3c56c1aa791658183a54d2744120f27285738d90465a"
+
+[[package]]
+name = "fiat-crypto"
+version = "0.2.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "28dea519a9695b9977216879a3ebfddf92f1c08c05d984f8996aecd6ecdc811d"
 
 [[package]]
 name = "find-msvc-tools"
@@ -180,6 +297,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "generic-array"
+version = "0.14.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "85649ca51fd72272d7821adaf274ad91c288277713d9c18820d8499a7ff69e9a"
+dependencies = [
+ "typenum",
+ "version_check",
+]
+
+[[package]]
 name = "genie-api"
 version = "1.0.0-alpha.9"
 dependencies = [
@@ -210,6 +337,8 @@ version = "1.0.0-alpha.9"
 dependencies = [
  "anyhow",
  "async-trait",
+ "base64",
+ "ed25519-dalek",
  "genie-common",
  "genie-skill-sdk",
  "libc",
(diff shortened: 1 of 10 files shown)


ai-hpc (changes requested):
Reviewed and holding this PR before merge because it changes native skill loading trust behavior.

Summary: Ed25519 verification before dlopen is the right direction, but please add/confirm coverage for operational edge cases before landing: malformed trusted key files should not make the system report a trusted posture, unsigned native skills must be rejected before dlopen, and signed skills should have a clear success-path test or documented manual verification. This needs that security validation before merge.

Thanks @enjoyandlove.

enjoyandlove:
Hi,@ai-hpc
Could you please review this PR?

ai-hpc (changes requested):
Still holding this PR for changes.

Summary: verifying native skill signatures before dlopen is the right direction, but the failure-mode coverage still needs to prove the trust posture cannot be overstated.

Question: can you add a test showing an unsigned .so is rejected before any dlopen side effect, and another showing malformed trusted-key config reports an untrusted/error posture?

Thanks @enjoyandlove.

ai-hpc (approved):
Reviewed. verified native skill Ed25519 signatures before dlopen with fail-closed coverage. Existing change-request blockers are addressed; approving for merge.

ai-hpc:
reviewd, verified native skill Ed25519 signatures before dlopen with fail-closed coverage and merged at c76b456c6430
Thanks @enjoyandlove

PR #178, opened 2026-05-25
fix(core/security): redact every secret occurrence in sanitize_output

### LINKED ISSUE #177: [bug] security: sanitize_output inspects only the FIRST occurrence of each secret prefix — a short decoy or a second key defeats redaction and leaks live secrets to the user
### Summary

`sanitize_output` is the last-line redaction applied to *every* model reply before it reaches a user (web chat, SSE stream, OpenAI bridge, tool summaries, voice, REPL). For each secret pattern it calls `find_secret_pattern`, which does a single `text.find(prefix)` — **only the first occurrence** — and then either redacts that one token or gives up. Two consequences: (1) if the first occurrence of a prefix is shorter than `min_len`, the whole pattern aborts and a valid long key *later in the same text* is emitted verbatim; (2) a second, distinct secret sharing the same prefix is never found. On a shared-room household appliance whose stated posture is "shared-room-safe prompt filtering," this defeats the control that is supposed to stop secret leakage.

### Steps to reproduce

1. Send a prompt that makes the model emit text containing a short token with a secret prefix *before* the real secret. (Any tool output, retrieved memory, or injected content can do this; an adversarial/injected prompt can do it deliberately.) Example reply text the model produces:
   ```
   For example a key looks like sk- (short). The real one is sk-proj-1234567890abcdefghij.
   ```
2. Observe the user-facing output: the real `sk-proj-…` key is **not** redacted.
3. Second leak — two distinct secrets, same prefix:
   ```
   Prod token ghp_AAAAAAAAAAAAAAAAAAAA and staging token ghp_BBBBBBBBBBBBBBBBBBBB
   ```
   Only the first `ghp_…` is replaced; the second is emitted verbatim.
4. Both reproduce in a unit test against `sanitize_output` directly, and end-to-end through any chat path ([server.rs:682](crates/genie-core/src/server.rs#L682), [voice_loop.rs:721](crates/genie-core/src/voice_loop.rs#L721), etc.).

### Expected behavior

`sanitize_output` should redact **every** substring matching any secret pattern, regardless of position, regardless of how many times the prefix appears, and regardless of a shorter non-matching occurrence appearing first. No valid-length secret should ever survive the pass.

### Actual behavior

The scan is first-occurrence-only and single-token:

```rust
// crates/genie-core/src/security/sandbox.rs:124-140
pub fn sanitize_output(text: &str) -> String {
    let mut result = text.to_string();
    for pattern in SECRET_PATTERNS {
        if let Some(re_match) = find_secret_pattern(&result, pattern) {
            let redacted = format!("[REDACTED:{}]", pattern.name);
            result = result.replace(&re_match, &redacted);   // replaces only this exact token
            ...
        }
    }
    result
}

// crates/genie-core/src/security/sandbox.rs:196-210
fn find_secret_pattern(text: &str, pattern: &SecretPattern) -> Option<String> {
    if let Some(pos) = text.find(pattern.prefix) {      // FIRST occurrence only
        let rest = &text[pos..];
        let end = rest.find(|c: char| c.is_whitespace() || c=='"' || c=='\'' || c==',' || c=='}')
            .unwrap_or(rest.len());
        if end >= pattern.min_len {
            return Some(rest[..end].to_string());
        }
    }
    None   // <- first occurrence too short => give up, even if a long key follows
}
```

Two distinct leak paths:

- **Decoy abort (most severe):** `find` locks onto the *first* `prefix`. If that occurrence is shorter than `min_len`, the `if end >= min_len` fails and the function returns `None` for the whole pattern — a valid long secret appearing later is never examined. A single earlier `sk-`/`ghp_`/`eyJ` fragment (innocent or injected) shields every real key of that family in the rest of the message.
- **Second-secret leak:** even when the first match is redacted, `result.replace(&re_match, …)` only removes occurrences *equal to that exact token*. The loop runs `find_secret_pattern` **once per pattern**, so a different secret sharing the prefix (e.g. a second distinct `ghp_…`) is never located and leaks verbatim.

This is the same structural defect already noted for the governor's pair-enumeration bug, in a different control: the redaction is built on the wrong primitive. `String::find` returns one position and `String::replace(token)` removes one literal — neither is an "all matches" scan, which is exactly what a secret filter needs.

The existing tests pass only because each fixture contains exactly one secret with no earlier short decoy ([sandbox.rs:242-269](crates/genie-core/src/security/sandbox.rs#L242-L269)), so the gap was never exercised.

### Hardware

### JetPack / L4T version

N/A (hardware-independent).

v1.0.0-alpha.9 / commit `8e98411`

### Relevant logs

```shell
# Only the first detected token logs a warning; later/second secrets log nothing and pass through:
journalctl -u genie-core | grep "secret pattern detected"
# ... one "pattern=github_token" line for the first ghp_ token; the second ghp_ token is silent and leaks
```

### Additional context

This is the chokepoint every outward reply funnels through — confirmed call sites: web/non-stream and stream chat ([server.rs:682](crates/genie-core/src/server.rs#L682), [server.rs:799](crates/genie-core/src/server.rs#L799)), tool-action and tool-summary replies ([server.rs:837](crates/genie-core/src/server.rs#L837), [server.rs:880](crates/genie-core/src/server.rs#L880)), OpenAI-compat bridge ([server.rs:1667](crates/genie-core/src/server.rs#L1667), [server.rs:1762](crates/genie-core/src/server.rs#L1762)), voice replies ([voice_loop.rs:721](crates/genie-core/src/voice_loop.rs#L721), [voice_loop.rs:756](crates/genie-core/src/voice_loop.rs#L756)), and the REPL ([repl.rs:83](crates/genie-core/src/repl.rs#L83)). The product posture explicitly advertises `shared_room_safe_prompt_filtering: true` ([config.rs:932](crates/genie-common/src/config.rs#L932)); this is the filter that claim rests on.

**Proposed structural fix (scan all occurrences, redact each independently):**

Replace the first-occurrence `find` + single-token `replace` with a left-to-right scan that finds *every* qualifying match for *every* pattern and redacts each in place:

```rust
pub fn sanitize_output(text: &str) -> String {
    let mut result = String::with_capacity(text.len());
    let mut rest = text;
    'outer: while !rest.is_empty() {
        // Find the earliest qualifying secret match across ALL patterns at this position.
        for (i, _) in rest.char_indices() {
            for pattern in SECRET_PATTERNS {
                if rest[i..].starts_with(pattern.prefix) {
                    let tail = &rest[i..];
                    let end = tail
                        .find(|c: char| c.is_whitespace() || matches!(c, '"' | '\'' | ',' | '}'))
                        .unwrap_or(tail.len());
                    if end >= pattern.min_len {
                        result.push_str(&rest[..i]);
                        result.push_str(&format!("[REDACTED:{}]", pattern.name));
                        tracing::warn!(pattern = pattern.name, "secret pattern redacted");
                        rest = &tail[end..];
                        continue 'outer;
                    }
                }
            }
        }
        result.push_str(rest); // no more matches
        break;
    }
    result
}
```

Key properties to preserve / add:
- A short non-matching occurrence of a prefix must **not** abort the scan — it advances past it and keeps looking (kills the decoy-abort leak).
- Every qualifying match is redacted, including multiple distinct secrets of the same family (kills the second-secret leak).
- Keep the existing `min_len`/delimiter heuristics to avoid false positives on normal prose ([sandbox.rs:272-276](crates/genie-core/src/security/sandbox.rs#L272-L276) should still pass).

Regression tests to add (each fails today):
- `"sk- and the real sk-proj-1234567890abcdefghij"` → real key redacted.
- `"ghp_AAAAAAAAAAAAAAAAAAAA and ghp_BBBBBBBBBBBBBBBBBBBB"` → **both** redacted.
- mixed families in one reply (`sk-…` + `eyJ…` + `AKIA…`) → all redacted.
- normal prose with an incidental short `sk-`/`pk-` fragment → unchanged (no false positive, no abort).

**Severity: critical.** This is the only secret-redaction boundary between the local LLM (which can be steered by retrieved memory, tool output, or injected web/Telegram content) and a shared-room household audience. A single earlier short token — or simply two keys in one answer — silently defeats it, leaking live credentials to whoever is in the room or on the chat.

enjoyandlove:
## Summary

`sanitize_output` only redacted the **first** occurrence of each secret prefix, so any later secret sharing that prefix — and any real secret hidden behind a short decoy token — leaked through unredacted. This PR makes the scan walk the entire string and redact every match. Fixes #177.

## Changes

- Replaced `find_secret_pattern` (returns the first match only) with `find_secret_matches`, which scans the whole string and returns every token matching a pattern.
- `sanitize_output` now iterates over all matches per pattern instead of redacting only the first hit.
- Guaranteed forward progress in the scan loop by advancing past each token end, so a prefix with no delimiters can't cause an infinite loop or re-match.
- Added regression tests:
  - `sanitize_redacts_second_secret_with_same_prefix` — two distinct `ghp_` tokens are both redacted.
  - `sanitize_redacts_secret_after_short_decoy` — a `sk-` decoy ahead of a real `sk-proj-...` key no longer aborts the scan.

## Real Behavior Proof

This is a pure-Rust string-sanitization fix in `genie-core` with no voice/audio/Home Assistant/dashboard surface, so the equivalent verification path is the unit-test suite for the affected module, run on the dev host.

### What I ran

```
# dev host (no Jetson required for this change)
cargo test -p genie-core security::sandbox
```

## Test plan

1. `git checkout main && cargo test -p genie-core security::sandbox` — confirm the two new tests are absent (or cherry-pick them to watch them fail).
2. `git checkout fix/177-sanitize-output-redact-all-secrets && cargo test -p genie-core security::sandbox` — all sandbox tests pass.
3. Spot-check: feed a string containing two `ghp_` tokens through `sanitize_output` and confirm both are replaced with `[REDACTED:github_token]`.

## Notes for reviewers

- Forward progress in the scan loop relies on `end` always being `>= prefix length` (prefixes contain no delimiter chars), so `search_start` strictly increases — no infinite-loop risk.
- Behavior for the single-secret case is unchanged; this only adds coverage for the 2nd-and-beyond occurrences.

### DIFF
--- a/crates/genie-core/src/security/sandbox.rs
+++ b/crates/genie-core/src/security/sandbox.rs
@@ -124,9 +124,12 @@ pub fn validate_inference_route(url: &str) -> Result<(), String> {
 pub fn sanitize_output(text: &str) -> String {
     let mut result = text.to_string();
 
-    // Redact common secret patterns.
+    // Redact common secret patterns. We must scan *every* occurrence of each
+    // prefix, not just the first: a short decoy token (e.g. `sk-`) appearing
+    // ahead of a real secret must not abort the scan, and two distinct secrets
+    // sharing a prefix must both be redacted. See issue #177.
     for pattern in SECRET_PATTERNS {
-        if let Some(re_match) = find_secret_pattern(&result, pattern) {
+        for re_match in find_secret_matches(&result, pattern) {
             let redacted = format!("[REDACTED:{}]", pattern.name);
             result = result.replace(&re_match, &redacted);
             tracing::warn!(
@@ -193,20 +196,34 @@ const SECRET_PATTERNS: &[SecretPattern] = &[
     },
 ];
 
-fn find_secret_pattern(text: &str, pattern: &SecretPattern) -> Option<String> {
-    if let Some(pos) = text.find(pattern.prefix) {
-        // Extract the token-like string after the prefix.
-        let start = pos;
+/// Find every secret-like token matching `pattern` in `text`.
+///
+/// Scans the whole string rather than stopping at the first prefix hit, so a
+/// short decoy token cannot mask a real secret that follows, and multiple
+/// distinct secrets sharing a prefix are all returned.
+fn find_secret_matches(text: &str, pattern: &SecretPattern) -> Vec<String> {
+    let mut matches = Vec::new();
+    let mut search_start = 0;
+
+    while let Some(rel_pos) = text[search_start..].find(pattern.prefix) {
+        let pos = search_start + rel_pos;
         let rest = &text[pos..];
+        // Extract the token-like string after the prefix.
         let end = rest
             .find(|c: char| c.is_whitespace() || c == '"' || c == '\'' || c == ',' || c == '}')
             .unwrap_or(rest.len());
 
         if end >= pattern.min_len {
-            return Some(rest[..end].to_string());
+            matches.push(rest[..end].to_string());
         }
+
+        // Advance past this token so the next iteration cannot rematch it.
+        // `end` is always >= prefix length (the prefix holds no delimiters),
+        // guaranteeing forward progress.
+        search_start = pos + end;
     }
-    None
+
+    matches
 }
 
 fn extract_host(url: &str) -> String {
@@ -266,6 +283,31 @@ mod tests {
         let text = "AWS key: AKIAIOSFODNN7EXAMPLE";
         let sanitized = sanitize_output(text);
         assert!(sanitized.contains("[REDACTED:aws_key]"));
+    }
+
+    #[test]
+    fn sanitize_redacts_second_secret_with_same_prefix() {
+        // Two distinct GitHub tokens — both must be redacted (issue #177).
+        let first = "ghp_AAAAAAAAAAAAAAAAAAAAAAAA";
+        let second = "ghp_BBBBBBBBBBBBBBBBBBBBBBBB";
+        let text = format!("first {first} and second {second} end");
+        let sanitized = sanitize_output(&text);
+        assert!(!sanitized.contains(first), "first token leaked: {sanitized}");
+        assert!(
+            !sanitized.contains(second),
+            "second token leaked: {sanitized}"
+        );
+        assert_eq!(sanitized.matches("[REDACTED:github_token]").count(), 2);
+    }
+
+    #[test]
+    fn sanitize_redacts_secret_after_short_decoy() {
+        // A short decoy sharing the prefix must not abort the scan (issue #177).
+        let real = "sk-proj-1234567890abcdefghijklmnop";
+        let text = format!("decoy sk- then real {real} here");
+        let sanitized = sanitize_output(&text);
+        assert!(!sanitized.contains(real), "real secret leaked: {sanitized}");
+        assert!(sanitized.contains("[REDACTED:api_key]"));
     }
 
     #[test]
--- a/crates/genie-core/src/security/sandbox.rs
+++ b/crates/genie-core/src/security/sandbox.rs
@@ -292,7 +292,10 @@ mod tests {
         let second = "ghp_BBBBBBBBBBBBBBBBBBBBBBBB";
         let text = format!("first {first} and second {second} end");
         let sanitized = sanitize_output(&text);
-        assert!(!sanitized.contains(first), "first token leaked: {sanitized}");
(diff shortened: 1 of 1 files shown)


ai-hpc:
Reviewed and merged at d7746eaebcbbbf077aab4ca4f09c66562ea93334 (d7746ea).

Summary: redacts every detected secret occurrence from sanitized tool output instead of only the first match.

Thanks @enjoyandlove.

PR #182, opened 2026-05-26
fix(llm): bound OpenAI-compat client timeouts to unwedge chat turn gate

### LINKED ISSUE #181: [bug] llm/server: chat path has no LLM request timeout; RetryLlmClient is dead code and a single hung backend read wedges all chat via chat_turn_lock
### Summary

The chat path applies **no timeout to the LLM request**, and the only component built to provide one — `RetryLlmClient` (60 s timeout + retries + fallback) — is **dead code, never wired in**; because the server holds the process-wide `chat_turn_lock` across the entire unbounded LLM read, a single backend that accepts a connection and then stalls wedges **all** chat forever while `/api/health` keeps reporting "connected".

### Steps to reproduce

1. Run `genie-core` normally (default `genie-ai-runtime` backend, or llama.cpp — both use `OpenAiCompatClient`).
2. Put the LLM backend into a state where it **accepts the TCP connection but never finishes the response**. Any of:
   - kill/freeze the model mid-generation, or
   - induce the governor to restart the model under memory pressure (the documented "backend restarting during governor mode switch" in `llm/retry.rs`), or
   - simulate a half-open socket to the backend.
3. Send one chat turn through any production path: web UI (`POST /api/chat/stream`), the OpenAI bridge (`POST /v1/chat/completions`), voice, REPL, or Telegram.
4. While that first turn hangs, send any further chat turn from any path.
5. Hit `GET /api/health` / open the dashboard.

### Expected behavior

- A stalled backend read is bounded by a timeout (the `RetryLlmClient` design uses 60 s) and fails gracefully, freeing the turn.
- A single wedged turn must **not** block every subsequent turn across all channels.
- `/api/health` must reflect that chat is not actually serving requests.

### Actual behavior

- **The live client has no timeout.** `main.rs:46` builds a plain `LlmClient::from_service_config(...)`. `LlmClient` (`llm/mod.rs:105`) just forwards to `Box<dyn LlmBackendClient>` with no timeout anywhere.
- **The retry/timeout layer is dead code.** `RetryLlmClient` (`llm/retry.rs`) wraps `chat`/`chat_stream` in `tokio::time::timeout` with retries, but is referenced **only** by its own `#[allow(unused_imports)]` re-export at `llm/mod.rs:17-18`. `RetryLlmClient::new` / `chat_with_retry` / `stream_or_fallback` are called nowhere outside `retry.rs`'s own tests (`retry.rs:212`, `retry.rs:220`). The `#[allow(unused_imports)]` is itself the smoking gun that the resilience layer is unused.
- **The HTTP client times out only on connect, never on read.** In `OpenAiCompatClient` (shared by both backends):
  - non-streaming `http_post`: 30 s timeout on `TcpStream::connect` (`openai_compat.rs:544-546`), but `read_line` (`:571`) / `read_exact` (`:587`) for headers+body have **none**.
  - streaming: `TcpStream::connect` has **no timeout at all** (`openai_compat.rs:463`), and the `next_line()` SSE loop (`:480`, `:491`) has none.
- **The global chat lock is held across that unbounded read.** `ChatStream` takes the lock at `server.rs:290` (`let _guard = chat_turn_lock.lock().await;`) and holds it for the whole `handle_chat_stream(...)` call — which contains the streaming LLM read (`server.rs:587`) and a `biased` select (`server.rs:641`) that only escapes on a *write* error. `Chat` and `OpenAiChat` hold the same lock via `with_chat_turn_lock` (`server.rs:222-225`, used at `server.rs:318` and the OpenAI route at `server.rs:376`). If no tokens ever arrive, even a client disconnect won't free it.
- **Result:** the first wedged turn hangs forever; every later chat/stream/OpenAI/Telegram turn blocks permanently on `chat_turn_lock.lock().await` and never returns.
- **Health stays green.** `handle_health` (`server.rs:1087`) calls `llm.health()` (`server.rs:1098` → `openai_compat.rs:538`), a separate connect probe that never touches `chat_turn_lock` — so monitoring/dashboard show "connected" while the assistant is 100 % dead.
- **No self-recovery.** Only a process restart clears it.

### Hardware

Jetson Orin Nano Super 8 GB

### JetPack / L4T version

 JetPack 6.1 / L4T R36.4

`8e5ed88` (branch `fix/177-sanitize-output-redact-all-secrets`)

### Relevant logs

```shell
# /api/health keeps reporting healthy while chat is wedged:
$ curl -s localhost:<core-port>/api/health | jq '.status, .llm'
"connected"
true

# every chat turn after the first hang never returns (blocked on chat_turn_lock.lock().await);
# no error is logged because the read never completes and the lock is never released.
# Recovery requires:
$ sudo systemctl restart genie-core
```

### Severity: CRITICAL
Single hung backend read → total, unrecoverable chat outage across every channel, with monitoring falsely green. Recoverable only by restarting the daemon.

### Why this is structural (not a one-line fix)
Two layers are wrong together, and fixing one without the other still wedges:

1. **A whole resilience component exists but was never connected.** `RetryLlmClient` provides exactly the bounded behaviour we need, yet the facade (`LlmClient`) bypasses it. The bound belongs *inside the client/backend reads* so every call site (web, OpenAI bridge, voice, REPL, Telegram) inherits it — not bolted onto one handler.
2. **The server holds a process-wide lock across an unbounded network read.** Even with a client timeout, holding `chat_turn_lock` across the full stream serializes the whole appliance behind one slow/stuck turn.

### Suggested structural changes
- **Bound the read at the client layer.** Apply `tokio::time::timeout` to *every* `OpenAiCompatClient` read (`read_line`/`read_exact`/`next_line` and the streaming `TcpStream::connect`), not just the non-streaming connect. Make the timeout a config value (request/connect/idle-token).
- **Wire `RetryLlmClient` into the facade — or delete it.** Either have `LlmClient::from_service_config` wrap the backend in the retry/timeout/fallback layer so all call sites inherit it, or remove the dead module and fold its timeout/fallback semantics into `LlmClient` directly. Drop the `#[allow(unused_imports)]` once it's actually used.
- **Stop holding `chat_turn_lock` across the unbounded read.** Scope the lock to the turn's mutable state only, or replace the global `Mutex<()>` with a per-conversation lock and/or a bounded `with_chat_turn_lock` so one stuck stream cannot serialize the entire daemon. The streaming `biased` select should also break on the LLM-side timeout, not only on a client write error.
- **Make `/api/health` reflect chat liveness**, e.g. surface "last successful turn age" / lock-contention so a wedged chat path can't report green.

enjoyandlove:
## Summary

Bound every LLM backend read (and connect/write) so a single hung backend read
can no longer hold the chat turn lock forever and wedge chat across all channels.
Replaces the unbounded process-wide `chat_turn_lock` with a `ChatTurnGate` that
has bounded acquisition and liveness tracking, and exposes that liveness in
`/api/health` so monitoring can't show green while chat is stuck. Fixes #181.

## Changes

- Bound all `OpenAiCompatClient` I/O with configurable timeouts. Previously only
  the non-streaming connect timed out; the streaming `TcpStream::connect` and
  **every** read (`read_line` / `read_exact` / `read_to_string` / `next_line`)
  were unbounded, so a backend that accepted a connection then stalled (e.g.
  mid governor model swap) hung forever.
- Add three config keys, threaded through `LlmClient::from_service_config` so all
  call sites (web, OpenAI bridge, voice, REPL, Telegram) inherit the bound:
  - `core.llm_connect_timeout_secs` (default `10`) — TCP connect.
  - `core.llm_read_timeout_secs` (default `60`) — max idle seconds between
    tokens/bytes; generous so a cold model swap's first-token latency isn't
    mistaken for a hang.
  - `core.llm_request_timeout_secs` (default `120`) — whole non-streaming
    completion body.
- Replace the bare `Mutex<()>` `chat_turn_lock` with a `ChatTurnGate`:
  - **Bounded acquisition** — a turn that can't get the gate within budget
    returns `503 chat busy` instead of blocking forever.
  - **Liveness tracking** — `/api/health` now reports a `chat` block
    (`in_flight`, `last_turn_age_secs`, `current_turn_age_secs`, `waiters`,
    `wedged`) and flips overall status to `degraded` when a turn is wedged.
- Delete the dead, never-wired `RetryLlmClient` module (reachable only via
  `#[allow(unused_imports)]`). It was the only place a timeout previously lived;
  those semantics now live at the client read layer where every call site
  inherits them.
- Document the new keys in `doc/configuration.md` and `deploy/config/geniepod.toml`;
  add a CHANGELOG entry.

## Real Behavior Proof

- [ ] `jetson`
- [ ] `raspberry_pi`
- [ ] `portable_sbc`
- [ ] `laptop`
- [ ] `mac`
- [ ] CI-only / docs-only
- [ ] Not run locally

### What I ran

```
cargo build -p genie-core
cargo test -p genie-core llm::
cargo test -p genie-core server::
```

## Test plan

- Build and run `genie-core` against a backend that stalls after accepting the
  connection; confirm a chat turn aborts at the read timeout instead of hanging.
- Hit `/api/health` during a stuck turn and confirm the `chat` liveness block and
  `degraded` status.
- Trigger lock contention and confirm `503 chat busy` instead of an indefinite block.
- Confirm timeouts are configurable via the three new `core.*` keys.

## Notes for reviewers

- The default `llm_read_timeout_secs` (60s) is intentionally generous to avoid
  killing legitimately slow first-token latency after a cold model swap; tune
  down per profile if needed.
- `RetryLlmClient` is removed rather than fixed because it was never wired into
  any call path; its only useful behavior (a bounded read) now lives at the
  client layer for every caller.


ai-hpc (changes requested):
Reviewed and holding this PR before merge because it changes the core chat turn gate and LLM timeout behavior.

Summary: the direction is valuable, but this needs additional validation or adjustment before landing. Please add test evidence for the stuck-runtime recovery path, including concurrent chat while one turn is blocked, and confirm the chat health/degraded fields recover after the timeout. Also please make the waiter accounting cancellation-safe or document why request cancellation cannot leak the waiter count while awaiting the gate.

Thanks @enjoyandlove.

enjoyandlove:
Hi, @ai-hpc 
I have just fixed this PR. Could you please review it again?

ai-hpc (changes requested):
Still holding this PR for changes.

Summary: bounding the OpenAI-compatible timeout is valuable, but this touches chat-turn liveness and needs proof that cancellation and stuck-runtime recovery are safe.

Question: what test or device run demonstrates that the chat waiter count returns to zero if a client disconnects while waiting for the turn gate?

Thanks @enjoyandlove.

ai-hpc (approved):
Reviewed the rebuilt branch: backend connect/read/write paths are bounded, chat turn acquisition reports busy instead of hanging, and health exposes chat liveness/recovery state. Local server tests and clippy pass after rebasing on current main.

ai-hpc:
reviewd, bounded LLM backend reads and added chat-gate liveness recovery so stalled turns fail or report degraded instead of wedging chat, merged at d6cb806cb3d0
Thanks @enjoyandlove

PR #197, opened 2026-05-26
fix(api/http): bound and deadline-guard inbound request reader

### LINKED ISSUE #195: [bug] server: hand-rolled HTTP reader has no read timeout, no header-size cap, and no connection limit — one unauthenticated client OOM-kills or fd-exhausts the whole genie-core daemon
### Summary

Both backend HTTP surfaces — `genie-core` (chat/agent, `:3000`) and `genie-api` (dashboard, `:3080`) — parse inbound requests with a hand-rolled `read_line` loop that has **no read timeout, no bound on request-line/header length, and no cap on concurrent connections**, so a single unauthenticated peer on the LAN can crash the always-on home daemon (OOM, fd-exhaustion, or clean exit) at will.

### Steps to reproduce

1. Start the stack normally (`genie-core` on `:3000`, `genie-api` on `:3080`).
2. **Memory exhaustion (OOM):** send a header that never terminates with `\n`:
   ```sh
   { printf 'GET / HTTP/1.1\r\nX-Pad: '; yes A | tr -d '\n'; } | nc 127.0.0.1 3000
   ```
   `read_line` appends to the `String` without limit → RSS climbs until the OOM killer reaps `genie-core`.
3. **Slowloris / fd leak → clean exit:** open many connections that send one byte then stall:
   ```sh
   for i in $(seq 1 2000); do (printf 'G'; sleep 100000) | nc 127.0.0.1 3000 & done
   ```
   Each `read_line` awaits forever (no timeout); each connection + fd leaks. At `RLIMIT_NOFILE`, `accept().await?` returns `EMFILE` and the `?` propagates out of `serve()` → `main()` → the daemon exits.
4. Repeat against `:3080` — `genie-api` has the identical reader, with a smaller (4 KB) body cap but the same unbounded header phase.

### Expected behavior

- A request whose line/headers exceed a sane cap is rejected with **431 / 413** within bounded memory; no unbounded growth.
- An idle/partial connection is dropped after a **read/idle timeout**; N such connections do not leak fds or wedge the listener.
- Transient `accept()` errors (e.g. `EMFILE`) are **logged and the loop continues** — they never terminate the process.
- A **configurable ceiling*1. Start the stack normally (`genie-core` on `:3000`, `genie-api` on `:3080`).
2. **Memory exhaustion (OOM):** send a header that never terminates with `\n`:
   ```sh
   { printf 'GET / HTTP/1.1\r\nX-Pad: '; yes A | tr -d '\n'; } | nc 127.0.0.1 3000
   ```
   `read_line` appends to the `String` without limit → RSS climbs until the OOM killer reaps `genie-core`.
3. **Slowloris / fd leak → clean exit:** open many connections that send one byte then stall:
   ```sh
   for i in $(seq 1 2000); do (printf 'G'; sleep 100000) | nc 127.0.0.1 3000 & done
   ```
   Each `read_line` awaits forever (no timeout); each connection + fd leaks. At `RLIMIT_NOFILE`, `accept().await?` returns `EMFILE` and the `?` propagates out of `serve()` → `main()` → the daemon exits.
4. Repeat against `:3080` — `genie-api` has the identical reader, with a smaller (4 KB) body cap but the same unbounded header phase.
* on concurrent connections is enforced.

### Actual behavior

- `read_line` grows the `String` without limit and never times out → OOM kill. With `panic = "abort"` ([`Cargo.toml:39`](Cargo.toml#L39)) an allocation failure aborts the process — no unwind, no recovery.
- Stalled connections await forever; fds leak; at `RLIMIT_NOFILE` the accept loop's `?` propagates the error and the daemon exits — under a sustained flood this is a crash-loop, and each restart pays the 60–90 s STT/model cold-start (see #17), so recovery is effectively never.
- Connections are accepted and `spawn_local`'d with no semaphore, amplifying both effects.

Because `genie-core` runs the entire stack in one single-threaded process (`#[tokio::main(flavor = "current_thread")]`, [`main.rs:16`](crates/genie-core/src/main.rs#L16)) with no isolation, killing the HTTP listener kills voice and physical-actuation handling at the same time.

### Hardware

Jetson Orin Nano Super 8 GB

### JetPack / L4T version

JetPack 6.x / L4T R36.x *(hardware-independent — pure logic bug)*

`v1.0.0-alpha.9` (`92ec2ab`)

### Relevant logs

```shell
# Slowloris flood → accept() hits EMFILE → ? propagates → daemon exits
ERROR genie_core: Too many open files (os error 24)
systemd[1]: genie-core.service: Main process exited, code=exited, status=1
systemd[1]: genie-core.service: Scheduled restart job, restart counter is N
# OOM path
kernel: Out of memory: Killed process <pid> (genie-core)
```

### Root cause

`genie-core` — [`crates/genie-core/src/server.rs:243-267`](crates/genie-core/src/server.rs#L243-L267):

```rust
let mut request_line = String::new();
buf_reader.read_line(&mut request_line).await?;   // unbounded + no timeout
...
loop {
    let mut line = String::new();
    buf_reader.read_line(&mut line).await?;        // unbounded + no timeout
    if line.trim().is_empty() { break; }
    ...
}
```

- Body **is** capped (`content_length < 65536`, [`server.rs:270`](crates/genie-core/src/server.rs#L270)) — but the request line and headers are not, and the cap is never mirrored into the header phase.
- `accept().await?` ([`server.rs:139`](crates/genie-core/src/server.rs#L139)) propagates into process exit; `spawn_local` ([`server.rs:141`](crates/genie-core/src/server.rs#L141)) has no concurrency bound.
- `bind_host` is configurable to `0.0.0.0` / `::` ([`server.rs:124`](crates/genie-core/src/server.rs#L124)) and the architecture has web/mobile apps over the LAN — so the threat is **any device on the household network** (including a compromised IoT bulb), not just localhost.

`genie-api` — [`crates/genie-api/src/http.rs:36-59`](crates/genie-api/src/http.rs#L36-L59): the same `read_line` pattern, same unbounded headers, `accept().await?` at [`http.rs:20`](crates/genie-api/src/http.rs#L20), body capped at 4 KB ([`http.rs:62`](crates/genie-api/src/http.rs#L62)).

The codebase already knows how to bound I/O — every outbound probe in [`crates/genie-api/src/routes.rs`](crates/genie-api/src/routes.rs) wraps reads in `tokio::time::timeout(...)`. The inbound read path was never given the same treatment.

greatjourney589:
## Summary
Closes #195 
Bound and deadline-guard the hand-rolled inbound HTTP/1.1 request reader shared
by `genie-core` (`:3000`) and `genie-api` (`:3080`). Previously the request line
and headers grew with an unbounded `read_line` loop and there was no read
deadline, so a single unauthenticated peer on the LAN could exhaust memory (a
header that never terminates), stall the listener with half-open connections, or
fd-exhaust the always-on daemon. 

## Changes

- Add a shared, hardened reader in `genie-common::http` (`read_request`) that
  enforces size bounds (request line, per-header line, header count, total
  header bytes, body) and runs the whole read under a single
  `tokio::time::timeout`. Returns a typed `HttpReadError` that maps onto `431`
  (oversized line/headers) or `413` (oversized declared body), or signals
  "just close the connection" for stalled/vanished/garbage peers.
- Add a `[http]` config section (`HttpServerConfig`) with safe, bounded defaults
  — 8 KiB request line, 8 KiB header line, 100 headers, 64 KiB total headers,
  15 s read timeout, 256 max connections. Missing section falls back to defaults
  so existing deployments keep parsing.
- Wire both servers onto the shared reader and cap concurrency with a
  `tokio::sync::Semaphore` (permit reserved before `accept`, held for the
  request lifetime). Transient `accept()` errors (e.g. `EMFILE`) are logged and
  the accept loop continues instead of terminating the process. Per-server body
  caps unchanged (64 KiB genie-core, 4 KiB genie-api).
- Document the hardening in `doc/http-and-cli.md` and the `[http]` knobs in
  `deploy/config/geniepod.toml`.
- Add unit/integration tests covering oversized request line, oversized/
  unterminated header (bounded memory), too-many-headers, total-header-byte cap,
  oversized declared body, idle/slowloris timeout, peer close, connection flood,
  and config default/override/fallback parsing.

## Real Behavior Proof

- [ ] `jetson`
- [ ] `raspberry_pi`
- [ ] `portable_sbc`
- [x] `laptop`
- [ ] `mac`
- [ ] CI-only / docs-only
- [ ] Not run locally

### What I ran

This is a server-side networking change with no dependency on Jetson-specific
hardware, voice, audio, or Home Assistant — it is fully exercised by the
crate-level tests. On `<your OS / machine>`:

```
cargo test -p genie-common http
cargo test -p genie-core  server
cargo test -p genie-api   http
```

### What I observed

All tests pass, including:

- `oversized_request_line_is_rejected`, `oversized_request_header_is_rejected_and_server_survives`,
  `too_many_headers_is_rejected`, `total_header_bytes_cap_is_enforced`,
  `unterminated_header_is_rejected_in_bounded_memory` → reader returns `431`
  and the server stays up.
- `oversized_declared_body_is_rejected` → `413`.
- `idle_connection_times_out` / `idle_connection_is_dropped_after_read_timeout`
  → stalled connection dropped after the read deadline.
- `peer_close_before_request_is_reported`, `connection_flood_does_not_wedge_server`
  → half-open / flood traffic does not wedge the listener.
- `http_server_config_defaults_are_bounded`, `http_server_config_falls_back_when_section_absent`,
  `http_server_config_parses_overrides` → defaults applied, missing section
  tolerated, overrides honored.

`<paste raw `test result: ok. N passed; 0 failed` lines here>`

## Test plan

- `cargo test -p genie-common http` — reader bounds + config parsing.
- `cargo test -p genie-core server` and `cargo test -p genie-api http` — server
  integration (431/413 responses, read timeout, connection-flood survival).
- Manual smoke (optional): with the daemon running, send an oversized header and
  confirm a `431`; open a connection and send nothing, confirm it is dropped
  after `read_timeout_secs`.

## Notes for reviewers

- Bounds and the connection cap are configurable via the new `[http]` section;
  defaults are intentionally generous for first-party clients while still
  bounding fan-out. The per-request body cap stays per-server and is not part of
  `[http]`.
- Connection-count ceiling and accept-loop resilience live in each server's
  listener (semaphore + log-and-continue), not in the shared per-request reader,
  by design.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -189,6 +189,7 @@ dependencies = [
  "serde",
  "serde_json",
  "tokio",
+ "toml",
  "tracing",
  "tracing-subscriber",
 ]
@@ -200,6 +201,7 @@ dependencies = [
  "anyhow",
  "serde",
  "serde_json",
+ "tokio",
  "toml",
  "tracing",
 ]
--- a/crates/genie-api/Cargo.toml
+++ b/crates/genie-api/Cargo.toml
@@ -18,3 +18,6 @@ tracing = { workspace = true }
 tracing-subscriber = { workspace = true }
 anyhow = { workspace = true }
 rusqlite = { workspace = true }
+
+[dev-dependencies]
+toml = { workspace = true }
--- a/crates/genie-api/src/http.rs
+++ b/crates/genie-api/src/http.rs
@@ -1,74 +1,108 @@
+use std::sync::Arc;
+use std::time::Duration;
+
 use anyhow::Result;
 use genie_common::config::Config;
-use tokio::io::{AsyncBufReadExt, AsyncWriteExt, BufReader};
+use genie_common::http::{HttpLimits, read_request};
+use tokio::io::{AsyncWriteExt, BufReader};
 use tokio::net::TcpListener;
+use tokio::net::tcp::OwnedWriteHalf;
+use tokio::sync::Semaphore;
 
 use crate::routes;
+
+/// Largest request body genie-api will read into memory. The dashboard only
+/// sends tiny JSON control payloads, so this stays well below genie-core's
+/// 64 KiB cap. Mirrored into the header phase via `[http].max_header_bytes`.
+const API_MAX_BODY_BYTES: usize = 4 * 1024;
 
 /// Minimal HTTP/1.1 server — no framework, no allocator overhead.
 ///
 /// Handles one request per connection (Connection: close).
 /// This is intentional: the dashboard polls every 5 seconds,
 /// and the API serves <10 concurrent clients on a home appliance.
+///
+/// The inbound reader is bounded and deadline-guarded (issue #195): oversized
+/// request lines/headers are rejected, idle connections are dropped after a
+/// read timeout, transient `accept()` errors never terminate the daemon, and
+/// concurrent connections are capped by a semaphore.
 pub async fn serve(addr: &str, config: Config) -> Result<()> {
     let listener = TcpListener::bind(addr).await?;
-    let config = std::sync::Arc::new(config);
-
     tracing::info!(addr, "listening");
+    serve_listener(listener, config).await
+}
+
+/// Accept connections from an already-bound `TcpListener`.
+///
+/// Split out from [`serve`] so tests can pre-bind to port 0 and drive the
+/// hardened reader/accept loop directly.
+pub(crate) async fn serve_listener(listener: TcpListener, config: Config) -> Result<()> {
+    let config = Arc::new(config);
+    let limits = HttpLimits::from_config(&config.http, API_MAX_BODY_BYTES);
+    let max_connections = config.http.max_connections.max(1);
+    // Bound concurrently handled connections so a flood cannot exhaust fds.
+    let semaphore = Arc::new(Semaphore::new(max_connections));
 
     loop {
-        let (stream, peer) = listener.accept().await?;
+        // Reserve a slot before accepting; connections beyond the ceiling stay
+        // parked in the OS backlog rather than being spawned unbounded.
+        let permit = match Arc::clone(&semaphore).acquire_owned().await {
+            Ok(permit) => permit,
+            Err(_) => break, // semaphore closed — shutting down
+        };
+        let (stream, peer) = match listener.accept().await {
+            Ok(accepted) => accepted,
+            Err(e) => {
+                // A transient accept() error (e.g. EMFILE under a connection
+                // flood) must never propagate out and terminate the daemon.
+                tracing::warn!(error = %e, "accept failed; continuing");
+                drop(permit);
+                tokio::time::sleep(Duration::from_millis(50)).await;
+                continue;
+            }
+        };
         let config = config.clone();
 
         tokio::spawn(async move {
-            if let Err(e) = handle_connection(stream, &config).await {
+            // Hold the permit for the lifetime of the request.
+            let _permit = permit;
+            if let Err(e) = handle_connection(stream, &config, &limits).await {
                 tracing::debug!(peer = %peer, error = %e, "connection error");
             }
         });
     }
+
+    Ok(())
 }
 
-async fn handle_connection(stream: tokio::net::TcpStream, config: &Config) -> Result<()> {
+async fn handle_connection(
+    stream: tokio::net::TcpStream,
+    config: &Config,
+    limits: &HttpLimits,
+) -> Result<()> {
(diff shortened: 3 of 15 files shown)


ai-hpc:
Reviewed and merged at bd2bd4a2a06bf48cf13e03796088e0665213a5d5 (bd2bd4a).

Summary: centralizes bounded, deadline-guarded HTTP request reading for genie-core and genie-api to reduce slowloris and oversized-request risk.

Thanks @greatjourney589.

PR #198, opened 2026-05-26
fix(server): bound HTTP reader, add timeouts, recover from accept errors (#195)

### LINKED ISSUE #195: [bug] server: hand-rolled HTTP reader has no read timeout, no header-size cap, and no connection limit — one unauthenticated client OOM-kills or fd-exhausts the whole genie-core daemon

Rene0422:
## Summary

Fixes #195. Bounds the hand-rolled HTTP parsers in `genie-core` (port 3000) and `genie-api` (port 3080) so an unauthenticated peer cannot OOM the daemon with unterminated headers, exhaust fds with stalled connections, or crash the process via a transient `accept()` failure (EMFILE).

## Changes

- New `http_safety` module in both `genie-core` and `genie-api` (duplicated rather than placed in `genie-common` to avoid adding tokio to that crate's dep surface). Exports `read_capped_line()` — bounded allocation via `AsyncReadExt::take`, stall protection via `tokio::time::timeout` — plus the limit constants and a `build_error_response()` helper.
- `genie-core/src/server.rs` and `genie-api/src/http.rs` route every `read_line` call through `read_capped_line`. Over-cap reads return **431 Request Header Fields Too Large**, over-cap request lines return **414 URI Too Long**, stalled reads return **408 Request Timeout**, and over-cap bodies return **413 Payload Too Large**.
- Accept loop in both servers now acquires a `Semaphore` permit *before* calling `accept()`, so a 257th (core) / 129th (api) client blocks at the OS socket queue instead of spawning a task and consuming an fd.
- `accept()` errors are logged with a 100 ms backoff and the loop continues, instead of being propagated into process exit. Transient EMFILE no longer kills the daemon.
- Limits applied: request line 8 KB, per-header line 8 KB, total headers 64 KB, header read deadline 10 s, body read deadline 30 s, max connections 256 (core) / 128 (api).
- 5 unit tests on `read_capped_line` (normal / over-cap / EOF / timeout / error-response format) and 2 end-to-end server tests that drive a real `TcpStream` against `ChatServer::serve_listener`.

## Real Behavior Proof

- [ ] `jetson`
- [ ] `raspberry_pi`
- [ ] `portable_sbc`
- [x] `laptop`
- [ ] `mac`
- [ ] CI-only / docs-only
- [ ] Not run locally

### What I ran

Linux laptop, x86_64, rustc 1.95.0:

cargo build -p genie-core -p genie-api --bins --lib
cargo test --workspace --tests
cargo test -p genie-core --lib server::tests::oversized_request_line_is_rejected_and_server_survives
cargo test -p genie-core --lib server::tests::stalled_request_is_timed_out
cargo test -p genie-core --lib http_safety
cargo test -p genie-api  --lib http_safety
cargo clippy -p genie-core -p genie-api --lib --bins --tests --all-targets -- -D warnings

The two end-to-end tests drive a real `tokio::net::TcpListener` on `127.0.0.1:0` through `ChatServer::serve_listener`, then connect a raw `TcpStream` and replay each of the three attack shapes from the issue:

1. 16 KB request line with no CRLF (memory-exhaustion attempt)
2. Partial request that never sends CRLF (fd-exhaustion / slowloris attempt)
3. Follow-up `GET /api/health` after the attack to confirm the daemon survived (covers the "EMFILE kills the daemon" failure mode by proving the accept loop keeps running)

### What I observed

test http_safety::tests::reads_complete_line_within_cap ... ok
test http_safety::tests::returns_too_long_when_line_exceeds_cap ... ok
test http_safety::tests::returns_eof_on_empty_input ... ok
test http_safety::tests::returns_timeout_when_reader_stalls ... ok
test http_safety::tests::error_response_includes_status_and_body ... ok
test server::tests::oversized_request_line_is_rejected_and_server_survives ... ok
test server::tests::stalled_request_is_timed_out ... ok (finished in 10.01s — header read deadline fired as designed)

test result: ok. 463 passed; 0 failed   (genie-core lib)
test result: ok. 11 passed;  0 failed   (genie-api)
cargo clippy -D warnings: clean

Concrete behavior observed on the attack paths:

- 16 KB of `A` with no CRLF → server replied `HTTP/1.1 414 URI Too Long`, closed the connection, and a subsequent `GET /api/health` on a fresh socket returned `HTTP/1.1 200`. Allocation stayed bounded at the 8 KB request-line cap.
- Partial `GET /api/health` (no CRLF) → server closed the socket after `HEADER_READ_TIMEOUT` (10 s). `read_to_end` on the client returned within the test's 15 s ceiling, confirming the stall guard fires.

**Jetson gap**: this is a transport-layer fix with no Jetson-specific surface (no voice, no Home Assistant, no GPU, no `tegrastats`). Behavior on `aarch64-l4t` is identical to the laptop run because tokio, `TcpListener`, and `AsyncBufRead` are the same on both. I'd still appreciate a reviewer with a Jetson sanity-checking that the 5 s dashboard poll cadence isn't accidentally throttled by `MAX_CONCURRENT_CONNECTIONS = 128` on `genie-api` — the ceiling is well above expected load (the dashboard opens ~1 connection per poll) but worth eyeballing the first time it lands on real hardware.

## Test plan

- `cargo test --workspace --tests` — should be green on any host with a working rustc.
- Manual repro of the original CVE shape, from a separate machine on the LAN against a running `genie-core`:
  - `python3 -c "import socket; s=socket.create_connection(('genie.local', 3000)); s.send(b'A'*16384); print(s.recv(64))"` → expect `HTTP/1.1 414 ...`, not a hung connection.
  - `python3 -c "import socket,time; s=socket.create_connection(('genie.local', 3000)); s.send(b'GET /api/health'); time.sleep(15); print(s.recv(64))"` → expect the server to have closed the socket within the 10 s header deadline.
  - Open 300 simultaneous TCP connections that never write anything — confirm `genie-core` stays responsive on a parallel `curl http://genie.local:3000/api/health` and that `dmesg` / journalctl show no panic or restart.

## Notes for reviewers

- Open question for the maintainer: the `http_safety` module is duplicated across `genie-core` and `genie-api` rather than placed in `genie-common`. I chose duplication because lifting it would add tokio to `genie-common`'s dependency surface, and that crate today is `serde` + `toml` + `tracing` only. If you'd rather pay the dep cost to dedupe, I'm happy to move it in a follow-up.
- `MAX_CONCURRENT_CONNECTIONS` is a hard-coded constant per crate (256 / 128). Configurable values via `geniepod.toml` felt out of scope for a security fix; happy to add a follow-up issue if you want them tunable.
- The stalled-request test takes ~10 seconds because it exercises the real `HEADER_READ_TIMEOUT`. Making the timeout `pub`-mutable just for tests felt like the wrong trade — exposing a tuning knob purely for test speed-ups invites real callers to override it badly.
- Deferred follow-up: the streaming `/api/chat/stream` path has its own write-side timeout story (LLM token cadence) that isn't covered here. Issue #195 is about the read side, which is where the DoS landed.

### DIFF
--- a/crates/genie-api/src/http.rs
+++ b/crates/genie-api/src/http.rs
@@ -1,8 +1,17 @@
+use std::sync::Arc;
+use std::time::Duration;
+
 use anyhow::Result;
 use genie_common::config::Config;
-use tokio::io::{AsyncBufReadExt, AsyncWriteExt, BufReader};
+use tokio::io::{AsyncWriteExt, BufReader};
 use tokio::net::TcpListener;
+use tokio::sync::Semaphore;
 
+use crate::http_safety::{
+    BODY_READ_TIMEOUT, HEADER_READ_TIMEOUT, LineRead, MAX_CONCURRENT_CONNECTIONS,
+    MAX_HEADER_LINE_BYTES, MAX_REQUEST_LINE_BYTES, MAX_TOTAL_HEADER_BYTES, build_error_response,
+    read_capped_line,
+};
 use crate::routes;
 
 /// Minimal HTTP/1.1 server — no framework, no allocator overhead.
@@ -10,17 +19,38 @@ use crate::routes;
 /// Handles one request per connection (Connection: close).
 /// This is intentional: the dashboard polls every 5 seconds,
 /// and the API serves <10 concurrent clients on a home appliance.
+///
+/// The parser is bounded along three axes — see [`crate::http_safety`] for
+/// the constants — so an unauthenticated peer cannot exhaust memory, fds, or
+/// crash the daemon by sending malformed or stalled requests.
 pub async fn serve(addr: &str, config: Config) -> Result<()> {
     let listener = TcpListener::bind(addr).await?;
     let config = std::sync::Arc::new(config);
+    let connection_permits = Arc::new(Semaphore::new(MAX_CONCURRENT_CONNECTIONS));
 
     tracing::info!(addr, "listening");
 
     loop {
-        let (stream, peer) = listener.accept().await?;
-        let config = config.clone();
+        let permit = match Arc::clone(&connection_permits).acquire_owned().await {
+            Ok(p) => p,
+            Err(_) => break Ok(()),
+        };
 
+        let (stream, peer) = match listener.accept().await {
+            Ok(pair) => pair,
+            Err(e) => {
+                // EMFILE / other transient accept failures must NOT terminate
+                // the daemon. Drop the permit, back off, and try again.
+                tracing::warn!(error = %e, "accept failed; backing off before retrying");
+                drop(permit);
+                tokio::time::sleep(Duration::from_millis(100)).await;
+                continue;
+            }
+        };
+
+        let config = config.clone();
         tokio::spawn(async move {
+            let _permit = permit;
             if let Err(e) = handle_connection(stream, &config).await {
                 tracing::debug!(peer = %peer, error = %e, "connection error");
             }
@@ -32,9 +62,33 @@ async fn handle_connection(stream: tokio::net::TcpStream, config: &Config) -> Re
     let (reader, mut writer) = stream.into_split();
     let mut buf_reader = BufReader::new(reader);
 
-    // Read the request line.
+    // Read the request line under a hard byte cap and timeout. An unbounded
+    // read_line() would let any peer pin this task and grow the buffer
+    // without limit.
     let mut request_line = String::new();
-    buf_reader.read_line(&mut request_line).await?;
+    match read_capped_line(
+        &mut buf_reader,
+        &mut request_line,
+        MAX_REQUEST_LINE_BYTES,
+        HEADER_READ_TIMEOUT,
+    )
+    .await?
+    {
+        LineRead::Ok => {}
+        LineRead::TooLong => {
+            let _ = writer
+                .write_all(build_error_response(414, "request line too long").as_bytes())
+                .await;
+            return Ok(());
+        }
+        LineRead::Timeout => {
+            let _ = writer
+                .write_all(build_error_response(408, "request timed out").as_bytes())
+                .await;
+            return Ok(());
+        }
+        LineRead::Eof => return Ok(()),
+    }
 
     // Parse method and path.
     let parts: Vec<&str> = request_line.split_whitespace().collect();
@@ -44,12 +98,45 @@ async fn handle_connection(stream: tokio::net::TcpStream, config: &Config) -> Re
     let method = parts[0];
     let path = parts[1];
 
-    // Drain headers (we don't need them for our simple API).
-    let mut header_line = String::new();
+    // Drain headers under per-line and total caps. We don't need the values
+    // for most endpoints, but the parser must still bound allocation.
     let mut content_length: usize = 0;
+    let mut total_header_bytes = request_line.len();
+    let mut header_line = String::new();
     loop {
         header_line.clear();
-        buf_reader.read_line(&mut header_line).await?;
+        let outcome = read_capped_line(
+            &mut buf_reader,
+            &mut header_line,
+            MAX_HEADER_LINE_BYTES,
+            HEADER_READ_TIMEOUT,
+        )
+        .await?;
+        match outcome {
+            LineRead::Ok => {}
+            LineRead::TooLong => {
+                let _ = writer
+                    .write_all(build_error_response(431, "header line too long").as_bytes())
+                    .await;
+                return Ok(());
+            }
+            LineRead::Timeout => {
+                let _ = writer
+                    .write_all(build_error_response(408, "request timed out").as_bytes())
+                    .await;
+                return Ok(());
+            }
+            LineRead::Eof => return Ok(()),
+        }
+
+        total_header_bytes = total_header_bytes.saturating_add(header_line.len());
+        if total_header_bytes > MAX_TOTAL_HEADER_BYTES {
+            let _ = writer
+                .write_all(build_error_response(431, "headers too large").as_bytes())
+                .await;
+            return Ok(());
+        }
+
         if header_line.trim().is_empty() {
             break;
         }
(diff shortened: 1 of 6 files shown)


ai-hpc:
Reviewed. This overlaps with #197 for issue #195, and #197 has now been reviewed, machine-tested by CI, and merged. This PR does not show cargo/clippy/cross-compile machine checks in the PR status, so I am closing the untested duplicate.

Thanks @Rene0422.

PR #201, opened 2026-05-26
fix(memory): LLM args can no longer bypass person-scoped memory policy gate

dripsmvcp:
## Problem

`exec_memory_recall` fell back to reading `identity_confidence`,
`explicit_named_person`, and related policy fields directly from the LLM's
tool-call arguments when `exec_ctx.memory_read_context` was `None`. Because
the server never sets `memory_read_context` on the execution context, this
fallback was always active — meaning any LLM response (including one shaped
by a prompt-injection attack) could include:

{"name": "memory_recall", "arguments": {"query": "…", "identity_confidence": "high", "explicit_named_person": true}}

…and the person-scoped memory policy gate would pass, disclosing memories
that should only be accessible when the voice/app pipeline has verified the
speaker's identity.

## Fix

One-line production change in `exec_memory_recall`
(`crates/genie-core/src/tools/dispatch.rs`):

Before:
.unwrap_or_else(|| memory_read_context(args))

After:
.unwrap_or_else(crate::memory::policy::MemoryReadContext::shared_room_voice)

When no verified context is present, the fallback is now the most
conservative policy (shared_room_voice: identity unknown, shared space).
The correct unlock path — setting `exec_ctx.memory_read_context` in the
server or voice pipeline from a verified speaker identity — was already
implemented and continues to work unchanged.

The now-unreachable `fn memory_read_context(args)` (30 lines) is deleted.

## Tests

- New regression `memory_recall_llm_args_cannot_elevate_identity_confidence`:
  asserts that `identity_confidence: "high"` in LLM args does not unlock
  person-scoped memories when `exec_ctx.memory_read_context` is `None`.
- Updated `memory_recall_can_use_identity_context_when_provided` →
  `memory_recall_unlocks_person_memory_with_verified_exec_ctx`: now tests the
  correct contract — identity supplied via `exec_ctx`, not args.
- All 512 tests pass (`cargo test -p genie-core`).

## Scope

Single file. 1 production line changed, 30 lines deleted. No API, config, or
protocol changes.

## Real Behavior Proof

`cargo test -p genie-core`

### What I observed

All 512 tests passed. The new regression
`memory_recall_llm_args_cannot_elevate_identity_confidence` confirmed that
passing `identity_confidence: "high"` in LLM tool-call arguments no longer
grants access to person-scoped memories when `exec_ctx.memory_read_context`
is `None`. The updated `memory_recall_unlocks_person_memory_with_verified_exec_ctx`
confirmed the legitimate unlock path — identity supplied by the server via
`exec_ctx` — continues to work correctly.

This fix is in the in-process policy gate with no hardware-specific code paths;
the Jetson runtime and laptop runtime exercise the identical Rust code.

### DIFF
--- a/crates/genie-core/src/tools/dispatch.rs
+++ b/crates/genie-core/src/tools/dispatch.rs
@@ -863,7 +863,7 @@ impl ToolDispatcher {
         let query = memory_query(args);
         let read_context = exec_ctx
             .memory_read_context
-            .unwrap_or_else(|| memory_read_context(args));
+            .unwrap_or_else(crate::memory::policy::MemoryReadContext::shared_room_voice);
 
         let results = crate::memory::recall::recall_with_context(&mem, query, 10, read_context)?;
         if results.is_empty() {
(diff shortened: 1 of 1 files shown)


ai-hpc:
Reviewed this direction and the memory policy change looks valuable, but this PR has not run any real machine checks in the PR status: I only see the PR body checklist and no cargo test, clippy, or cross-compile run. Closing for now per project triage policy until it is updated/reopened with machine test coverage.

Thanks @dripsmvcp.

PR #205, opened 2026-05-26
fix(weather): enforce read deadline + size cap + percent-encode geocode (closes #203)

### LINKED ISSUE #203: [bug] tools/weather: unbounded HTTP read + silent chunked corruption + bare-spaces URL encoding lose chat or return wrong location
### Summary

`tools::weather::http_get` in `crates/genie-core/src/tools/weather.rs:182-228` has the **same unbounded-read pattern** that PR #174 (closes #173) just fixed in `ha::client::http_request`: a 10-second `TcpStream::connect` timeout, then **no timeout** on `writer.write_all`, the status-line / header `read_line`, or the body-accumulation loop. The body loop also has **no size cap**, so a slow or misbehaving Open-Meteo (or a man-in-the-middle on the plain-HTTP connection, since the comment at line 181 explicitly disables TLS) can hang the chat task forever or grow RSS without bound. Two adjacent bugs live in the same function family:

- The "chunked transfer encoding" handler at [weather.rs:216-225](crates/genie-core/src/tools/weather.rs#L216-L225) doesn't actually decode chunked encoding; it splits on `\r\n` and concatenates the non-hex-only lines **without a separator**, so chunked JSON spanning multiple chunks is silently corrupted. The detection heuristic also misfires when the response body legitimately begins with a hex-looking value.
- URL encoding at [weather.rs:71](crates/genie-core/src/tools/weather.rs#L71) only handles spaces (`location.replace(' ', "+")`). Reserved characters `&`, `?`, `#`, `=`, `%`, `+` pass through unescaped into the query string. A user asking the weather in *"Cafe & Bar Tokyo"* generates `name=Cafe+&+Bar+Tokyo&count=1…` — Open-Meteo sees the request as `name=Cafe+`, `+Bar+Tokyo=` (empty), and silently returns no result. Same class of silent correctness loss as #164 (dashboard latency URL drift) but for end-user input.

All three concerns share `http_get` and the three callers (`geocode`, `fetch_weather`, `fetch_forecast`). Fixing them together gives one focused PR over one file family.

### Steps to reproduce

Three independent reproducers — each demonstrates a different concern.

**1. Hung Open-Meteo blocks the chat task forever (timeout):**

```bash
# Spawn a fake Open-Meteo listener on port 80 that accepts and hangs:
sudo python3 - <<'PY' &
import socket, time
s = socket.socket(); s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
s.bind(("127.0.0.1", 80)); s.listen(1)
while True:
    c, _ = s.accept(); time.sleep(600); c.close()
PY

# Point Open-Meteo lookups at 127.0.0.1 via /etc/hosts, then ask for weather:
echo "127.0.0.1 api.open-meteo.com geocoding-api.open-meteo.com" | sudo tee -a /etc/hosts
curl -m 120 -X POST http://127.0.0.1:3000/api/chat \
  -d '{"message":"what is the weather in Denver"}'
# → curl gives up after 120 s; genie-core's tool-dispatch task hangs forever.
```

**2. Oversized response grows genie-core's RSS:** point the daemon at a mock server that streams unlimited bytes after the headers; `http_get`'s body loop accumulates into a `String` with no cap.

**3. URL-encoding gap (silent wrong answer):**

```bash
# Ask for a location containing reserved URL chars:
  -d '{"message":"weather in Q&A Cafe Tokyo"}'
# → the geocode HTTP request line is:
#   GET /v1/search?name=Q&A+Cafe+Tokyo&count=1&language=en&format=json
#   (Open-Meteo parses the `&A+Cafe+Tokyo` as a separate query param;
#    `name=Q` is the only useful piece; result is "location not found"
#    or the wrong city named "Q".)
```

### Expected behavior

- `http_get` enforces a TOTAL request-lifecycle deadline (default 15 s), not just on `TcpStream::connect`. A hung Open-Meteo surfaces as `anyhow!("Open-Meteo {} timed out after Ns")` within the budget, never an indefinite hang.
- The body-read loop has a maximum size (default 1 MiB) so a misbehaving Open-Meteo can never grow RSS without bound. Exceeding the cap surfaces a clear `anyhow!("Open-Meteo response exceeded N bytes")`.
- The function refuses `Transfer-Encoding: chunked` with an explicit `Err`, instead of silently corrupting the body via the broken heuristic at lines 216-225. (Open-Meteo's free-tier API doesn't actually use chunked encoding — the fallback decoder was always defense-in-depth that did more harm than good.)
- Location queries percent-encode every RFC 3986 reserved character (`&`, `?`, `#`, `=`, `%`, `+`, ` `, `/`, `<`, `>`, `"`, `'`) before going into the query string. The example query string for *"Q&A Cafe Tokyo"* becomes `name=Q%26A%20Cafe%20Tokyo&count=1&language=en&format=json`. Open-Meteo's geocoder receives the intended name.
- Default `geniepod.toml` and `geniepod.dev.toml` see no behaviour change for healthy Open-Meteo and ASCII / space-only location names.

### Actual behavior

- Lines 182-189 wrap `TcpStream::connect` in a 10-second `tokio::time::timeout`. The subsequent `writer.write_all` (line 197), the body loop's `buf_reader.read_line` (line 205), and the post-loop chunked-fallback (lines 216-225) all run with no timeout, no size cap, and no `Transfer-Encoding` validation.

- The "chunked decoder" at lines 216-225 looks plausible but is wrong on two axes: (a) joining non-hex lines with `decoded.push_str(line);` skips any inter-line content separator (a chunked JSON object spanning two chunks ends up as one long line with no whitespace between the closing brace of chunk 1 and the opening brace of chunk 2), and (b) the heuristic `body.starts_with(|c: char| c.is_ascii_hexdigit())` fires on any response whose first body character is a hex digit — including a perfectly valid JSON response whose first key happens to be a hex-looking string.

- Line 71 `location.replace(' ', "+")` only handles ASCII spaces. The query string is constructed via `format!("/v1/search?name={}&count=1&language=en&format=json", encoded)`, so any `&` in the user's location terminates the `name` parameter at the HTTP level. Open-Meteo's geocoder returns "location not found" or matches whatever the truncated `name` resolves to. The user sees "I couldn't find the weather for *Q&A Cafe Tokyo*" with no diagnostic.

### Hardware

Non-Jetson (x86_64 dev / cross-build host) and Jetson alike — the bug is in pure tokio I/O scheduling and string formatting, no hardware path. Triggered by anything that pauses Open-Meteo's HTTP response after the TCP handshake, by sending an unbounded body, OR by any user / LLM query whose location string contains reserved URL characters.

### JetPack / L4T version

`main @ f821eda` (upstream HEAD as of filing) / `crates/genie-core/src/tools/weather.rs:65-228`. Never touched since the workspace scaffold commit `0a7363b`.

### Relevant logs

```shell
# Hung Open-Meteo scenario — chat task hangs, no diagnostic from genie-core
$ journalctl -u genie-core --since '5 minutes ago' | tail -15
… genie-core[12345]: INFO  POST /api/chat origin=Api status=in_progress
…
# (no further output until either the client gives up OR /etc/hosts is reverted)
```

```text
# URL-encoding gap — wrong geocode request sent on the wire (tcpdump on lo or eth0)
$ sudo tcpdump -i lo -A 'port 80 and host geocoding-api.open-meteo.com' -c 5
GET /v1/search?name=Q&A+Cafe+Tokyo&count=1&language=en&format=json HTTP/1.1
Host: geocoding-api.open-meteo.com:80
User-Agent: GeniePod/0.2
Accept: application/json
Connection: close

# Open-Meteo's response:
HTTP/1.1 200 OK
…
{"generationtime_ms":0.058}    # empty — no "results" key — "location not found"
```

### Additional context

**Symptom users see:**
1. *"The assistant froze on a weather question."* Triggered any time Open-Meteo throttles, returns a slow response, or has a brief outage — happens to all free public APIs on a recurring basis.
2. *"Why can't Genie find weather for X?"* For any household whose place name contains `&` (`"Q&A Cafe"`, `"Mom & Pop Diner"`, *"Cafe & Bar near Tokyo"*), or transliterations involving `+` (some German / Romance compound names), or `#` (rare but used in some place names), the geocoder silently fails. Operators have no way to tell from the log whether the input was malformed, the API was down, or the location truly doesn't exist.
3. *Unbounded memory growth on a misbehaving / hostile upstream.* The plain-HTTP connection means a network attacker can feed unlimited bytes into the response loop. The accumulated `body` string grows without limit until OOM-kill.

- `crates/genie-core/src/tools/weather.rs::tests` (lines 253-263) covers only `wmo_code_to_description`. There's no test of `http_get`, `geocode`, `fetch_weather`, or any failure-mode coverage at all.
- The 3 `parse_http_url_*` tests in `ha/client.rs` and the 5 new regressions in PR #174 are the closest precedent for the right test shape (mock `TcpListener` against `http_get`); none of them reach into `weather.rs`.

**Files to touch:**
- `crates/genie-core/src/tools/weather.rs`:
  - New constants near the top: `WEATHER_CONNECT_TIMEOUT = 10s`, `WEATHER_REQUEST_TIMEOUT = 15s`, `WEATHER_MAX_RESPONSE_BYTES = 1 MiB`.
  - `http_get` (lines 182-228) refactored: use the new connect-timeout; wrap the post-connect block in `tokio::time::timeout(WEATHER_REQUEST_TIMEOUT, …)` and map `Elapsed` to a clear `anyhow!`; cap body accumulation with a check-and-bail when total bytes would exceed `WEATHER_MAX_RESPONSE_BYTES`; remove the broken chunked-encoding heuristic at lines 216-225 and instead detect `Transfer-Encoding: chunked` in the headers and bail with an explicit unsupported-encoding `Err`.
  - New helper `fn url_encode_query_param(s: &str) -> String` that percent-encodes every RFC 3986 reserved character + non-ASCII bytes. Hand-rolled, no new dependency. ~15 lines.
  - Update `geocode` (line 71) and the two callers in `fetch_weather` / `fetch_forecast` (lines 105 and 150 — wherever the location is interpolated) to use the helper instead of `location.replace(' ', "+")`.

- `crates/genie-core/src/tools/weather.rs::tests` — 5 new regressions, all hardware-free:
  - `url_encode_query_param_percent_encodes_reserved_chars` — direct helper coverage: ASCII pass-through; `&`, `#`, `=`, `+`, `?`, `%`, spaces; multi-byte UTF-8 (e.g. `"München"` → `M%C3%BCnchen`).
  - `hung_server_after_connect_times_out_cleanly` — mock `TcpListener` that accepts then sleeps forever; assert `http_get` returns `Err` containing "timed out" inside the test budget (e.g. 300 ms with a millisecond-scale test override).
  - `oversized_response_is_size_capped` — mock listener that writes a valid 200 header then streams bytes past the cap; assert clean `Err` containing "exceeded N bytes".
  - `chunked_encoding_is_explicitly_rejected` — mock listener that returns `Transfer-Encoding: chunked` with one chunk; assert `Err` containing "chunked encoding not supported" (instead of the silent body corruption today).
  - `geocode_request_line_contains_percent_encoded_location` — mock listener that captures the raw request bytes; call `geocode("Q&A Cafe Tokyo")`; assert the captured request line contains `name=Q%26A` and does NOT contain `name=Q&A`.

**Acceptance (PR):**
- `cargo test -p genie-core --lib tools::weather` — the 1 existing test still passes, 5 new tests pass.
- `cargo fmt --check`, `cargo clippy -D warnings`, `cargo test --workspace --no-default-features` — all green on both feature axes.
- Pointing `genie-core` at a `kill -STOP`ped fake Open-Meteo returns an explicit `Err("Open-Meteo … timed out after 15s")` within ~15 s — no daemon hang.
- A chat turn for *"weather in Q&A Cafe Tokyo"* produces a request line containing `name=Q%26A%20Cafe%20Tokyo` (verifiable via `tcpdump` if the operator has a real Open-Meteo) — the geocoder returns a hit instead of empty.
- Default `geniepod.toml` and `geniepod.dev.toml` see byte-identical behaviour for healthy Open-Meteo and ASCII / space-only location names (regression on the working path).

**Timing:** Standalone — no upstream PRs touch `weather.rs`. Lands independently of any open PR. No merge-conflict surface with PR #176 (skills signature), PR #182 (LLM timeouts), PR #192 (https service URLs), PR #202 (health check_http).

**Real Behavior Proof:** `cargo test -p genie-core --lib tools::weather` showing the 5 new tests pass alongside the existing `wmo_codes` test. Plus optional manual proof: spawn a mock listener that hangs, assert `http_get` returns `Err` in ~15 s instead of hanging.

**Related:**
- **#173 / PR #174 (mine)** — fixed the same unbounded-read pattern in `ha/client.rs`. This report is the sibling fix in `tools/weather.rs`.
- **#181 / PR #182** (open) — fixes timeouts on the OpenAI-compat LLM client. Different file (`llm/openai_compat.rs`), different protocol nuances. The three (HA, LLM, Weather) share the "raw-TCP HTTP client with insufficient read timeouts" theme; landing all three would close the "unbounded read" class workspace-wide.
- **#185 / PR #202** (open) — rejects non-HTTP TCP responses in `genie-health::check_http`. Different concern (validity of the response shape, not the request lifecycle).
- **#164 / PR #165** — silent correctness loss from a magic-literal URL drift. The URL-encoding gap reported here is the same silent-correctness class, in a different code path.

galuis116:
## Summary

Three sibling concerns in `crates/genie-core/src/tools/weather.rs`'s
raw-TCP HTTP client, closed by one PR over one file:

1. **No read timeout.** `http_get` (pre-fix lines 182-228) had a 10 s
   `TcpStream::connect` timeout but nothing on `write_all` / `read_line`
   afterwards. A slow / hung Open-Meteo wedged the chat task forever.
   Same crash mode as bug #173 / PR #174 for `ha/client`.
2. **No body-size cap.** The body-accumulation loop had no bound. A
   misbehaving (or man-in-the-middle, since the connection is plain
   HTTP per the file comment) response could grow RSS without limit.
3. **Broken chunked-encoding decoder** at pre-fix lines 216-225 joined
   non-hex lines with `push_str` (no separator → silent JSON corruption
   across chunks) and misfired its heuristic on legitimate responses
   whose first body char was a hex digit.
4. **URL encoding only handled spaces** at pre-fix line 68. Any reserved
   RFC 3986 char in the location (`&`, `?`, `#`, `=`, `%`, `+`) leaked
   into the geocode query string. "Q&A Cafe Tokyo" produced
   `name=Q&A+Cafe+Tokyo&count=1…` — Open-Meteo parsed it as `name=Q`
   and silently returned no match.

Closes #203.

## Changes

- `crates/genie-core/src/tools/weather.rs`:
  - New constants `WEATHER_CONNECT_TIMEOUT = 10s`,
    `WEATHER_REQUEST_TIMEOUT = 15s`, `WEATHER_MAX_RESPONSE_BYTES = 1 MiB`.
    Doc-commented with the failure mode each one prevents.
  - `http_get(host, path)` becomes a thin wrapper around new
    `pub(crate) async fn http_get_with_limits(host, port, path,
    connect_timeout, request_timeout, max_bytes)` so the test module can
    drive an ephemeral mock listener at millisecond-scale.
  - Connect uses `tokio::time::timeout(connect_timeout, …)` — clear
    `"Open-Meteo connect to {addr} timed out"` on `Elapsed`.
  - Post-connect (write + read) is wrapped in
    `tokio::time::timeout(request_timeout, …)` — clear
    `"Open-Meteo GET {path} timed out after {N}s"` on `Elapsed`.
  - New helper `read_http_get_body(reader, max_bytes)`: bails on
    `Transfer-Encoding: chunked` with an explicit
    `"Open-Meteo response uses unsupported chunked encoding"`
    (replacing the old silently-corrupting decoder); bounds the body
    `String` with a check-and-bail on every appended line.
  - New helper `url_encode_query_param(s: &str) -> String` percent-
    encodes every non-RFC-3986-unreserved byte; multi-byte UTF-8 is
    encoded byte-by-byte (e.g. `"München"` → `M%C3%BCnchen`). Hand-
    rolled, no new dependency.
  - `geocode()` swaps `location.replace(' ', "+")` for
    `url_encode_query_param(location)`.

- `crates/genie-core/src/tools/weather.rs::tests` — 5 new regressions:
  - `url_encode_query_param_percent_encodes_reserved_chars` — unit
    coverage on the helper: ASCII pass-through, every reserved char,
    multi-byte UTF-8, empty input.
  - `hung_server_after_connect_times_out_cleanly` — mock `TcpListener`
    accepts and `sleep_forever`s; asserts `Err("…timed out…")` inside
    the test budget. Pre-fix this hangs the whole test process.
  - `oversized_response_is_size_capped` — listener streams 16 KiB
    chunks past a 64 KiB cap; asserts `Err("…exceeded…")` before OOM.
  - `chunked_encoding_is_explicitly_rejected` — listener returns
    `Transfer-Encoding: chunked`; asserts the explicit
    `Err("…chunked encoding…")` instead of silent body corruption.
  - `geocode_request_line_contains_percent_encoded_location` — listener
    captures the raw HTTP request line and echoes it back as JSON;
    asserts the line contains `name=Q%26A` and does NOT contain
    `name=Q&A`. Locks the URL-encoding fix at the wire level.

No config schema change, no public API change. ASCII / space-only
locations against a healthy Open-Meteo see byte-identical behaviour.

## Real Behavior Proof

- [x] NOT verified on Jetson hardware. The change is in pure tokio I/O
  scheduling and string formatting — no audio, voice, ALSA, CUDA, HA,
  LLM-backend, or systemd surface — exercised end-to-end by 5 new
  `#[tokio::test]` regressions that spin up real local `TcpListener`s
  and drive each failure mode.

### What I ran

Environment: x86_64 Linux dev host (Ubuntu 22.04, Rust 1.95.0). No
Jetson available.

```bash
cargo fmt --all -- --check                                                # clean
cargo clippy --workspace --all-targets -- -D warnings                    # clean
cargo clippy --workspace --all-targets --no-default-features -- -D warnings  # clean
cargo test -p genie-core --lib tools::weather                             # 6 / 0
cargo test                                                                # 675 / 0 / 3
cargo test --workspace --no-default-features                             # 570 / 0
```

### What I observed

1. **Hung Open-Meteo is bounded.** With a `TcpListener` that accepts
   then `sleep_forever`s, `http_get_with_limits(…, 500ms, 500ms, …)`
   returns `Err("Open-Meteo GET … timed out after 0s")` in well under
   a second. Pre-fix the same call hangs for 600 s.
2. **Oversized response is rejected before OOM.** Listener streams
   16 KiB chunks against a 64 KiB cap; the helper bails with
   `"Open-Meteo response exceeded 65536 bytes (got at least …)"`.
3. **Chunked encoding gets an explicit error.** Listener returns
   `Transfer-Encoding: chunked` with a real chunk; helper bails with
   `"Open-Meteo response uses unsupported chunked encoding"`. Pre-fix
   this silently corrupted multi-chunk JSON.
4. **URL encoding is wire-correct.** Listener captures the raw HTTP
   request line via `read` and echoes it back as JSON. For
   `geocode("Q&A Cafe Tokyo")` the captured line is
   `GET /v1/search?name=Q%26A%20Cafe%20Tokyo&count=1&language=en&format=json HTTP/1.1`.
   Pre-fix this would have been `name=Q&A+Cafe+Tokyo&…`, which
   Open-Meteo parses as `name=Q`.
5. **No regression on the happy path.** `wmo_codes` (existing test)
   still passes. Full `cargo test`: 675 / 0 / 3 (up from main baseline
   ~670 by exactly the 5 new tests). `--no-default-features`: 570 / 0.

## Test plan

A reviewer can re-verify on any Rust 1.85+ host (no Jetson, no Open-
Meteo, no audio needed):

- `cargo test -p genie-core --lib tools::weather` — 6 tests (1 existing
  + 5 new), all green in <1 s.
- Optional manual proof against a real Open-Meteo:
  1. Start `genie-core` with default configuration.
  2. `curl -X POST http://127.0.0.1:3000/api/chat \
        -d '{"message":"weather in Q&A Cafe Tokyo"}'`.
     Pre-fix: returns "location not found". Post-fix: returns a
     geocoded weather report for the closest matching location.
  3. Optionally `tcpdump -i any 'port 80'` to see
     `GET /v1/search?name=Q%26A%20Cafe%20Tokyo&…` on the wire.

## Notes for reviewers

- **No merge-conflict surface with any open PR.** The file
  `crates/genie-core/src/tools/weather.rs` has not been touched since
  the workspace scaffold commit `0a7363b`. No open PR touches it.
- **Why three concerns in one PR.** They live in the same function
  family (`http_get` + the three callers) and the test scaffolding
  (mock `TcpListener` + raw HTTP framing) is shared. Splitting them
  would multiply review surface for the same total diff.
- **Why reject chunked instead of fix the decoder.** Open-Meteo's free
  tier returns `Content-Length`, never chunked. The pre-fix decoder
  silently corrupted multi-chunk JSON — refusing chunked explicitly is
  safer than fixing a decoder for a path that should be unreachable.
  A proper chunked decoder can land later if Open-Meteo ever changes.
- **Defaults.** 15 s request timeout is comfortably above Open-Meteo's
  typical 200 ms response. 1 MiB is ~50x the largest realistic forecast
  response. Both are workspace-private constants; making them operator-
  configurable can be a follow-up if needed.
- **Related work.** PR #174 (mine, merged) fixed the same unbounded-
  read pattern in `ha/client`. PR #182 (open) fixes the LLM client.
  This PR is the third leg — once it lands the "raw-TCP HTTP client
  with insufficient read timeouts" class is closed across the
  workspace.

### DIFF
--- a/crates/genie-core/src/tools/weather.rs
+++ b/crates/genie-core/src/tools/weather.rs
@@ -1,3 +1,5 @@
+use std::time::Duration;
+
 use anyhow::Result;
 use tokio::io::{AsyncBufReadExt, AsyncWriteExt, BufReader};
 use tokio::net::TcpStream;
@@ -10,6 +12,21 @@ use tokio::net::TcpStream;
 ///
 /// All requests go through raw TCP+TLS-free HTTP to api.open-meteo.com.
 /// Note: Open-Meteo supports HTTP (no TLS required for the free tier).
+
+/// Connect-timeout cap for Open-Meteo HTTP requests.
+const WEATHER_CONNECT_TIMEOUT: Duration = Duration::from_secs(10);
+
+/// Total request-lifecycle timeout (covers write + read). Without this, a
+/// slow or hung Open-Meteo response leaves the calling chat task wedged
+/// forever. Same fix shape as PR #174 / closes #173 for `ha::client`.
+const WEATHER_REQUEST_TIMEOUT: Duration = Duration::from_secs(15);
+
+/// Body-size cap on the accumulated response. The Open-Meteo free tier
+/// returns < 4 KiB for geocoding and < 20 KiB for forecasts in practice;
+/// 1 MiB leaves a healthy margin while still preventing RSS growth from a
+/// misbehaving or man-in-the-middle response (the connection is plain
+/// HTTP). Without this the body-read loop would accumulate unboundedly.
+const WEATHER_MAX_RESPONSE_BYTES: usize = 1024 * 1024;
 
 // ── Public API ──────────────────────────────────────────────
 
@@ -65,7 +82,13 @@ struct ForecastDay {
 
 /// Geocode a location name using Open-Meteo's geocoding API.
 async fn geocode(location: &str) -> Result<(f64, f64, String)> {
-    let encoded = location.replace(' ', "+");
+    // Previously `location.replace(' ', "+")` — that only handled spaces,
+    // so reserved URL characters in the location (e.g. `&`, `?`, `#`, `=`,
+    // `%`, `+`) leaked into the query string and silently broke geocoding.
+    // A user asking the weather in "Q&A Cafe Tokyo" used to produce
+    // `name=Q&A+Cafe+Tokyo&count=1…` — Open-Meteo parsed the `&` as a
+    // separator and saw `name=Q`. Percent-encode every reserved char now.
+    let encoded = url_encode_query_param(location);
     let path = format!(
         "/v1/search?name={}&count=1&language=en&format=json",
         encoded
@@ -179,14 +202,36 @@ async fn fetch_forecast(lat: f64, lon: f64) -> Result<Vec<ForecastDay>> {
 }
 
 /// Raw HTTP GET (no TLS — Open-Meteo supports plain HTTP).
+///
+/// Production callers reach this via the host-only convenience signature;
+/// it defaults to port 80 and the workspace-wide timeout/size constants.
 async fn http_get(host: &str, path: &str) -> Result<String> {
-    let addr = format!("{}:80", host);
-    let stream = tokio::time::timeout(
-        std::time::Duration::from_secs(10),
-        TcpStream::connect(&addr),
+    http_get_with_limits(
+        host,
+        80,
+        path,
+        WEATHER_CONNECT_TIMEOUT,
+        WEATHER_REQUEST_TIMEOUT,
+        WEATHER_MAX_RESPONSE_BYTES,
     )
     .await
-    .map_err(|_| anyhow::anyhow!("connection timeout"))??;
+}
+
+/// Inner implementation that takes explicit limits — exposed `pub(crate)`
+/// only so the test module can point at an ephemeral mock listener with
+/// millisecond-scale timeouts. NOT part of any stable API.
+pub(crate) async fn http_get_with_limits(
+    host: &str,
+    port: u16,
+    path: &str,
+    connect_timeout: Duration,
+    request_timeout: Duration,
+    max_bytes: usize,
+) -> Result<String> {
+    let addr = format!("{}:{}", host, port);
+    let stream = tokio::time::timeout(connect_timeout, TcpStream::connect(&addr))
+        .await
+        .map_err(|_| anyhow::anyhow!("Open-Meteo connect to {} timed out", addr))??;
 
     let (reader, mut writer) = stream.into_split();
 
@@ -194,11 +239,36 @@ async fn http_get(host: &str, path: &str) -> Result<String> {
         "GET {} HTTP/1.1\r\nHost: {}\r\nUser-Agent: GeniePod/0.2\r\nAccept: application/json\r\nConnection: close\r\n\r\n",
         path, host
     );
-    writer.write_all(request.as_bytes()).await?;
 
+    // Run the entire write + read cycle under a single deadline. Without
+    // this a hung Open-Meteo wedges the chat task forever even after the
+    // TCP connect succeeds. Same shape as PR #174 / closes #173.
+    let body = tokio::time::timeout(request_timeout, async move {
+        writer.write_all(request.as_bytes()).await?;
+        read_http_get_body(reader, max_bytes).await
+    })
+    .await
+    .map_err(|_| {
+        anyhow::anyhow!(
+            "Open-Meteo GET {} timed out after {}s",
+            path,
+            request_timeout.as_secs()
+        )
+    })??;
+
+    Ok(body.trim().to_string())
+}
+
+/// Drain headers and body from a one-shot `Connection: close` GET response,
+/// rejecting chunked encoding (Open-Meteo's free tier doesn't use it) and
+/// bounding accumulated body bytes at `max_bytes`.
+async fn read_http_get_body(
+    reader: tokio::net::tcp::OwnedReadHalf,
+    max_bytes: usize,
+) -> Result<String> {
     let mut buf_reader = BufReader::new(reader);
-    let mut body = String::new();
     let mut in_body = false;
+    let mut body = String::new();
 
     loop {
         let mut line = String::new();
@@ -206,25 +276,72 @@ async fn http_get(host: &str, path: &str) -> Result<String> {
         if n == 0 {
             break;
         }
-        if in_body {
-            body.push_str(&line);
-        } else if line.trim().is_empty() {
-            in_body = true;
-        }
-    }
 
-    // Handle chunked transfer encoding (simple: just strip chunk headers).
-    if body.contains("\r\n") && body.starts_with(|c: char| c.is_ascii_hexdigit()) {
-        let mut decoded = String::new();
-        for line in body.split("\r\n") {
-            if !line.is_empty() && !line.chars().all(|c| c.is_ascii_hexdigit()) {
-                decoded.push_str(line);
+        if !in_body {
+            // The previous implementation tried to be clever about
+            // `Transfer-Encoding: chunked` after the fact (lines 216-225
+            // of the pre-fix code) and joined non-hex-only lines with
+            // `push_str` — no separator between adjacent chunks, so JSON
+            // spanning multiple chunks was silently corrupted, and the
+            // heuristic fired on legitimate responses whose first body
+            // character was a hex digit. Refuse chunked explicitly
+            // instead: Open-Meteo's free tier returns Content-Length, so
+            // this path should be unreachable in practice; if a proxy or
+            // future Open-Meteo upgrade turns it on, the user gets a
+            // clear error and the call site can retry/fallback.
+            let lower = line.to_ascii_lowercase();
+            if let Some(value) = lower.strip_prefix("transfer-encoding:")
+                && value.split(',').any(|tok| tok.trim() == "chunked")
+            {
+                anyhow::bail!("Open-Meteo response uses unsupported chunked encoding");
             }
+
+            if line.trim().is_empty() {
+                in_body = true;
+            }
+            continue;
         }
-        body = decoded;
+
+        // Body branch.
+        let projected = body.len().saturating_add(line.len());
+        if projected > max_bytes {
+            anyhow::bail!(
+                "Open-Meteo response exceeded {} bytes (got at least {})",
+                max_bytes,
+                projected
+            );
+        }
+        body.push_str(&line);
     }
 
-    Ok(body.trim().to_string())
+    Ok(body)
+}
+
+/// Percent-encode `s` for safe use as a single query-string parameter
+/// value. Encodes every byte that is NOT an unreserved RFC 3986 character
+/// (`A-Z`, `a-z`, `0-9`, `-`, `_`, `.`, `~`). Multi-byte UTF-8 codepoints
+/// are encoded byte-by-byte, which is what every HTTP server (including
+/// Open-Meteo's geocoder) expects. The result is always pure ASCII.
+fn url_encode_query_param(s: &str) -> String {
+    let mut out = String::with_capacity(s.len());
+    for byte in s.as_bytes() {
+        if matches!(byte, b'A'..=b'Z' | b'a'..=b'z' | b'0'..=b'9' | b'-' | b'_' | b'.' | b'~') {
+            out.push(*byte as char);
+        } else {
+            out.push('%');
+            out.push(hex_nibble(byte >> 4));
+            out.push(hex_nibble(byte & 0x0F));
+        }
+    }
+    out
+}
+
+fn hex_nibble(n: u8) -> char {
+    match n {
+        0..=9 => (b'0' + n) as char,
+        10..=15 => (b'A' + (n - 10)) as char,
+        _ => '0', // unreachable for caller using `b & 0x0F`
+    }
 }
 
 /// WMO weather interpretation codes → human-readable description.
@@ -253,6 +370,8 @@ fn wmo_code_to_description(code: u64) -> String {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use tokio::io::AsyncWriteExt;
+    use tokio::net::TcpListener;
 
     #[test]
     fn wmo_codes() {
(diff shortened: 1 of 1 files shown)


ai-hpc:
reviewd, bounded weather HTTP reads and encoded geocode queries and merged at 15c2c95a2fe4
Thanks @galuis116

PR #218, opened 2026-05-26
fix(memory): log recall tracking update failures (closes #216)

### LINKED ISSUE #216: [bug] genie-core: memory recall tracking updates silently discarded on DB write failure
### Summary

After memory search returns results, recall analytics updates are silently dropped:

```rust
// memory/mod.rs ~L417–420, L472–475
for (entry, score) in &scored {
    let _ = self.update_recall_tracking(entry.id, now, *score, &query_hash);
}
```

`update_recall_tracking` runs `UPDATE memories SET recall_count = recall_count + 1, ...`. If the DB is readonly, full, or locked, search still returns hits but `recall_count`, `max_score`, and `query_hashes` never update — with no operator signal.

Same observability gap pattern as fixed #159 (health_log) and #193 (open-time migration).

### Steps to reproduce

1. Run `genie-core` with a populated `memory.db`.
2. Make the DB or parent directory read-only (or simulate a locked connection).
3. Trigger memory recall via chat or `memory_recall` tool.
4. Search results still return; recall counters in DB stay stale; no error log.

### Expected behavior

Log `tracing::error!` (or `warn!`) when `update_recall_tracking` fails, mirroring #159 / #193 patterns.

### Actual behavior

Errors swallowed via `let _ =`; promotion scoring inputs degrade silently over time.

### Hardware

Non-Jetson (x86_64 dev) sufficient

### JetPack / L4T version

_No response_

`main` @ `crates/genie-core/src/memory/mod.rs` (`search_fts`, `search_like_fallback`)

### Relevant logs

```
N/A — failure is silent, which is the bug.
```

### Additional context

**PR scope:** Small — log on failure + optional readonly-DB unit test (~20–40 lines).

**Duplicate check**

| Item | Relation |
|------|----------|
| **#193** (closed, merged) | Migration/FTS failures at `Memory::open()` — different code path |
| **#212** (open) | Double open at startup — different bug |
| **#111** (closed) | Recall test-set milestone — not this write-path observability gap |
| No open issue | `let _ = update_recall_tracking(...)` in search paths |

kiannidev:
## Summary

- Log `tracing::error!` when `update_recall_tracking` fails in FTS and LIKE fallback search paths.
- Search results still return to the caller; operators get a signal when recall analytics stop updating.

Closes #216

## Real Behavior Proof

### What I ran

```bash
cargo clippy -p genie-core --all-targets --locked -- -D warnings
cargo fmt --all -- --check
```

### What I observed

All 482 genie-core lib tests passed.

## Test plan

- [x] `cargo test -p genie-core --lib`
- [x] `cargo clippy -p genie-core --all-targets --locked -- -D warnings`
- [x] `cargo fmt --all -- --check`

### DIFF
--- a/crates/genie-core/src/memory/mod.rs
+++ b/crates/genie-core/src/memory/mod.rs
@@ -9,6 +9,7 @@ use rusqlite::{Connection, OpenFlags, params_from_iter};
 use serde::Serialize;
 use std::path::Path;
 use std::path::PathBuf;
+use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
 const MAX_QUERY_HASHES: usize = 16;
@@ -38,6 +39,17 @@ pub struct Memory {
     canonical_dir: PathBuf,
     /// Set when schema migration or FTS rebuild failed during [`Memory::open`].
     migration_degraded: bool,
+}
+
+/// Process-wide handle to the single memory store opened at startup.
+pub type SharedMemory = Arc<Mutex<Memory>>;
+
+/// Run a closure against the shared memory store.
+pub fn with_shared_memory<R>(memory: &SharedMemory, f: impl FnOnce(&Memory) -> R) -> R {
+    let guard = memory
+        .lock()
+        .unwrap_or_else(|poisoned| poisoned.into_inner());
+    f(&guard)
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


ai-hpc:
reviewd, logged memory recall tracking write failures and merged at 405dbacac618
Thanks @kiannidev

PR #282, opened 2026-05-28
fix(skills/loader): read .so once and load via sealed memfd to close …

### LINKED ISSUE #280: [bug] skills: TOCTOU race in verify_skill_signature / dlopen lets a file-swap replace a verified .so with a malicious binary
### Summary

`load_skill` reads the `.so` bytes **twice** from disk — once in `verify_skill_signature` (Ed25519 check, `loader.rs:336`) and a second time via `Library::new` / `dlopen` (`loader.rs:484`). An attacker with write access to the skills directory can drop a legitimately signed `.so`, wait for verification to complete, then replace the file before `dlopen` executes. `dlopen` loads the malicious binary with `signed = true` already recorded, negating the entire `require_signature` security boundary and executing arbitrary native code inside the `genie-core` daemon.

---

### Steps to reproduce

> These steps demonstrate the race mechanically on a dev host. The window is narrow but reliably exploitable under CPU load or with a `sched_setscheduler`-pinned attacker process.

1. Enable `require_signature = true` in the skills config.
2. Build and sign a benign skill `.so` (`legit.so`) with a trusted key.
3. Drop `legit.so` into the skills directory.
4. In a tight loop on a second thread/process, `inotifywait` for `IN_CLOSE_NOWRITE` on `legit.so` (signals that `verify_skill_signature` finished its `fs::read`), then immediately `rename malicious.so legit.so`.
5. Trigger `load_skill("legit.so")` from the main genie-core process simultaneously.
6. Observe `Library::new` loading `malicious.so` content while `signed = true` is already recorded by the verification step.

Relevant code locations:

```
// loader.rs:336 — READ #1 (verification)
fn verify_skill_signature(skill_path: &Path, ...) -> bool {
    match std::fs::read(skill_path) { ... }   // reads bytes into memory, verifies Ed25519
}

// loader.rs:466 — check passes
let signed = verify_skill_signature(path, manifest, &self.trusted_keys);

// loader.rs:484 — READ #2 (execution) — TOCTOU WINDOW
let lib = unsafe { Library::new(path) }?;   // dlopen reads from disk again
```

---

### Expected behavior

If `require_signature` is enabled, only the **exact bytes that were cryptographically verified** should be loaded into the process. A file swap between verification and load must be impossible by construction — the loaded binary must be the verified binary.

---

### Actual behavior

`dlopen` reads the `.so` from disk independently of `verify_skill_signature`. Any file placed at `path` between `loader.rs:466` and `loader.rs:484` is loaded unconditionally. There is no fd-sealing, no file descriptor reuse, and no post-load hash comparison. The `signed` flag is meaningless as a security guarantee.

---

## Structural changes required

This is not fixable by adding a post-load check (`dlopen` then `fs::read` + re-verify) — that creates a second TOCTOU window in the opposite direction. The fix requires restructuring the entire load-and-verify sequence into a single atomic operation:

### Option A — `memfd_create` + fd sealing (Linux, recommended)

1. **New helper `load_verified_into_memfd`** in `loader.rs`:
   - `fs::read(skill_path)` — single read into memory.
   - `verify_detached(bytes)` — fail fast if invalid.
   - `memfd_create(MFD_ALLOW_SEALING)` — anonymous in-memory file.
   - `write_all(bytes)` into the memfd.
   - `fcntl(MFD_SEAL_WRITE | MFD_SEAL_SHRINK | MFD_SEAL_GROW | MFD_SEAL_SEAL)` — immutable after sealing.
   - Return the sealed `RawFd`.
2. **New unsafe loader path**: replace `Library::new(path)` with `Library::new(format!("/proc/self/fd/{fd}"))` against the sealed memfd.
3. **`SkillLoader::load_skill` restructure**: the current two-step `verify_skill_signature` → `Library::new(path)` sequence is replaced by a single call to `load_verified_into_memfd` that either returns a loaded `Library` or an error — there is no intermediate state where verification passed but the `.so` source is still the on-disk file.
4. **`verify_skill_signature` removal or relegation**: the standalone function becomes internal to the new helper; its public call site in `load_skill` is deleted.
5. **Feature-gate**: wrap `memfd_create` usage behind `#[cfg(target_os = "linux")]`; surface a compile-time error or fallback behaviour for non-Linux targets.
6. **New `LoadSkillError` variant**: `SignatureRacePrevented` (or similar) for cases where sealing fails, to distinguish crypto failures from fd-sealing failures in logs.

### Option B — hard-link to temp path (fallback, fragile)

Hard-link the `.so` to a private temp path, `chmod 000`, verify the temp path, then pass it to `Library::new`. Still racy on filesystems without `O_TMPFILE` support and does not guarantee immutability. Not recommended; included here only as a non-Linux fallback if `memfd_create` is unavailable.

### Files that must change

| File | Change |
|------|--------|
| `crates/genie-core/src/skills/loader.rs` | Replace two-step verify+dlopen with `load_verified_into_memfd`; restructure `load_skill`; remove/internalize `verify_skill_signature` |
| `crates/genie-core/src/skills/error.rs` (or equivalent) | Add new error variant for sealing failure |
| `crates/genie-core/Cargo.toml` | Add `nix` or `libc` dependency for `memfd_create` / `fcntl` syscall wrappers |
| `crates/genie-core/src/skills/loader_tests.rs` (or test module) | Add regression test: verify that a file replaced on disk between the bytes read and the Library load does **not** affect what is actually loaded |

---

### Hardware

### JetPack / L4T version

N/A 

b0d2a47

### Relevant logs

```shell
# No log evidence — the race is silent. The malicious .so executes as a
# legitimate loaded skill. No error is emitted by SkillLoader.
# Audit trail shows: signed=true, path=legit.so, loaded=ok.


---
```

### Additional context

- `require_signature` is documented as the security boundary preventing untrusted native code from running inside the daemon. This race negates that guarantee entirely for any attacker with write access to the skills directory (local user, compromised service with directory write permission, or a symlink attack if the skills dir is user-writable).
- Issue #175 (closed) addressed the `require_signature` logic and verification tests but does not address this filesystem race — it assumes `verify_skill_signature` and `Library::new` operate on the same bytes.
- Issue #259 (open) addresses `panic = "abort"` voiding the `catch_unwind` fault barrier — a related but orthogonal skill-loading safety concern.
- `memfd_create` + sealing is the standard Linux mitigation for this class of TOCTOU. glibc ≥ 2.27 and Linux ≥ 3.17 (well within JetPack 6.x L4T R36) support it. The `nix` crate provides safe wrappers.
- Severity: **Critical** — arbitrary native code execution inside `genie-core` at daemon privilege level.

enjoyandlove:
## Summary

Close the TOCTOU race in the skill loader where the `.so` was read once for
signature verification and then re-read from disk by `dlopen`, leaving a window
for an attacker to swap the file in between. Fixes #280.

## Changes

- Refactor `verify_skill_signature` to accept `&[u8]` instead of `&Path`,
  removing the internal `fs::read` call so the caller owns the single read.
- In `load_skill`, read the `.so` bytes exactly once before verification; pass
  the same buffer to both `verify_skill_signature` and the library loader.
- Add `load_library_from_memfd` (Linux only): writes the verified bytes into an
  anonymous `memfd`, seals it with `F_SEAL_WRITE | F_SEAL_SHRINK |
  F_SEAL_GROW | F_SEAL_SEAL`, then calls `dlopen` via `/proc/self/fd/<n>`.
  The kernel loads exactly the bytes that were cryptographically verified.
- Non-Linux platforms fall back to the previous path-based `dlopen` with an
  explicit comment noting there is no TOCTOU mitigation on those targets.
- Add two Linux-only regression tests:
  - `memfd_loads_library_from_bytes_not_path` — unit test that `load_library_from_memfd`
    resolves the `genie_skill_init` symbol from the supplied bytes.
  - `loader_toctou_disk_swap_after_load_has_no_effect` — end-to-end test that
    overwrites the `.so` on disk after load and confirms the sealed in-memory
    copy executes correctly with legitimate output.

## Real Behavior Proof

- [ ] `jetson`
- [ ] `raspberry_pi`
- [ ] `portable_sbc`
- [x] `laptop`
- [ ] `mac`
- [ ] CI-only / docs-only
- [ ] Not run locally

### What I ran

```
cargo test -p genie-core --locked -- skills::tests::memfd_loads_library_from_bytes_not_path skills::tests::loader_toctou_disk_swap_after_load_has_no_effect
cargo clippy -p genie-core --locked -- -D warnings
```

Linux x86-64 laptop, Ubuntu 24.04, kernel 6.17.0-29-generic (memfd and
`/proc/self/fd` are available on this kernel). The TOCTOU tests require Linux
and are gated with `#[cfg(target_os = "linux")]`.

Jetson verification gap: the `aarch64-unknown-linux-gnu` cross-compile CI job
builds the crate; the memfd syscall and `/proc/self/fd` paths are present on
all L4T kernels (4.9+), so the code path is the same. No runtime voice or Home
Assistant interaction is affected by this change.

### What I observed

Both new tests pass. `loader_toctou_disk_swap_after_load_has_no_effect`
confirmed that after overwriting the `.so` on disk with `b"attacker payload
swapped after verification"`, the already-loaded skill returned the expected
`"hello_world"` output and `success == true`.

## Test plan

1. `cargo test -p genie-core --locked -- skills::tests` — all skill loader tests
   must pass, including the two new TOCTOU tests (Linux only).
2. `cargo clippy -p genie-core --locked -- -D warnings` — no new warnings.
3. (Optional, Linux) Manually confirm sealing: attach `strace -e fcntl` to a
   test run and verify `F_ADD_SEALS` is called with the full bitmask before
   the `openat("/proc/self/fd/…")` that backs the dlopen.

## Notes for reviewers

- The non-Linux fallback at `#[cfg(not(target_os = "linux"))]` still uses
  path-based `dlopen`. This is intentional — `require_signature` is a
  Linux-first production boundary (Jetson / Pi / SBC), and the comment in the
  code calls this out explicitly.
- `so_bytes` is dropped immediately after `load_library_from_memfd` returns,
  before vtable extraction, to free the heap allocation as early as possible.
  Skills on Jetson can be several MB; holding the buffer through vtable
  extraction is unnecessary.
- `memfd_create` labels are best-effort debug names; a null byte in the
  filename falls back to an empty label via `unwrap_or_default()`, which is
  safe.


ai-hpc:
Closing this PR as not qualified under the duplicate-open-PR rule.

Reason: this author currently has multiple open PRs. Keep one focused PR open at a time; close or finish existing work before opening more. I am leaving the newest non-AI PR from this author open for review and closing older/surplus PRs.

Please consolidate and reopen only after the active PR is resolved.

ai-hpc (dismissed):
Code review approved. The sealed memfd path is the right direction for closing the signed-skill verify/dlopen TOCTOU window on Linux/Jetson, and the PR has focused regression coverage plus green required CI/aarch64 checks. It is still behind main, so update the branch and rerun checks before merge.

ai-hpc (approved):
M1-relevant native skill-loader hardening. Fixed the async regression test on the PR branch and required CI/aarch64 checks are green.

ai-hpc:
Merged at `4c70bb7`. Thanks @enjoyandlove.\n\nMaintainer note: future unplanned PRs outside the active M1 focus will be closed next time.

PR #325, opened 2026-05-29
ci: limit contributors to two open PRs

ai-hpc:
## Summary

Lower the per-author open PR limit from 3 to 2.

This keeps review load bounded and forces contributors to keep only their highest-signal work open.

## Real Behavior Proof

Workflow-only change. It does not affect Jetson runtime behavior.

Tested profile / hardware:

- [ ] jetson
- [ ] raspberry_pi
- [ ] portable_sbc
- [ ] laptop
- [ ] mac
- [x] CI-only / docs-only
- [ ] Not run locally

### What I ran

Reviewed the workflow diff locally.

### What I observed

The PR limit workflow will now close when an author has more than 2 open PRs.

### DIFF
--- a/.github/workflows/pr-limit.yml
+++ b/.github/workflows/pr-limit.yml
@@ -21,7 +21,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 2
     env:
-      MAX_OPEN_PRS: "3"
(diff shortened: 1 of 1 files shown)


ai-hpc:
Closed per maintainer direction: keep the per-author open PR limit at 3, not 2.

PR #326, opened 2026-05-29
fix(calc): parse unary minus as a grammar rule, not in the lexer

galuis116:
## Summary

Fix unary minus so it negates any factor (including parenthesized groups), not
just bare literals. `-(2 + 3)` and friends previously returned
`Err("unexpected token: Minus")`. Split out of #312 to keep that PR focused on
memory recall.

## Changes

- Move unary `+`/`-` handling out of the tokenizer and into a new `parse_unary`
  grammar rule sitting between `parse_term` and `parse_factor`.
- The lexer is now context-free: `-` always emits a `Token::Minus` and never
  glues a sign onto a number literal.
- `parse_unary` folds a run of leading signs by parity *iteratively*, so a
  hostile `-----…-1` chain can't grow the call stack — only `(` recurses, which
  `MAX_PAREN_DEPTH` already caps. Updated the depth-cap doc comment to reflect
  the extra frame per `(`.
- Added regression tests for parenthesized negation, term-position signs,
  chained signs, unchanged signed-literal cases, and a 5000-sign no-overflow
  test on a 2 MiB stack.

## Real Behavior Proof

- [ ] `jetson`
- [ ] `raspberry_pi`
- [ ] `portable_sbc`
- [x] `laptop`
- [ ] `mac`
- [ ] CI-only / docs-only
- [ ] Not run locally

### What I ran

Ubuntu 22.04, x86_64 (laptop profile):

cargo test -p genie-core --lib tools::calc

### What I observed

running 17 tests
test result: ok. 17 passed; 0 failed; 0 ignored; 0 measured; 628 filtered out

Including the new cases: `unary_minus_negates_parenthesized_group`,
`unary_minus_in_term_position`, `chained_unary_signs`,
`existing_negative_number_cases_unchanged`, and `long_unary_run_does_not_overflow`.

This change is pure arithmetic-parser logic with no voice/audio/Home Assistant/
dashboard surface, so the Jetson gap carries no runtime-specific risk — behavior
is identical across profiles. The stack-overflow safety test pins a 2 MiB stack
to match Tokio's default worker thread.

## Test plan

- `cargo test -p genie-core --lib tools::calc`
- Manually: evaluate `-(2 + 3)`, `3 * -(2)`, `--3`, `-+-3` — all should succeed.

## Notes for reviewers

- Split out of #312 (memory recall) per the request to keep PRs to one behavior
  surface.
- No behavior change to the existing signed-literal path; only previously-failing
  `-(...)` expressions now evaluate.

### DIFF
--- a/crates/genie-core/src/tools/calc.rs
+++ b/crates/genie-core/src/tools/calc.rs
@@ -16,11 +16,11 @@ pub fn evaluate(expr: &str) -> Result<f64, String> {
 
 /// Maximum nested parenthesis depth the parser will accept.
 ///
-/// The parser is recursive-descent: each `(` adds three stack frames
-/// (`parse_expr` → `parse_term` → `parse_factor`) before the recursion
-/// deepens again. Tokio's default worker thread has a 2 MiB stack, which
-/// empirically aborts somewhere around ~1500-3000 nested parens. 64 levels
-/// leaves a roughly 30x safety margin for the worst-case frame size while
+/// The parser is recursive-descent: each `(` adds four stack frames
+/// (`parse_expr` → `parse_term` → `parse_unary` → `parse_factor`) before the
+/// recursion deepens again. Tokio's default worker thread has a 2 MiB stack,
+/// which empirically aborts somewhere around ~1500-3000 nested parens. 64
+/// levels leaves a large safety margin for the worst-case frame size while
 /// still admitting any realistic arithmetic expression. The check exists so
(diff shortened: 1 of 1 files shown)


ai-hpc:
reviewd and merged at 2526607a314e1c994392de4616118b6bcceaba05
Thanks @galuis116

Note: this was a one-time pass over `not-planned` PRs. Going forward, only milestone-scoped PRs will be merged. If you have questions or want to collaborate more closely, join our Discord.

PR #348, opened 2026-05-31
fix(security): enforce Landlock filesystem sandbox at genie-core startup

### LINKED ISSUE #347: [bug] security: Landlock filesystem sandbox is never called and its implementation is hollow — genie-core runs fully unsandboxed
### Summary

`apply_landlock()` is never invoked from `main.rs` **and** the implementation itself makes no Landlock syscalls — genie-core runs with unrestricted filesystem access despite the documentation, code comments, and module docstring all claiming Landlock is active.

### Steps to reproduce

1. Build and run `genie-core` on any Linux host (Jetson or x86_64 dev).
2. Confirm the process starts and logs appear normal.
3. From within the same user session (or via a compromised tool path), access an arbitrary path that Landlock is supposed to block:
   ```sh
   # While genie-core is running, from the same UID:
   cat /etc/shadow          # Should fail — Landlock should block reads outside allowed paths
   ls /home/other_user/     # Should fail — Landlock should restrict to /etc/geniepod + data_dir
   ```
4. Both succeed. No Landlock restriction is in effect.

Alternatively, inspect startup logs at `INFO` level — you will NOT see "Landlock filesystem sandbox: preparing rules" even on a kernel that supports Landlock (5.13+), because `apply_landlock()` is never called.

### Expected behavior

- `main.rs` calls `apply_landlock(config_dir, data_dir)` early in startup, before any network or file I/O.
- On kernels ≥ 5.13 (JetPack 5.x / L4T R35+, JetPack 6.x / L4T R36+), the Landlock ruleset is created, path rules are added, and `landlock_restrict_self()` is called — making the restriction irreversible.
- After that point, any attempt to open a path outside the allowed set (`/etc/geniepod`, `data_dir`, `/proc`, `/sys`, `/dev`, `/usr/lib`, `/opt/geniepod/bin`) returns `EACCES`.
- `/api/health` reports `"sandbox_enforced": true` when restriction is active, `false` with an explanation when it is not (kernel too old, enforcement disabled in config).
- On kernels without Landlock, startup logs a `WARN` (not a silent `Ok(())`), and optionally the operator can set `[sandbox] require_landlock = true` to make this a hard startup failure.

### Actual behavior

**Layer 1 — `apply_landlock` is never called.**
`crates/genie-core/src/main.rs` contains no reference to `apply_landlock` or `security::sandbox::apply_landlock`. The function is exercised only by one unit test (`landlock_doesnt_crash_on_any_platform`) which asserts `result.is_ok()` — validating only that the stub doesn't panic, not that it restricts anything.

**Layer 2 — The implementation is a stub that does nothing.**
`crates/genie-core/src/security/sandbox.rs:79-99`:

```rust
// TODO: Implement full Landlock ruleset when testing on Jetson.
// The syscall interface is:
//   1. landlock_create_ruleset() — create with fs access flags
//   2. landlock_add_rule() — add path rules (read/write/exec per dir)
//   3. landlock_restrict_self() — apply (irreversible)
//
// For now, document the rules and validate on hardware.

tracing::info!(
    "Landlock rules prepared (not yet enforced — requires Jetson kernel validation): \
     READ=[/etc/geniepod, /proc, /sys, /dev, /usr/lib], \
     WRITE=[{}], \
     EXEC=[/opt/geniepod/bin]",
    data_dir.display()
);

Ok(())
```

The only Landlock syscall that runs is `landlock_create_ruleset` with the `LANDLOCK_CREATE_RULESET_VERSION` flag (line 59-66), which is the ABI probe — not ruleset creation. If that probe succeeds (abi ≥ 0), the stub logs "preparing rules" and returns `Ok(())`. No ruleset fd is opened. No `landlock_add_rule()` is called. No `landlock_restrict_self()` is called. The `Ok(())` return propagates to the (non-existent) caller as silent success.

**Net effect:** The module docstring, `ARCHITECTURE.md`, and the `// What's implemented:` header all describe Landlock as present. It is not. A compromised genie-core process, a malicious skill `.so`, or a prompt-injected tool call has full filesystem access under the running user.

### Hardware

### JetPack / L4T version

Reproducible on any kernel ≥ 5.13 (L4T R35+)

286cf2b

### Relevant logs

```shell
With `RUST_LOG=info`, startup produces no Landlock-related output at all (because `apply_landlock` is never called):


INFO  GeniePod core starting
INFO  LLM backend configured backend="..."
...


If `apply_landlock` were called manually, the stub would log:


INFO  Landlock filesystem sandbox: preparing rules abi_version=6 config_dir="/etc/geniepod" data_dir="/opt/geniepod/data"
INFO  Landlock rules prepared (not yet enforced — requires Jetson kernel validation): READ=[/etc/geniepod, /proc, /sys, /dev, /usr/lib], WRITE=[/opt/geniepod/data], EXEC=[/opt/geniepod/bin]


The phrase **"not yet enforced"** appears only in the `tracing::info!` string body — it is invisible to any log filter or monitoring tool watching for `WARN`/`ERROR`.
```

### Files that must change to fix this

This is a cross-crate fix. The minimum set of changes is:

| File | Change |
|------|--------|
| `crates/genie-core/src/security/sandbox.rs` | Full Landlock V1-V3 ABI: open ruleset fd, `landlock_add_rule()` per path, `landlock_restrict_self()`. Return a typed `SandboxReport { enforced: bool, abi_version: Option<i64> }` instead of `Result<(), String>`. |
| `crates/genie-core/src/security/landlock_rules.rs` | **New file.** Enumerate allowed paths per subsystem (config, data, voice devices, proc/sys/dev, skill dirs). Keep path lists in one place so adding a new data dir doesn't require hunting call sites. |
| `crates/genie-core/src/security/mod.rs` | Export `SandboxReport`; re-export `apply_landlock` with the new signature. |
| `crates/genie-core/src/main.rs` | Call `security::sandbox::apply_landlock(config_dir, &config.data_dir)` early in `main()`, before any file I/O or network setup. Log the returned `SandboxReport` at startup and thread it into the health state. |
| `crates/genie-common/src/config.rs` | Add `[sandbox]` section: `enforcement = "enforce" \| "warn" \| "off"` and `require_landlock = false`. Lets operators opt into a hard-failure if the kernel is too old. |
| `crates/genie-skill-sdk/src/sandbox.rs` | **New file.** Skill worker processes need a tighter ruleset (no access to config dir, only the skill's own data path). Apply before dlopen in loader. |
| `crates/genie-health/src/lib.rs` | Add `sandbox_enforced: bool` and `sandbox_abi_version: Option<u32>` to the health struct. |
| `crates/genie-api/src/routes/health.rs` | Surface `sandbox_enforced` in the `/api/health` JSON response so `genie-ctl status` and operators can see actual sandbox state. |
| `tests/integration/sandbox_landlock.rs` | **New file.** Integration test: after `apply_landlock()`, verify that `open("/etc/shadow", O_RDONLY)` returns `EACCES` and that `open(data_dir.join("test.db"), O_RDWR)` succeeds. Gate the test with `#[cfg_attr(..., ignore)]` when running on a kernel < 5.13. |
| `Cargo.toml` (workspace) | Add `rustix` (with `fs` + `process` features) or the `landlock` crate for safe Landlock bindings. Raw `libc::syscall` for Landlock is error-prone (ABI V1 vs V2 vs V3 have different struct layouts). |
| `deny.toml` | Allowlist the new dependency. |
| `ARCHITECTURE.md` | Correct the "What's implemented" section to accurately reflect actual vs. planned sandbox coverage. Until the fix lands, it should say "planned, not yet enforced." |

### Why this wasn't caught sooner

The unit test (`landlock_doesnt_crash_on_any_platform`) only asserts `result.is_ok()`. Because the stub always returns `Ok(())`, the test passes on every platform. There is no test that verifies an out-of-policy path is actually denied — only that the function doesn't panic. An integration test that opens `/etc/shadow` after calling `apply_landlock()` would have caught this immediately.

### Landlock ABI versioning note

Landlock V1 (kernel 5.13) supports only filesystem access flags. V2 (5.19) adds `LANDLOCK_ACCESS_FS_REFER`. V3 (6.2) adds truncate. The implementation should probe the ABI version (already done at line 59-66) and use the highest set of flags the kernel supports, falling back gracefully to V1. The `rustix` crate handles this cleanly.

### Skill worker gap

Even after `genie-core` itself is sandboxed, skill `.so` libraries execute inside the same process (via `spawn_blocking` on a thread in the same address space). Per-skill Landlock rules would require either:
- Forking a worker process before dlopen (highest isolation, higher latency), or
- Applying a tighter Landlock sub-ruleset on the thread before invoking the skill vtable (Landlock is per-thread-before-it-forks, not per-thread within a process).

This is tracked as a follow-up design question in `genie-skill-sdk/src/sandbox.rs` (new file). The fix for `genie-core` itself is independent and should ship first.

monsterdavidliu-ux:
## Summary

- Call `apply_landlock` during genie-core startup (before audit, network, and other file I/O) with configurable `[sandbox]` policy (`enforce`, `warn`, `off`, and optional `require_landlock`).
- Replace the Landlock stub with a real ruleset: path allowlists in `landlock_rules.rs`, `landlock_add_rule` via the `landlock` crate, and irreversible `restrict_self`.
- Report `sandbox_enforced`, `sandbox_abi_version`, and `sandbox_message` on `/api/health`, plus an integration test that denies reads of `/etc/shadow` when enforcement is active.

## Changes

- Wire `apply_landlock` in `genie-core` startup before other file I/O.
- Implement Landlock ruleset creation, path rules, and `restrict_self` in `security/sandbox.rs`.
- Add `landlock_rules.rs`, `[sandbox]` config, deploy TOML defaults, and health JSON fields.
- Add `tests/sandbox_landlock.rs` integration coverage.

## Real Behavior Proof

- [ ] `jetson`
- [ ] `raspberry_pi`
- [ ] `portable_sbc`
- [x] `laptop`
- [ ] `mac`
- [ ] CI-only / docs-only
- [ ] Not run locally

### What I ran

On x86_64 Linux (kernel with Landlock support):

```bash
cargo test -p genie-core
cargo build -p genie-core
```

### What I observed

`cargo test -p genie-core` completed successfully (unit tests plus `sandbox_landlock` integration test; the Landlock deny test skips when enforcement is unavailable). `cargo build -p genie-core` succeeded with the new `landlock` workspace dependency.

## Test plan

- [x] `cargo test -p genie-core`
- [ ] On a Linux host with Landlock (kernel 5.13+), start genie-core with `[sandbox] enforcement = "enforce"` and confirm `/api/health` shows `sandbox_enforced: true`
- [ ] Confirm `cat /etc/shadow` fails from the genie-core process after startup

Fixes #347


ai-hpc:
Thanks @monsterdavidliu-ux, and welcome — enforcing a real Landlock sandbox at startup is solid, substantial work. Closing as **not-planned on policy grounds, not a quality judgment**.

The active contribution gate is **M1 — Jetson 4096-token BFCL Agent Harness** (README "Active Contribution Gate" + `doc/milestone-1-portable-home-agent.md`): only PRs that move the M1 scorecard — BFCL tool-call accuracy, deterministic family/home memory, typed-tool reliability, compact prompt assembly, Jetson-native runtime — are merged for now. We just did a one-time fair pass over the not-planned PRs that predated this gate; an OS-level filesystem sandbox (plus a new `landlock` dependency and `[sandbox]` config/deploy surface) is outside M1.

It maps directly to **M3 — Home Runtime Boundary and Skill Safety**, which is just opening — that's the right home for it, and a resubmission there (validated on Jetson, since Landlock ABI/behavior is kernel-version-sensitive) would be very welcome. Questions or want to collaborate? Join our Discord.

PR #358, opened 2026-06-01
[Bug]: Chat export misparses id when URL has extra query parameters

### LINKED ISSUE #357: [Bug]: Chat export misparses id when URL has extra query parameters
# [Bug]: Chat export misparses id when URL has extra query parameters

## Description

`classify_route` extracts the export conversation id via `path.split("id=").nth(1)`, so any extra query parameter is appended to the id (e.g. `conv-1&download=1`).

## Steps to reproduce

1. Create conversation `conv-1`.
2. `GET /api/chat/export?id=conv-1&download=1`.
3. Export returns 404 even though `?id=conv-1` alone works.

## Expected behavior

Only the `id` query value is used.

## Actual behavior

Trailing query fragments become part of the conversation id lookup.

glorysr1209-png:
## Summary

Closes #357.

Parse the `id` query parameter from `/api/chat/export` instead of splitting on the substring `id=`.

## Changes

- `crates/genie-core/src/server.rs`: `export_conv_id_from_path` helper.

## Real Behavior Proof

- [x] `GET /api/chat/export?id=conv-1&download=1` resolves `conv-1`.

## Test plan

- [ ] `cargo test -p genie-core`

### DIFF
--- a/crates/genie-core/src/server.rs
+++ b/crates/genie-core/src/server.rs
@@ -312,6 +312,16 @@ impl RequestRoute<'_> {
     }
 }
 
+fn export_conv_id_from_path(path: &str) -> &str {
+    let query = path.split('?').nth(1).unwrap_or("");
+    for part in query.split('&') {
+        if let Some(value) = part.strip_prefix("id=") {
+            return value;
+        }
+    }
+    ""
+}
+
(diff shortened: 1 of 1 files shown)


ai-hpc:
Not planned for Milestone 1.

PR #369, opened 2026-06-01
fix(ota): implement download-verify-apply pipeline with Ed25519 manif…

### LINKED ISSUE #368: [bug] ota: `check_update()` reports `update_available: true` but download, checksum verification, and binary replacement are not implemented — OTA apply is inert and update channel is unguarded
### Summary

`OtaManager::check_update()` is wired and live, can surface `update_available: true` to the dashboard and `genie-ctl`, but the entire apply side of the pipeline — download, SHA-256/signature verification, atomic binary replacement — does not exist; `prepare_staging`, `backup_current`, and `rollback` are dead code that nothing ever calls.

### Steps to reproduce

1. Build and run `genie-core` on any supported target (Jetson or x86_64 dev host).
2. Call `GET /api/health` — confirm `version` is present and OTA check is wired at startup.
3. Grep the entire `crates/genie-core/src/ota/mod.rs` for any of: `download`, `verify`, `checksum`, `sha256`, `blake3`, `signature`, `apply_update`, `rename`, `tokio::fs::copy` (outside the backup/rollback stubs).
4. Observe that none of those symbols exist.
5. Follow the `download_url` field in `ReleaseInfo` — it is populated by `fetch_latest_release()` from the GitHub API JSON, but every caller immediately discards the value; no function ever fetches the URL.

**Code pointers:**
- `crates/genie-core/src/ota/mod.rs:6–21` — module doc promises "Verify checksums" as step 5
- `crates/genie-core/src/ota/mod.rs:59–80` — `check_update()` is live, makes real network requests
- `crates/genie-core/src/ota/mod.rs:83–132` — `fetch_latest_release()` populates `download_url` and returns it
- `crates/genie-core/src/ota/mod.rs:109–123` — `download_url` is derived from `browser_download_url` in the GitHub asset JSON — this URL is controlled by whoever published the release and is not pinned or verified
- `crates/genie-core/src/ota/mod.rs:140–189` — `prepare_staging`, `backup_current`, `rollback` exist but are never invoked by any code path
- `crates/genie-core/src/ota/mod.rs:216–238` — `github_api_get` shells to `curl -sS` with no CA bundle pinning and no `--max-time` flag; a hung DNS/TCP handshake blocks the caller indefinitely

### Expected behavior

`check_update()` returns `update_available: true` → operator (or automated path) triggers apply → system:
1. Downloads the aarch64 binary asset to the staging directory over HTTPS.
2. Computes its SHA-256 (or Blake3) digest and compares against a signed manifest published alongside the release.
3. Verifies the manifest signature with a pinned Ed25519 public key baked into the binary (same pattern as `crates/genie-core/src/skills/signature.rs` which already implements Ed25519 verify).
4. Backs up current binaries atomically.
5. Replaces each binary via `rename(staging/bin → install/bin)` (atomic on same filesystem).
6. Triggers a rolling restart.
7. If the health check fails within 60 s, rolls back from backup.

### Actual behavior

Steps 1–7 above do not exist. The apply side is completely absent. `check_update()` is the only live function; it fetches release metadata and returns a status struct. The `download_url` value is never consumed. The staging and backup directories are never created or written. No binary is ever downloaded, verified, or replaced.

**Secondary issues found while auditing:**
- `github_api_get` has no `--max-time` on the `curl` invocation (line 219) — a network hang in `check_update()` will block the calling task indefinitely; no timeout propagates.
- `browser_download_url` is taken verbatim from the GitHub API JSON response. A MITM on the unverified `api.github.com` TLS connection, or a compromised release tag, can substitute an arbitrary URL with no verification guard.
- `backup_current` iterates a hardcoded list of binary names (`genie-core`, `genie-ctl`, …) but does not check that the destination staging directory has been created first; if called without `prepare_staging`, the `tokio::fs::copy` calls fail silently (the function returns `Ok(())` on a missing source, line 159).

### Hardware

### JetPack / L4T version

N/A

 v1.0.0-alpha.10 / 13b8884

### Relevant logs

```shell
# check_update() succeeds, reports update_available — but apply does not exist
$ curl -s http://127.0.0.1:7420/api/health | jq .version
"1.0.0-alpha.10"

# No ota/apply endpoint exists anywhere in the router (server.rs classify_route)
$ grep -r "ota" crates/genie-core/src/server.rs
(no output)

# The OTA module has no download or verify functions
$ grep -E "(fn download|fn verify|fn apply|checksum|sha256|signature)" crates/genie-core/src/ota/mod.rs
(no output)

# prepare_staging / backup_current / rollback are never called
$ grep -rn "prepare_staging\|backup_current\|rollback\|OtaManager" crates/ --include="*.rs" \
  | grep -v "^crates/genie-core/src/ota/mod.rs"
(no output — dead code)
```

### Additional context

**Risk model:** The gap matters in two ways.

1. **Silent false-positive:** Any path that surfaces `update_available: true` (dashboard, `genie-ctl`, future automation) gives the operator a misleading signal that an update is ready to apply when there is no safe apply path. If an operator works around this by manually replacing binaries, they bypass all verification.

2. **Supply-chain attack surface for future implementers:** The next developer to add a download step will find `backup_current` and `rollback` already in place and may assume verification is handled elsewhere. The result is a download-and-execute path with no integrity check. The project already has the right primitive for this — Ed25519 verify in `crates/genie-core/src/skills/signature.rs` (issue #175 / #280) — but it is not plumbed into the OTA path.

**Files that need changes (non-exhaustive):**

| File | Change needed |
|------|--------------|
| `crates/genie-core/src/ota/mod.rs` | Add `download_binary()`, `verify_sha256()`, `apply_update()`, add `--max-time` to curl |
| `crates/genie-core/src/ota/verify.rs` (new) | Ed25519 manifest signature verification, pinned public key |
| `crates/genie-common/src/config.rs` | Add `[ota]` section: `enabled`, `auto_apply`, `public_key_path` |
| `crates/genie-core/src/server.rs` | Add `POST /api/ota/check` and `POST /api/ota/apply` routes |
| `crates/genie-ctl/src/main.rs` | Add `ota check` / `ota apply` subcommands |
| `crates/genie-core/src/main.rs` | Wire OTA manager to server and config |

The governor should pause mode switching during apply (as the doc already notes) — a coordination point with `crates/genie-governor/src/governor.rs` is needed.

enjoyandlove:
## Summary

Implements the missing OTA apply pipeline. `check_update()` was live and could surface `update_available: true`, but the entire download–verify–replace side did not exist: `prepare_staging`, `backup_current`, and `rollback` were dead code that nothing ever called, `github_api_get` had no `--max-time` flag (indefinite hang on a stalled connection), and `backup_current` returned `Ok(())` silently when the staging directory did not exist. This PR closes that gap end-to-end.

Closes #368.

## Changes

- **`crates/genie-core/src/ota/mod.rs`** — rewrote the module: added `--max-time` to every `curl` call; added `download_file`, `download_to_bytes`, `verify_sha256`, and `apply_update`; fixed `backup_current` to return `Err` (not silent `Ok`) when the backup directory is missing; extended `ReleaseInfo` with `checksum_url` and `checksum_sig_url`; extended `UpdateStatus` with `signed_release`.
- **`crates/genie-core/src/ota/verify.rs`** (new) — `OtaVerifyingKey`: loads an Ed25519 public key from `[ota].public_key_path`, verifies detached signatures over manifest bytes with `verify_strict` (same pattern as the skill loader). `parse_manifest`: parses `sha256sums`-format manifests into `(filename, hex_hash)` pairs.
- **`crates/genie-core/src/prompt_sha.rs`** — added `sha256_hex_bytes(&[u8])` alongside the existing `sha256_hex(&str)` so the OTA verifier can hash downloaded binary content without a UTF-8 round-trip.
- **`crates/genie-common/src/config.rs`** — added `[ota]` section (`OtaConfig`: `enabled`, `auto_apply`, `public_key_path`, `network_timeout_secs`); defaults to disabled so existing deployments are unaffected.
- **`crates/genie-core/src/server.rs`** — added `POST /api/ota/check` (read-only) and `POST /api/ota/apply` (mutating, token-gated when a local API token is configured); wired `OtaManager` as an optional field on `ChatServer`.
- **`crates/genie-core/src/main.rs`** — constructs `OtaManager` from config at startup and attaches it to `ChatServer`.
- **`crates/genie-ctl/src/main.rs`** — added `genie-ctl ota check`, `ota apply`, and `ota rollback` subcommands. `rollback` is a direct filesystem copy (no API call) so it works when `genie-core` is not running after a failed apply.
- **`crates/genie-{health,governor,api}/src/*.rs`** — added `ota: Default::default()` to `Config {}` struct literal initializers in test helpers that were broken by the new field.

## Real Behavior Proof

- [ ] `jetson`
- [ ] `raspberry_pi`
- [ ] `portable_sbc`
- [x] `laptop`
- [ ] `mac`
- [ ] CI-only / docs-only
- [ ] Not run locally

### What I ran

x86_64 Linux, stable Rust toolchain, no cross-compile.

```
cargo check --workspace --locked
cargo test --workspace --locked
```

All crates check clean (zero warnings after removing the unused `Verifier` import). Full workspace test suite: **0 failures** across all test targets.

The new unit tests in `ota/mod.rs` cover: version comparison (including pre-release stripping), `derive_binary_url` path manipulation, and `OtaManager` directory layout.

The new unit tests in `ota/verify.rs` cover: load-and-verify a valid signature, tampered manifest rejection, missing key file error, `parse_manifest` happy path, malformed line, and short hash rejection.

The new tests in `prompt_sha.rs` cover: `sha256_hex_bytes` against the FIPS-180 empty-string vector and cross-check against the existing `sha256_hex` variant.

### What I observed

```
Finished `dev` profile [unoptimized + debuginfo] target(s) in 7.52s
```

Full workspace test run — representative subset:

```
running 712 tests
test result: ok. 712 passed; 0 failed; 0 ignored
running 102 tests
test result: ok. 102 passed; 0 failed; 0 ignored
running 18 tests
test result: ok. 18 passed; 0 failed; 0 ignored
```

**Validation gap:** the live network path (`github_api_get` → GitHub API → asset download → curl apply) was not exercised end-to-end because no real release exists for this branch and the test environment has no signed `ota.pub` key. The apply pipeline depends on `curl` being present and on the operator having placed a matching Ed25519 public key at `[ota].public_key_path` before enabling OTA. A reviewer with Jetson hardware and a signed test release should verify the `ota apply` golden path before merge.

## Test plan

1. Confirm `cargo check --workspace --locked` and `cargo test --workspace --locked` are green on your machine.
2. On a device with `genie-core` running and `[ota].enabled = false` (the default), confirm `POST /api/ota/apply` returns `503` with `"OTA apply is disabled"`.
3. Set `[ota].enabled = true` without a valid `public_key_path` file, confirm `apply` returns `500` with `"OTA public key not found"`.
4. Run `genie-ctl ota check` — confirm it prints current and latest version and reports `Signed release: no` for any release that lacks `checksums.sha256` and `checksums.sha256.sig` assets.
5. Run `genie-ctl ota rollback` on a device where the backup directory does not exist — confirm it exits with `"nothing to roll back"` rather than panicking.
6. (Jetson / signed release required) Generate an Ed25519 key pair, sign a `checksums.sha256` manifest for a real build, publish as a GitHub release asset, set `public_key_path` and `enabled = true`, and run `genie-ctl ota apply`. Confirm binaries are replaced and the old ones are in the backup directory.

## Notes for reviewers

- The `apply_update` function does **not** trigger a service restart; the operator must `systemctl restart genie-core` (and other replaced services) manually after a successful apply. Automated restart is deferred: it requires deciding which units to restart and a health-check + rollback loop that belongs in a follow-up.
- The governor pause during apply (noted in the issue) is not yet wired. The OTA apply is a blocking async operation so mode switches may race with it. Coordination with `genie-governor` is a follow-up.
- `rollback` in `genie-ctl ota rollback` is a direct `std::fs::copy` so it works when `genie-core` is not running. The HTTP `POST /api/ota/apply` error response includes a `"hint"` field pointing the operator to `genie-ctl ota rollback` if apply fails after the backup step.
- `download_url` in `ReleaseInfo` is still derived from `browser_download_url` in the GitHub API response (same as before), but the `apply_update` function now refuses to write any binary to the install directory until the manifest signature over `checksums.sha256` verifies cleanly against the pinned key — so a compromised or MITM'd `browser_download_url` cannot bypass integrity checking.

### DIFF
--- a/crates/genie-api/src/routes.rs
+++ b/crates/genie-api/src/routes.rs
@@ -943,6 +943,7 @@ mod tests {
             web_search: WebSearchConfig::default(),
             connectivity: ConnectivityConfig::default(),
             http: Default::default(),
+            ota: Default::default(),
         }
     }
 
--- a/crates/genie-common/src/config.rs
+++ b/crates/genie-common/src/config.rs
@@ -41,6 +41,9 @@ pub struct Config {
 
     #[serde(default)]
     pub http: HttpServerConfig,
+
+    #[serde(default)]
+    pub ota: OtaConfig,
 }
 
 #[derive(Debug, Deserialize)]
@@ -900,6 +903,50 @@ pub enum WebSearchProvider {
     Searxng,
 }
 
+/// OTA (over-the-air) update configuration.
+///
+/// Controls the download–verify–apply pipeline. Disabled by default so the
+/// check-only path keeps working (as before) without exposing the apply side
+/// to an unconfigured deployment. Set `enabled = true` and configure a public
+/// key to unlock `POST /api/ota/apply` and `genie-ctl ota apply`.
+#[derive(Debug, Deserialize, Clone)]
+pub struct OtaConfig {
+    /// Enable the OTA apply pipeline. When false, the check endpoint is still
+    /// available but `/api/ota/apply` and `genie-ctl ota apply` are rejected.
+    #[serde(default)]
+    pub enabled: bool,
+
+    /// Automatically apply an update when `check_update()` finds one (requires
+    /// `enabled = true`). Off by default — operator must confirm with the CLI or
+    /// API even when OTA is enabled.
+    #[serde(default)]
+    pub auto_apply: bool,
+
+    /// Path to the Ed25519 public key file (base64-encoded 32-byte key) used to
+    /// verify release manifests before any binary is written to disk. Required
+    /// when `enabled = true`; apply is refused if the key cannot be loaded.
+    #[serde(default = "defaults::ota_public_key_path")]
+    pub public_key_path: PathBuf,
+
+    /// Network timeout in seconds for GitHub API calls and asset downloads.
+    /// Bounds every `curl` call so a hung DNS or TCP handshake cannot block
+    /// the caller indefinitely (the bug documented in the issue for line 219
+    /// of the original ota/mod.rs).
+    #[serde(default = "defaults::ota_network_timeout_secs")]
+    pub network_timeout_secs: u64,
+}
+
+impl Default for OtaConfig {
+    fn default() -> Self {
+        Self {
+            enabled: false,
+            auto_apply: false,
+            public_key_path: defaults::ota_public_key_path(),
+            network_timeout_secs: defaults::ota_network_timeout_secs(),
+        }
+    }
+}
+
 /// Inbound HTTP-server hardening shared by `genie-core` (`:3000`) and
 /// `genie-api` (`:3080`).
 ///
@@ -1684,6 +1731,7 @@ mod tests {
             web_search: WebSearchConfig::default(),
             connectivity: ConnectivityConfig::default(),
             http: HttpServerConfig::default(),
+            ota: OtaConfig::default(),
         }
     }
 
@@ -2550,6 +2598,38 @@ expected_runtime_contract_hash = "abc123"
     }
 
     #[test]
+    fn ota_defaults_to_disabled() {
+        let config = test_config();
+        assert!(!config.ota.enabled);
+        assert!(!config.ota.auto_apply);
+        assert_eq!(
+            config.ota.public_key_path,
+            PathBuf::from("/etc/geniepod/ota.pub")
+        );
+        assert_eq!(config.ota.network_timeout_secs, 30);
+    }
+
+    #[test]
+    fn ota_config_parses() {
+        let config: OtaConfig = toml::from_str(
+            r#"
+enabled = true
+auto_apply = false
+public_key_path = "/etc/geniepod/ota-release.pub"
+network_timeout_secs = 60
+"#,
+        )
+        .unwrap();
+        assert!(config.enabled);
+        assert!(!config.auto_apply);
+        assert_eq!(
+            config.public_key_path,
+            PathBuf::from("/etc/geniepod/ota-release.pub")
+        );
+        assert_eq!(config.network_timeout_secs, 60);
+    }
+
+    #[test]
     fn http_server_config_falls_back_when_section_absent() {
         // Existing deployments have no [http] section yet — the whole config
         // must still parse and use the hardened defaults.
@@ -2896,5 +2976,11 @@ mod defaults {
     }
     pub fn esp32c6_uart_response_timeout_ms() -> u64 {
         250
+    }
+    pub fn ota_public_key_path() -> PathBuf {
+        PathBuf::from("/etc/geniepod/ota.pub")
+    }
+    pub fn ota_network_timeout_secs() -> u64 {
+        30
     }
 }
--- a/crates/genie-core/src/main.rs
+++ b/crates/genie-core/src/main.rs
@@ -316,10 +316,7 @@ async fn main() -> Result<()> {
         let origin_resolver =
             genie_core::origin_auth::OriginResolver::from_config(&config.core.origin_auth);
 
-        let ota_manager = genie_core::ota::OtaManager::new(
-            &config.data_dir,
-            config.ota.clone(),
-        );
+        let ota_manager = genie_core::ota::OtaManager::new(&config.data_dir, config.ota.clone());
 
         let chat_server = genie_core::server::ChatServer::new(
             llm,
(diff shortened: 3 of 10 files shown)


ai-hpc:
Not planned for Milestone 1.

PR #371, opened 2026-06-01
fix(skills): record signature status in audit trail and flag writable…

### LINKED ISSUE #370: [bug] skills/loader: TOCTOU race between Ed25519 .so signature verification and dlopen allows adversarial skill substitution — full in-process native code injection"
### Summary

`SkillLoader::load_skill()` reads the `.so` bytes to verify their Ed25519 signature, then calls `Library::new(path)` (`dlopen`) on the same path-string moments later; an attacker who can `rename()` a malicious binary over the skills directory between those two calls bypasses signature enforcement entirely and injects arbitrary native code into the genie-core address space — the only process-level guard that could mitigate this (Landlock) is hollow per #347.

### Steps to reproduce

1. Enable `require_signature = true` in `[skills]` config and place a legitimately-signed `hello.so` in the skills directory.
2. At the moment `genie-core` calls `load_all()` / `load_skill()` (at startup or via hot-reload), have a second process execute:
   ```sh
   # compile a malicious .so that execs a shell in genie-core's address space
   cp /tmp/evil.so /opt/geniepod/skills/hello.so   # atomic rename or O_TRUNC write
   ```
3. The window is the CPU instructions between `verify_skill_signature()` returning `true` and `Library::new(path)` completing the `dlopen` syscall — on a loaded Jetson this is a realistic multi-millisecond window, not a theoretical one.
4. `Library::new(path)` opens the now-substituted binary; `dlopen` maps and runs its `.init` section before returning. Arbitrary code now runs inside genie-core with its uid, open file descriptors, memory, and active TLS sessions.

```rust
// crates/genie-core/src/skills/loader.rs  (load_skill, current code)
let signed = verify_skill_signature(path, manifest, &self.trusted_keys); // ← reads bytes, closes fd
// --- TOCTOU WINDOW --- attacker calls rename("/tmp/evil.so", path) here
let lib = unsafe { Library::new(path) }  // ← dlopen opens path again, gets evil.so
```

The comment in the source ("the bytes have been verified against a trusted key above") is incorrect — a `rename()` between the two calls means the verified bytes are NOT the bytes that execute.

### Expected behavior

The bytes that pass signature verification are guaranteed to be the bytes that `dlopen` maps. One correct fix: open the file once with `open(O_RDONLY)`, verify the signature over the fd contents via `read()`, then pass `/proc/self/fd/<n>` to `Library::new()`. The fd is a stable reference to the inode; a concurrent `rename()` in the directory cannot change which bytes the fd points to.

Alternative: read the verified bytes into memory, write them to a `memfd_create` anonymous fd, and load from that fd — also eliminates the race and requires no `/proc` path construction.

### Actual behavior

`verify_skill_signature()` calls `std::fs::read(skill_path)` (opens, reads, closes), then `Library::new(path)` calls `dlopen(path_string)` (opens again). Two distinct kernel `open` calls on the same filename string. A `rename()` between them loads a different inode than the one whose bytes were verified.

The code comment on line ~480 of `loader.rs` ("with require_signature on, the bytes have been verified against a trusted key above") is therefore misleading — it implies the gate is closed when it is not.

### Hardware

### JetPack / L4T version

N/A

v1.0.0-alpha.9 (commit a70670f)

### Relevant logs

```shell
WARN genie_core::skills::loader: skill loaded  skill="hello" path="/opt/geniepod/skills/hello.so"
# (no further warning — the race is silent; the evil .so init section runs before the log line above)
```

### Additional context

**Files requiring changes** (all paths relative to repo root):

| File | Change needed |
|---|---|
| `crates/genie-core/src/skills/loader.rs` | Primary fix: open `.so` once, verify over fd contents, load from `/proc/self/fd/<n>` or `memfd_create` — eliminate the two-open pattern in `load_skill()` |
| `crates/genie-core/src/skills/signature.rs` | Add `verify_detached_fd(fd: RawFd, ...)` alongside the existing `verify_detached`; `load_skill` should call the fd variant |
| `crates/genie-core/src/skills/mod.rs` | Export the new fd-based verification path; update module docs to reflect the guaranteed-atomic load invariant |
| `crates/genie-core/src/security/sandbox.rs` | Landlock (#347) would restrict writes to the skills directory and shrink the attacker's write window; completing that implementation is a defense-in-depth complement to the fd fix |
| `crates/genie-core/src/security/audit.rs` | Skill load events (path, signature status, key_id) must be written to the audit JSONL *after* the fd-verified load succeeds, not before, so the audit record reflects reality |
| `crates/genie-core/src/tools/dispatch.rs` | `SkillLoader` is called from the tool dispatcher; any reload path (hot-reload, `genie-ctl skills reload`) must also use the fd-safe loader — audit this call site |
| `crates/genie-core/src/security/taint.rs` | Skill `execute()` args and return values are untainted today; add taint propagation from skill inputs to ensure any exfiltration attempt by a loaded skill is visible to the audit layer |

**Compounding factors:**
- Landlock sandbox (the kernel-level control that would have blocked the `rename()` from another process) is not enforced — issue #347.
- LoopGuard and Tainted<T> are dead code — issue #349 — so a compromised skill can loop tool calls freely.
- Skills execute as native code in genie-core's address space with no isolation; a malicious `.init` section or `execute()` body has full access to memory, open SQLite handles (conversation history, memory), active TLS state, and environment variables including any API keys loaded at startup.

**Severity rationale:** This is a signature-bypass leading to arbitrary code execution in the home AI daemon that has access to the Home Assistant API, the local conversation database, and any credentials stored in the environment. On a Jetson deployed as a household appliance the attack surface includes other local network services that can write to the Jetson filesystem (e.g., a compromised NAS share mounted over NFS).

greatjourney589:
## Summary

Closes #370 (skills/loader TOCTOU). The core TOCTOU race — verify bytes, then `dlopen` the same path again — was already sealed by the `memfd_create` fix. This PR closes the remaining audit gaps called out in the issue: unsigned skill loads were silent, the skills directory itself was never checked for attacker-writable permissions (the write window is *before* any read, so the sealed-memfd fix doesn't help there), skill invocations in `tool-audit.jsonl` carried no signature metadata, and `NativeSkill` was missing from the taint label set.

## Changes

- `skills/loader.rs`: `"skill loaded"` tracing event now includes `signed=` and `key_id=` fields — unsigned skill loads are no longer silent in structured logs
- `security/audit.rs`: added `check_skills_dir_permissions`; startup audit now emits a `Critical` finding for a world-writable skills directory and a `Warning` for group-writable — an attacker with write access can plant a `.so` before any read, bypassing the memfd protection
- `security/taint.rs`: added `TaintLabel::NativeSkill` so skill return values can be tagged and observed flowing to display or network sinks
- `tools/dispatch.rs`: `ToolAuditEvent` gains `skill_signed: Option<bool>` and `skill_key_id: Option<String>` (omitted for built-in tools via `skip_serializing_if`); `audit_tool_call` populates them by looking up the manifest in the loaded skills list — every skill invocation in `tool-audit.jsonl` now records whether the executed code was cryptographically verified

## Real Behavior Proof

- [ ] `jetson`
- [ ] `raspberry_pi`
- [ ] `portable_sbc`
- [x] `laptop`
- [ ] `mac`
- [ ] CI-only / docs-only
- [ ] Not run locally

### What I ran

```
RUSTUP_TOOLCHAIN=stable cargo check -p genie-core
RUSTUP_TOOLCHAIN=stable cargo clippy -p genie-core -- -D warnings
RUSTUP_TOOLCHAIN=stable cargo test -p genie-core -- security::audit
RUSTUP_TOOLCHAIN=stable cargo test -p genie-core -- skills::loader
RUSTUP_TOOLCHAIN=stable cargo test -p genie-core -- skill_audit_event
```

Linux x86_64 dev machine, no Jetson hardware available. The changed paths are pure Rust logic (taint enum, filesystem permission checks, structured log fields, JSONL serialization) with no voice/audio/HA dependency — the test suite covers the full behaviour.

### What I observed

All tests pass, zero clippy warnings:

```
test security::audit::tests::audit_world_writable_skills_dir_is_critical ... ok
test security::audit::tests::audit_group_writable_skills_dir_is_warning ... ok
test security::audit::tests::audit_safe_skills_dir_has_no_permission_findings ... ok
test skills::loader::tests::loader_toctou_disk_swap_after_load_has_no_effect ... ok
test skills::loader::tests::loader_accepts_valid_signature ... ok
test skills::loader::tests::loader_rejects_tampered_so ... ok
test tools::dispatch::tests::skill_audit_event_includes_signature_fields ... ok
```

The new dispatch test confirms that executing `hello_world` (unsigned, no manifest) writes `"skill_signed": false` to `tool-audit.jsonl` and omits `skill_key_id` — verifying that unsigned skill loads are recorded, not silently accepted.

## Test plan

1. Create a world-writable skills directory and start `genie-core` — `journalctl -u genie-core` should show a `SECURITY: skills directory is world-writable` Critical finding at startup.
2. Load a signed skill (`genie-ctl skills install …`) and invoke it via chat — `tool-audit.jsonl` entry should have `"skill_signed": true` and `"skill_key_id": "<your-key-id>"`.
3. Load an unsigned skill with `require_signature = false` and invoke it — `tool-audit.jsonl` should have `"skill_signed": false` with no `skill_key_id` field.
4. Run `cargo test -p genie-core -- skills::loader security::audit skill_audit_event` — all must pass.

## Notes for reviewers

- The `NativeSkill` taint label is added but not yet wired into any sink policy (no call sites in `exec_skill` yet). It is a structural prerequisite for future enforcement — tagging skill outputs at call sites is the obvious next step but is out of scope for this fix.
- The skills-dir permission check calls `crate::skills::skills_dir()` directly inside `run_audit`; no circular dependency (neither `skills/loader.rs` nor `skills/signature.rs` imports from `security`).
- The `skill_signed` / `skill_key_id` lookup in `audit_tool_call` acquires the skills `Mutex` briefly and synchronously — no lock is held across any `await` point, so there is no deadlock or `await_holding_lock` hazard.

### DIFF
--- a/crates/genie-core/src/security/audit.rs
+++ b/crates/genie-core/src/security/audit.rs
@@ -425,8 +425,9 @@ mod tests {
         let mut findings = Vec::new();
         check_skills_dir_permissions(&path, &mut findings);
         assert!(
-            findings.iter().any(|f| f.id == "fs.skills_dir.group_writable"
-                && f.severity == Severity::Warning),
+            findings
+                .iter()
+                .any(|f| f.id == "fs.skills_dir.group_writable" && f.severity == Severity::Warning),
             "group-writable skills dir must be a Warning finding"
         );
 
@@ -438,10 +439,8 @@ mod tests {
     fn audit_safe_skills_dir_has_no_permission_findings() {
         use std::os::unix::fs::PermissionsExt;
 
-        let path = std::env::temp_dir().join(format!(
-            "geniepod-audit-skills-safe-{}",
-            std::process::id()
-        ));
+        let path =
+            std::env::temp_dir().join(format!("geniepod-audit-skills-safe-{}", std::process::id()));
         let _ = fs::remove_dir_all(&path);
         fs::create_dir_all(&path).unwrap();
         fs::set_permissions(&path, fs::Permissions::from_mode(0o755)).unwrap();
--- a/crates/genie-core/src/security/audit.rs
+++ b/crates/genie-core/src/security/audit.rs
@@ -43,6 +43,12 @@ pub fn run_audit(config_path: &Path, data_dir: &Path) -> Vec<AuditFinding> {
     // 5. Check if API port is bound to localhost only.
     check_localhost_binding(&mut findings);
 
+    // 6. Check skills directory permissions. A world- or group-writable skills
+    //    dir allows an attacker to plant a malicious .so before genie-core reads
+    //    it — even the sealed-memfd TOCTOU fix reads bytes from disk, so the
+    //    write window is before the read, not between verify and dlopen.
+    check_skills_dir_permissions(&crate::skills::skills_dir(), &mut findings);
+
     // Log all findings.
     for finding in &findings {
         match finding.severity {
@@ -257,6 +263,50 @@ fn check_not_root(findings: &mut Vec<AuditFinding>) {
     }
 }
 
+fn check_skills_dir_permissions(skills_dir: &Path, findings: &mut Vec<AuditFinding>) {
+    if !skills_dir.exists() {
+        return;
+    }
+
+    #[cfg(unix)]
+    {
+        use std::os::unix::fs::PermissionsExt;
+        if let Ok(meta) = std::fs::metadata(skills_dir) {
+            let bits = meta.permissions().mode() & 0o777;
+
+            if bits & 0o002 != 0 {
+                findings.push(AuditFinding {
+                    id: "fs.skills_dir.world_writable".into(),
+                    severity: Severity::Critical,
+                    message: format!(
+                        "skills directory is world-writable (mode {:o}) — \
+                         an untrusted process can plant a malicious .so before \
+                         genie-core reads it, bypassing signature verification",
+                        bits
+                    ),
+                    remediation: format!(
+                        "chmod 755 {0} && chown root:root {0}",
+                        skills_dir.display()
+                    ),
+                });
+            }
+
+            if bits & 0o020 != 0 {
+                findings.push(AuditFinding {
+                    id: "fs.skills_dir.group_writable".into(),
+                    severity: Severity::Warning,
+                    message: format!(
+                        "skills directory is group-writable (mode {:o}) — \
+                         a process in the same group can replace skill binaries",
+                        bits
+                    ),
+                    remediation: format!("chmod 755 {}", skills_dir.display()),
+                });
+            }
+        }
+    }
+}
+
 fn check_localhost_binding(findings: &mut Vec<AuditFinding>) {
     // This is informational — we always bind to 127.0.0.1.
     findings.push(AuditFinding {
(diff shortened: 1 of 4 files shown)


ai-hpc:
Not planned for Milestone 1.

PR #373, opened 2026-06-02
fix(voice/vad): pre-flight cache check and 10 s timeout prevent offli…

### LINKED ISSUE #372: [bug] [bug] voice/vad.rs: Silero VAD Python subprocess has no timeout and no model cache pre-flight — torch.hub.load hangs voice loop permanently on offline or LAN-only Jetson
### Summary

`vad::detect_speech()` spawns `python3` inline and calls `torch.hub.load('snakers4/silero-vad', ...)` with no local cache check and no subprocess timeout; on a freshly deployed or factory-reset Jetson behind a LAN-only firewall (the primary GeniePod target), `torch.hub.load` attempts a GitHub download that never completes, `tokio::process::Command::output().await` blocks forever, and the entire voice loop Tokio task is permanently deadlocked — recovery requires killing and restarting `genie-core`

### Steps to reproduce

1. Deploy genie-core on a Jetson with no internet access (home LAN behind a firewall that blocks github.com/raw.githubusercontent.com — the standard household setup).
2. Ensure the Silero VAD model is not cached (`~/.cache/torch/hub/snakers4_silero-vad_master/` does not exist, e.g. fresh flash of JetPack).
3. Start `genie-core` with voice enabled and trigger the wake word.
4. `voice_loop::run()` records audio, then calls `vad::detect_speech(wav_path)`.
5. `detect_speech` spawns:
   ```
   python3 -c "
     import torch
     model, utils = torch.hub.load(
         repo_or_dir='snakers4/silero-vad',
         model='silero_vad',
         trust_repo=True   # ← first-run: attempts internet download from GitHub
     )
     ..."
   ```
6. `torch.hub.load()` initiates TCP connections to `github.com` and `raw.githubusercontent.com`. With no internet, the TCP SYNs receive no RST — the kernel holds them open for the full retransmit window (~20–30 minutes on default Linux settings).
7. `Command::output().await` in `detect_speech()` awaits the subprocess's exit, which never comes.
8. The voice loop Tokio task is permanently suspended. No subsequent wake-word events are processed. `genie-core` responds to HTTP health checks (those run on separate tasks) but voice interaction is dead.

```rust
// crates/genie-core/src/voice/vad.rs  (detect_speech, current code)
let output = Command::new("python3")
    .args(["-c", &format!(r#"... torch.hub.load(...) ..."#, wav_path)])
    .output()   // ← no timeout; if torch.hub downloads, this awaits indefinitely
    .await?;
```

There is no `tokio::time::timeout()` wrapper. There is no check that the model file exists locally before spawning the subprocess. There is no `child.kill()` on error paths.

### Expected behavior

`detect_speech()` should:

1. **Pre-flight model check**: verify the Silero VAD model is present in the torch hub cache directory before spawning the subprocess; return an error immediately (allowing the voice loop to skip the utterance and re-arm) if the model is absent instead of attempting a network download.
2. **Subprocess timeout**: wrap `output().await` in `tokio::time::timeout(Duration::from_secs(10), ...)`. On timeout, kill the child process and return an `Err` that the voice loop can handle by logging and skipping the utterance.
3. **Graceful voice loop recovery**: `voice_loop::run()` should treat a `vad::detect_speech` timeout as a non-fatal skip — log the event, send `READY` back to the wake-word process, and continue.

### Actual behavior

`output().await` has no timeout. A missing model cache on an offline Jetson deadlocks the voice loop task permanently. The process must be `kill -9`'d or restarted via `systemctl restart genie-core` to recover.

The hang is completely silent in logs after the last recorded `[voice]` line — `journalctl -u genie-core` shows the STT latency banner from the *previous* successful cycle and then nothing, making it hard to diagnose without strace.

### Hardware

Jetson Orin Nano Super 8 GB

### JetPack / L4T version

 JetPack 6.1 / L4T R36.4

 v1.0.0-alpha.9 (commit a70670f)

### Relevant logs

```shell
# Last voice-loop output before the hang (journalctl -u genie-core):
INFO  genie_core::voice_loop: wake word detected  score=0.92
INFO  genie_core::voice::stt: STT complete  text="hey what time is it"  duration_ms=1043
# -- silence from this point; VAD subprocess is hanging on torch.hub.load --

# Python subprocess (not captured in journalctl, only visible via strace -p <pid> -f):
# [pid NNNNN] connect(4, {AF_INET, github.com:443}, ...) = EINPROGRESS
# [pid NNNNN] epoll_wait(5, ..., -1) = ... (blocked indefinitely)
```

### Additional context

**Files requiring changes** (all paths relative to repo root):

| File | Change needed |
|---|---|
| `crates/genie-core/src/voice/vad.rs` | Primary fix: (1) check for cached model before spawning subprocess; (2) wrap `output().await` in `tokio::time::timeout(Duration::from_secs(10), ...)`; (3) add `kill_on_drop(true)` to the `Command` builder so cancellation is automatic on task drop |
| `crates/genie-core/src/voice_loop.rs` | Treat `vad::detect_speech` `Err` (including new timeout error) as a non-fatal skip: log, increment a consecutive-failure counter, send `READY` back to the wake-word process; abort the voice session only after N consecutive failures |
| `crates/genie-core/src/voice/pipeline.rs` | `VoicePipeline::process_audio()` calls `stt.transcribe_file()` but bypasses the standalone `vad::detect_speech` today; if VAD is wired into this path in future, the same timeout pattern must apply |
| `crates/genie-core/src/voice/streaming.rs` | Streaming pipeline may spawn similar Python subprocess calls; audit for the same no-timeout pattern and apply `tokio::time::timeout` consistently |
| `crates/genie-core/src/voice/stt.rs` | CLI-mode whisper subprocess (`Command::new(whisper_cli_path)`) also has no timeout — a hung whisper process stalls the voice loop identically; apply `tokio::time::timeout(Duration::from_secs(30), ...)` here too |
| `crates/genie-core/src/voice/noise.rs` | `process_recording()` is pure Rust (no subprocess), but review for any future subprocess additions; document that all subprocess calls in the voice path must be wrapped in a timeout |
| `crates/genie-core/src/main.rs` | Add a startup pre-flight that checks for the Silero VAD model cache and emits a loud `tracing::warn!` (or blocks startup if voice is enabled) so operators know to run `python3 -c "import torch; torch.hub.load('snakers4/silero-vad', 'silero_vad', trust_repo=True)"` once while internet is available |

**Deployment scenario this breaks:**

The primary GeniePod use case is a Jetson installed at home, on a LAN, behind a NAT gateway. The Jetson is typically configured to disallow outbound internet to minimize attack surface (good practice for a home AI daemon with HA access). A fresh JetPack flash, a `journalctl -u genie-core` that shows everything healthy, then the first voice attempt silently deadlocks the voice loop — with zero diagnostics in the systemd journal.

**Relationship to prior timeout issues:**

- #181 (LLM stream: no timeout — closed): fixed for the LLM path; voice subprocess paths were not covered.
- #173 (HA client: only TCP connect has timeout — closed): fixed for the HA HTTP client; voice subprocess paths were not covered.
- #67 (STT temp WAV naming race — closed): different class of bug (filename collision vs. blocking hang).

The pattern of Python/binary subprocesses in the voice path (`python3` for VAD, `whisper-cli` for STT, `aplay`/`arecord` for audio I/O, `deep-filter` for noise reduction) means there are multiple independent subprocess calls per voice cycle, none of which are currently wrapped in a timeout. This issue covers VAD as the most critical case (first-run model download), but `stt.rs` CLI mode has the same structural gap and should be fixed in the same PR.

greatjourney589:
## Summary

`vad::detect_speech()` had no subprocess timeout and no check that the Silero VAD model was already cached locally. On a freshly flashed or factory-reset Jetson behind a LAN-only firewall (the standard GeniePod deployment), `torch.hub.load` attempts a GitHub download that never completes, blocking `Command::output().await` forever and permanently deadlocking the voice loop Tokio task. The same structural gap existed in the `whisper-cli` subprocess path in `stt.rs`. This PR fixes both and wires VAD into the voice cycle with non-fatal error handling.

Closes #372

## Changes

- **`voice/vad.rs`** — `silero_model_cached()` (new public fn): fast offline filesystem check for `${XDG_CACHE_HOME}/torch/hub/snakers4_silero-vad_master/`; `detect_speech()`: bail immediately on missing cache, switch from `.output()` to `spawn()` + `kill_on_drop(true)` + `tokio::time::timeout(10 s, child.wait_with_output())` so the subprocess is killed on cancellation; `is_available()`: same `kill_on_drop` + `timeout(5 s)` treatment
- **`voice/stt.rs`** — both whisper-cli invocations in `transcribe_via_cli()` (initial run and GPU-fallback CPU retry) now use `spawn()` + `kill_on_drop(true)` + `tokio::time::timeout(30 s, child.wait_with_output())`; a hung GPU allocation or frozen whisper-cli no longer stalls the voice loop
- **`voice_loop.rs`** — `vad::detect_speech()` wired into `voice_cycle()` between AEC and STT; `Ok((false, _))` and `Err(_)` are both treated as non-fatal skips: log, remove WAV, re-arm the wake-word listener; the voice loop continues running in both cases
- **`main.rs`** — startup pre-flight inside the `#[cfg(feature = "voice")]` block calls `vad::silero_model_cached()` before `voice_loop::run()` and emits a `tracing::warn!` with the exact pre-cache command if the model directory is absent, so the problem is visible in `journalctl -u genie-core` from the very first start

## Real Behavior Proof

- [ ] `jetson`
- [ ] `raspberry_pi`
- [ ] `portable_sbc`
- [x] `laptop`
- [ ] `mac`
- [ ] CI-only / docs-only
- [ ] Not run locally

### What I ran

```
# default features (voice + telegram)
cargo check -p genie-core

# chat-only build (no voice feature) — ensures the pre-flight guard inside
# #[cfg(feature = "voice")] doesn't break the non-voice binary
cargo check -p genie-core --no-default-features

# full lib test suite
```

Both `cargo check` invocations completed with zero errors. `cargo test` ran 701 tests with 0 failures.

The pre-flight logic (`silero_model_cached()`) was manually verified to return `false` on the development machine (no torch hub cache present) and `true` after creating the sentinel directory, confirming the filesystem check works as expected without running a Python subprocess.

The subprocess timeout path was verified by code inspection: `tokio::time::timeout` wraps `child.wait_with_output()` which takes the `Child` by value; when the timeout future is dropped the `Child` drops and `kill_on_drop(true)` sends SIGKILL to the process. This is the standard tokio pattern for bounded subprocess execution.

### What I observed

```
Finished `dev` profile [unoptimized + debuginfo] target(s) in 2.01s   # default features
Finished `dev` profile [unoptimized + debuginfo] target(s) in 3.60s   # --no-default-features
test result: ok. 701 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 4.89s
```

**Validation gap:** The critical path — `torch.hub.load` hanging on a LAN-only Jetson, then recovering within 10 s via the new timeout — cannot be reproduced without Jetson hardware behind a firewall. A reviewer with Jetson access should verify steps 3–4 of the test plan below before merge.

## Test plan

1. **Startup pre-flight warn (any machine with voice enabled):**
   Remove or rename `~/.cache/torch/hub/snakers4_silero-vad_master/`, start `genie-core --voice`, check that `journalctl -u genie-core` contains the warning with the pre-cache command.

2. **VAD non-fatal skip — model absent (any machine):**
   With the cache directory absent, trigger the wake word. Verify in logs that `[voice] VAD error (...): skipping utterance.` appears and the wake-word listener re-arms (next `[voice] Listening for the configured wake phrase...` line appears within a few seconds).

3. **VAD timeout — LAN-only Jetson, model absent:**
   On a Jetson behind a firewall blocking `github.com`/`raw.githubusercontent.com`, with no cached model, trigger the wake word. Verify the voice loop logs the timeout error within ~10 s and re-arms — previously this would hang permanently.

4. **Normal operation — model cached:**
   After running the pre-cache command once with internet access, trigger the wake word and confirm the full cycle completes: VAD detects speech, STT transcribes, LLM replies, TTS speaks.

5. **whisper-cli timeout:**
   Not easily reproducible without a GPU in a stuck state; code review of the `kill_on_drop(true)` + `timeout(30 s)` pattern is the verification path here.

## Notes for reviewers

- The `silero_model_cached()` check uses the `snakers4_silero-vad_master` directory name, which is what `torch.hub.load('snakers4/silero-vad', ...)` produces with the default `main`/`master` branch. If users pin a specific tag or commit, torch hub uses a different directory name and `silero_model_cached()` returns `false` even though the model IS cached. This is intentional: the safe fallback is "assume uncached, bail early" rather than "assume cached, risk a hang". Operators pinning tags can set `XDG_CACHE_HOME` or simply pre-cache without pinning.
- The 10 s VAD timeout was chosen to be well above the typical cold-load time (~200 ms on a warm Jetson iGPU) while staying far below the TCP retransmit window (~20–30 min) that makes the hang undetectable without `strace`.
- `streaming.rs`, `pipeline.rs`, and `noise.rs` were audited: `streaming.rs` and `noise.rs` contain no subprocess calls; `pipeline.rs` (`VoicePipeline::process_audio`) calls `stt.transcribe_file()` but does not call `vad::detect_speech()` directly — if VAD is wired into that path in future the same timeout pattern must apply (noted in code).

### DIFF
--- a/crates/genie-core/src/main.rs
+++ b/crates/genie-core/src/main.rs
@@ -276,6 +276,20 @@ async fn main() -> Result<()> {
                     &config.core.speaker_identity,
                 ),
             };
+            // Pre-flight: check Silero VAD model cache before starting the voice loop.
+            // On a LAN-only Jetson, a missing cache causes torch.hub.load to attempt a
+            // GitHub download that never completes, permanently deadlocking the voice task.
+            // Emit a loud warning so the operator knows to pre-cache while online.
+            if !genie_core::voice::vad::silero_model_cached() {
+                tracing::warn!(
+                    "Silero VAD model not found in torch hub cache \
+                     (~/.cache/torch/hub/snakers4_silero-vad_master/). \
+                     Voice VAD will skip every utterance until the model is cached. \
+                     Pre-cache it while online: \
+                     python3 -c \"import torch; torch.hub.load('snakers4/silero-vad', 'silero_vad', trust_repo=True)\""
+                );
+            }
+
             genie_core::voice_loop::run(
                 voice_cfg,
                 &llm,
--- a/crates/genie-core/src/voice/stt.rs
+++ b/crates/genie-core/src/voice/stt.rs
@@ -448,11 +448,16 @@ impl SttEngine {
                     .stderr(std::process::Stdio::piped())
                     .kill_on_drop(true)
                     .spawn()
-                    .map_err(|e| anyhow::anyhow!("failed to spawn whisper-cli (CPU retry): {}", e))?;
-                let retry = tokio::time::timeout(Duration::from_secs(30), retry_child.wait_with_output())
-                    .await
-                    .map_err(|_| anyhow::anyhow!("whisper-cli (CPU retry) timed out after 30 s"))?
-                    .map_err(|e| anyhow::anyhow!("whisper-cli (CPU retry) I/O error: {}", e))?;
+                    .map_err(|e| {
+                        anyhow::anyhow!("failed to spawn whisper-cli (CPU retry): {}", e)
+                    })?;
+                let retry =
+                    tokio::time::timeout(Duration::from_secs(30), retry_child.wait_with_output())
+                        .await
+                        .map_err(|_| {
+                            anyhow::anyhow!("whisper-cli (CPU retry) timed out after 30 s")
+                        })?
+                        .map_err(|e| anyhow::anyhow!("whisper-cli (CPU retry) I/O error: {}", e))?;
                 if !retry.status.success() {
                     let stderr2 = String::from_utf8_lossy(&retry.stderr);
                     anyhow::bail!("whisper-cli failed (CPU retry): {}", stderr2);
--- a/crates/genie-core/src/voice/stt.rs
+++ b/crates/genie-core/src/voice/stt.rs
@@ -1,6 +1,7 @@
 use anyhow::Result;
 use std::sync::Mutex;
 use std::sync::atomic::{AtomicU64, Ordering};
+use std::time::Duration;
 use tokio::io::{AsyncBufReadExt, AsyncWriteExt, BufReader};
 use tokio::process::{Child, Command};
 use tokio::sync::mpsc;
@@ -417,7 +418,17 @@ impl SttEngine {
             args.push("--no-gpu".to_string());
         }
 
-        let output = Command::new(&self.cli_path).args(&args).output().await?;
+        let child = Command::new(&self.cli_path)
+            .args(&args)
+            .stdout(std::process::Stdio::piped())
+            .stderr(std::process::Stdio::piped())
+            .kill_on_drop(true)
+            .spawn()
+            .map_err(|e| anyhow::anyhow!("failed to spawn whisper-cli: {}", e))?;
+        let output = tokio::time::timeout(Duration::from_secs(30), child.wait_with_output())
+            .await
+            .map_err(|_| anyhow::anyhow!("whisper-cli timed out after 30 s"))?
+            .map_err(|e| anyhow::anyhow!("whisper-cli I/O error: {}", e))?;
 
         // If GPU allocation fails (NvMap error), retry with --no-gpu.
         if !output.status.success() && !self.no_gpu {
@@ -431,7 +442,17 @@ impl SttEngine {
                     stderr.lines().next().unwrap_or("")
                 );
                 args.push("--no-gpu".to_string());
-                let retry = Command::new(&self.cli_path).args(&args).output().await?;
+                let retry_child = Command::new(&self.cli_path)
+                    .args(&args)
+                    .stdout(std::process::Stdio::piped())
+                    .stderr(std::process::Stdio::piped())
+                    .kill_on_drop(true)
+                    .spawn()
+                    .map_err(|e| anyhow::anyhow!("failed to spawn whisper-cli (CPU retry): {}", e))?;
+                let retry = tokio::time::timeout(Duration::from_secs(30), retry_child.wait_with_output())
+                    .await
+                    .map_err(|_| anyhow::anyhow!("whisper-cli (CPU retry) timed out after 30 s"))?
+                    .map_err(|e| anyhow::anyhow!("whisper-cli (CPU retry) I/O error: {}", e))?;
                 if !retry.status.success() {
                     let stderr2 = String::from_utf8_lossy(&retry.stderr);
                     anyhow::bail!("whisper-cli failed (CPU retry): {}", stderr2);
--- a/crates/genie-core/src/voice/vad.rs
+++ b/crates/genie-core/src/voice/vad.rs
@@ -9,15 +9,50 @@
 //! but runs AFTER recording is complete (not in the critical path).
 
 use anyhow::Result;
+use std::time::Duration;
 use tokio::process::Command;
+
+const VAD_TIMEOUT: Duration = Duration::from_secs(10);
+
+/// Return true when the Silero VAD model appears to be in the torch hub
+/// on-disk cache. This is a fast, offline, filesystem-only check.
+///
+/// torch.hub caches under `${XDG_CACHE_HOME}/torch/hub/` (or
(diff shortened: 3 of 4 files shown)


ai-hpc:
Not planned for Milestone 1.

Issue #379, opened 2026-06-03
Jetson: local model misroutes memory/calendar intents to the wrong tool (BFCL accuracy)


ai-hpc:
Found during Jetson validation (#375) on the local Qwen3-4B path. The local model selects the wrong tool for memory/calendar intents — this is the M1 BFCL accuracy gap (5.77% strict baseline) showing up on hardware.

## Observed misroutes (Jetson, port 3000)

| Input | Got | Should be |
| --- | --- | --- |
| `I'm allergic to peanuts` | `set_timer` (3600s, label "timer") | `memory_store` (health/dietary fact) |
| `remember my dentist appointment is next Tuesday 3pm` | `memory_recall` (read) → "I don't remember anything about [the whole sentence] yet" | `memory_store` (write, calendar event) |
| `I have a meeting on Saturday 10AM` | `set_timer` / stream error | calendar/memory event |
| `set a timer for an hour and a half` | 3600s | 5400s (duration under-extraction) |

## Notes

- The deterministic quick router correctly **abstains** on these (no `timer`/`remember`-store match, `quick.rs`), so the **LLM** is making the wrong tool selection — not the router.
- "remember X" routing to `memory_recall` instead of `memory_store` is the most clear-cut: a write verb went to the read tool, then recalled the entire sentence.

## Work

- Expand deterministic quick-router coverage for memory/calendar **store** intents (`I'm allergic to…`, `remember my … appointment`, `I have a meeting…`) and appointment/date parsing.
- Add BFCL fixtures for these intents and track the score.
- Consider prompt / tool-manifest tuning so the model picks `memory_store` over `set_timer` for facts/appointments.

Related: #375 (Jetson validation). M1 — BFCL tool-call accuracy + deterministic memory.

matedev01:
Moving to **M2**. Real on-device routing-accuracy gap (the local Qwen3-4B misroutes memory/calendar intents; the deterministic quick-router correctly abstains). Closing the local-LLM accuracy gap is M2 work.

Issue #375, opened 2026-06-02
Validate tool-dispatch gate + memory subsystem on Jetson Orin Nano 8GB


ai-hpc:
## Motivation

The tool-call gate (#22, merged in #374 + #365) and the memory subsystem (SQLite `FTS5` + BM25 + local hash-embedding semantic recall + typed projection tables) are both verified on the `laptop` profile only — every recent PR's Real Behavior Proof lists the Jetson run as a gap. The M1 contribution gate requires native, performance-sensitive, and runtime-facing changes to carry **Jetson Orin Nano 8GB** proof for latency, memory pressure, and deterministic behavior. This tracks closing that gap for the two subsystems we just hardened.

Hardware: NVIDIA Jetson Orin Nano (Super) 8GB, unified memory, `genie-ai-runtime` on `:8080`, default 4096-token harness.

## Scope — tool dispatch

Validate the `ToolDispatcher` gate (`crates/genie-core/src/tools/dispatch.rs`) on hardware:

- [ ] Gate overhead is negligible — the per-call ACL + per-tool rate-limit + confirmation checks add no measurable latency to tool responses (measure with/without).
- [ ] Two-step confirmation (`requires_confirmation_tools`) works end-to-end through real origins (voice, Telegram, API): first leg returns pending + token, confirming leg within TTL executes, outside TTL errors.
- [ ] Per-tool rate limit (`max_actions_per_minute_by_tool`) bounces a fast loop on-device.
- [ ] The voice fast-paths now routed through the gate (voice `web_search` quick-path + `voice/mod.rs` / `voice/pipeline.rs`) behave correctly on real audio: ACLs/audit apply, refusals are spoken back.
- [ ] Tool-audit log writes succeed on the device data dir under sustained load (exercises the #131 audit-IO surfacing); the per-decision `decision` field is recorded.
- [ ] Home actuation gate works with the configured HA provider: origin ACL, dashboard confirmation token, rate limit, bounded undo, append-only audit.
- [ ] BFCL local-LLM tool-call accuracy measured on Jetson via `genie-ctl bfcl-score-llm` against `genie-ai-runtime`; record the number (quick-router baseline is 5.77% strict on 208 HA-Intents cases).
- [ ] Quick-router + typed-tool response latency measured on Jetson.

## Scope — memory

Validate `crates/genie-core/src/memory/*` + `conversation.rs` on hardware:

[...]

PR #380, opened 2026-06-03
fix(tools): don't leak unparsed tool-call JSON to the user (closes #378)

### LINKED ISSUE #378: Tool-call gate: unparsed tool-call JSON is rendered to the user as a reply
Found during Jetson validation (#375).

## Symptom

When the local model emits a tool-call-shaped output that is **not valid JSON**, the raw JSON is shown to the user as the assistant reply. Observed on Jetson (Qwen3-4B):

```
i have a meeting on Saturday 10AM
{"tool":"set_timer","arguments":{"seconds":60*60*12,"label":"meeting"}}
```

`60*60*12` is a JS arithmetic expression, not a JSON number.

## Root cause

`try_tool_call_with_context` (`crates/genie-core/src/tools/parser.rs:19-20`) requires `serde_json::from_str(...).is_ok()`. Invalid JSON → `extract_json` returns `None` → the call is never recognized. The chat paths' `else` branch then renders `sanitize_output(&llm_response)` (e.g. `server.rs:1071`), which only redacts secrets — so the tool-call JSON reaches the user verbatim.

Affects all five dispatch entry points: the three `server.rs` chat paths, `repl.rs`, and `voice_loop.rs`.

## Expected

A model output that looks like a failed tool call must never be rendered to the user as a normal reply — fall back to a graceful response instead.

## Scope

Agent-side robustness fix. Detect "looks like an unparsed tool call" and suppress/replace it at the render boundary, with a test. Independent of the model's JSON-formatting accuracy (tracked separately).

Related: #375 (Jetson validation). M1 — typed-tool reliability.

ai-hpc:
## Summary

Closes #378. Found during Jetson validation (#375): when the local model emits a
tool-call-shaped output that is **not valid JSON** (e.g. `{"tool":"set_timer","arguments":{"seconds":60*60*12,...}}` — `60*60*12` is a JS expression), `try_tool_call_with_context` returns `None` and the chat paths rendered the **raw model output to the user verbatim**, leaking tool-call JSON.

## Changes

- `tools/parser.rs` — add `is_unparsed_tool_call()` (structurally looks like a tool call but no valid tool-call JSON can be extracted) + `UNPARSED_TOOL_CALL_FALLBACK`, with unit tests (incl. the exact Jetson leak string).
- `tools/mod.rs` — re-export both.
- `server.rs` — guard the three API chat render sites — `/api/chat/stream`, `/api/chat`, and the OpenAI-compatible bridge — so a botched tool call falls back to a graceful reply instead of the raw JSON. Default/normal replies are unchanged.

## Scope note

Voice and REPL stream + speak the response incrementally, so guarding them is a separate, more involved change (post-stream rewrite); noted as follow-up. This PR fixes the observed surface (the port-3000 chat API). It does **not** close the umbrella #375.

The model emitting invalid JSON / picking the wrong tool is tracked separately in #379 (routing accuracy); this PR only stops the leak.

## Real Behavior Proof

- [x] I have NOT verified on Jetson hardware, and I explain the validation gap below.

Tested profile / hardware:
- [x] `laptop`

### What I ran

    cargo build -p genie-core
    cargo test -p genie-core --lib tools::parser
    cargo test -p genie-core --lib -- --skip binary_size_budget
    cargo fmt --all --check

### What I observed
- fmt + clippy clean; 704 genie-core lib tests pass incl. 3 new detector tests (the invalid-JSON leak string is detected; valid tool calls and normal prose are not flagged).
- Pure agent-side render-boundary logic — no Jetson runtime path touched. The original leak was reproduced on Jetson under #375; on-device re-confirmation that the fallback now shows instead of raw JSON is the remaining gap.

### DIFF
--- a/crates/genie-core/src/server.rs
+++ b/crates/genie-core/src/server.rs
@@ -1070,7 +1070,11 @@ async fn handle_chat_stream(
         }
         summary
     } else {
-        let sanitized = crate::security::sandbox::sanitize_output(&llm_response);
+        let sanitized = if crate::tools::is_unparsed_tool_call(&llm_response) {
+            crate::tools::UNPARSED_TOOL_CALL_FALLBACK.to_string()
+        } else {
+            crate::security::sandbox::sanitize_output(&llm_response)
+        };
         if !state.pending.is_empty() && state.mode == StreamMode::Undecided {
             write_stream_event(
                 writer,
@@ -1196,7 +1200,11 @@ pub async fn process_chat_turn(
         )
         .await
     } else {
-        let sanitized = crate::security::sandbox::sanitize_output(&llm_response);
+        let sanitized = if crate::tools::is_unparsed_tool_call(&llm_response) {
+            crate::tools::UNPARSED_TOOL_CALL_FALLBACK.to_string()
+        } else {
+            crate::security::sandbox::sanitize_output(&llm_response)
+        };
         conversations.append_or_log(conv_id, "assistant", &sanitized, None);
         sanitized
     };
@@ -2240,6 +2248,8 @@ async fn handle_openai_chat(
         } else {
             tool_result.output
         }
+    } else if crate::tools::is_unparsed_tool_call(&llm_response) {
+        crate::tools::UNPARSED_TOOL_CALL_FALLBACK.to_string()
     } else {
         llm_response
     };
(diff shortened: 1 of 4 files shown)


ai-hpc:
Also bundling `deploy/scripts/genie-disable-gui.sh` into this PR (per request): a reversible helper to disable the desktop GUI on Jetson and free unified memory for the local LLM (relevant to #375/#376 memory pressure). Standalone deploy script — no Rust changes, doesn't affect the build/tests.

Issue #376, opened 2026-06-02
Re-evaluate 16k context with a Mamba-2 hybrid model and compare BFCL accuracy vs Qwen


ai-hpc:
## Motivation

BFCL tool-call accuracy is the M1 scorecard metric. The current default is **Qwen3-4B Q4_K_M** at the 4096-token Jetson baseline (quick-router baseline: 5.77% strict on 208 HA-Intents cases). The 4096 cap exists largely because a pure-attention model's KV cache grows linearly with context and competes with everything else for the Orin Nano's 8 GB unified memory.

A **Mamba-2 hybrid** model (e.g. Nemotron-H — already listed as a `nemotron-4b` option in `deploy/config/geniepod.toml` — or Zamba2) keeps near-constant state memory for its SSM layers, so a **16k context** becomes affordable on Jetson where Qwen would not. This issue re-evaluates whether that context headroom (and the hybrid model itself) improves tool-call accuracy enough to matter, measured head-to-head against Qwen.

This is a deliberate larger-context experiment. The 4096-token Qwen path stays the product default unless a hybrid + 16k path demonstrably wins on accuracy **and** fits the Jetson memory/latency budget (per the README context strategy and the `ARCHITECTURE.md` optional-provider rules).

## Scope

- [ ] Select a Jetson-servable Mamba-2 hybrid at ~3-4B params, Q4, served **locally** by `genie-ai-runtime` (candidate: Nemotron-H 4B; alt: Zamba2-2.7B). Confirm it loads within the 8 GB budget at 16k context.
- [ ] Run it behind the `LlmClient` facade with `context_window_tokens = 16384` as an intentional larger-context test — must not weaken the default 4096 Qwen path; `agent_harness` still governs the default.
- [ ] Score BFCL on the **same fixtures** as the Qwen baseline (`genie-ctl bfcl-predict-llm` + `bfcl-score-llm`: 208 HA-Intents English + the typed-tool fixtures): strict, tool-name, argument, and parse accuracy.
- [ ] Measure on Jetson Orin Nano 8 GB: unified-memory footprint at 16k (model + KV/state), first-token + tool-response latency — vs Qwen3-4B @ 4096.
- [ ] Produce a head-to-head comparison table and a recommendation.

## Deliverable

| Metric | Qwen3-4B @ 4096 | Mamba-2 hybrid @ 16k |
| --- | --- | --- |
| BFCL strict accuracy | 5.77% (current) | ? |
[...]

PR #383, opened 2026-06-03
feat(deploy): env-configurable llama.cpp ctx-size + model — enable the 16k Nemotron-4B experiment (#376)

ai-hpc:
## Summary

Enables the issue #376 experiment — **16k context, llama.cpp backend, Nemotron-4B Q4** — *without* touching the product defaults or the 4096 Jetson baseline.

`genie-llm.service` hardcoded `--ctx-size 2048` and the model env, and `setup-jetson.sh` deliberately wipes systemd drop-ins (`ctx.conf`/`model.conf`), so there was no clean override path. This drives both from env vars with the **current defaults preserved** (2048 ctx, phi-4-mini), readable from an optional `/etc/geniepod/llm.env`.

## Changes
- `deploy/systemd/genie-llm.service`: `--ctx-size ${GENIEPOD_LLM_CTX_SIZE}` (default 2048) + `EnvironmentFile=-/etc/geniepod/llm.env` (optional, survives setup re-runs). Behavior unchanged unless the env file is present.
- `deploy/config/llm.env.example`: template documenting `GENIEPOD_LLM_MODEL` / `GENIEPOD_LLM_CTX_SIZE`, with the 16k-Nemotron values.
- `doc/experiments/nemotron-16k-llama-cpp.md`: runbook — place the Nemotron GGUF, set the env + `geniepod.toml` (`backend=llama_cpp`, `llm_model_name="nemotron-4b"`, `context_window_tokens=16384`), restart, and measure BFCL/latency/memory vs the Qwen 4096 baseline.

Nemotron is already a supported prompt family (`prompt.rs` → `ModelFamily::Nemotron`). 16k intentionally exceeds the baseline, so the boot harness logs `jetson_baseline_context: fail` — expected for this opt-in path, not a regression.

## Real Behavior Proof
- [x] Built/validated locally (laptop).
- [ ] Not yet run on Jetson — that's the experiment itself (#376).

### What I ran / observed
- `cargo test -p genie-core --test tool_dispatch_test`: `systemd_units_valid`, `config_parses`, and the other deploy tests pass (the unit edit + new files don't break anything; no test asserts the unit's literal ctx-size).
- Defaults unchanged: the repo `geniepod.toml` stays 4096 and the unit still defaults to 2048.

### DIFF
--- a/deploy/config/llm.env.example
+++ b/deploy/config/llm.env.example
@@ -1,0 +1,21 @@
+# Optional per-host overrides for the llama.cpp LLM server (genie-llm.service).
+#
+# Copy to /etc/geniepod/llm.env to override the served model and/or context size
+# WITHOUT editing the systemd unit or using drop-ins (setup-jetson.sh wipes
+# ctx.conf/model.conf drop-ins; this EnvironmentFile is left alone). Only
+# meaningful when [services.llm].backend = "llama_cpp" in geniepod.toml.
+#
+#   sudo cp deploy/config/llm.env.example /etc/geniepod/llm.env   # then edit
+#   sudo systemctl restart genie-llm.service
+
+# Model GGUF served by llama-server (default: the unit's phi-4-mini path).
+#GENIEPOD_LLM_MODEL=/opt/geniepod/models/Qwen3-4B-Q4_K_M.gguf
+
+# llama-server context window / KV-cache size (default: 2048). Raising this costs
+# unified memory — that trade-off is exactly what issue #376 evaluates.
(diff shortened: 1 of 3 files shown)


ai-hpc:
Closing — the 16k experiment (#376) is concluded: nemotron-4b is a 4096-context model and can't be used for 16k, so this enablement isn't needed right now. The env-configurable `genie-llm.service` (model + ctx-size via /etc/geniepod/llm.env) is still a reasonable idea if a genuinely long-context model shows up later — easy to revive from this branch's commits if so.

PR #388, opened 2026-06-04
feat(eval): grounded entity-argument metric for BFCL scorer (#387)

### LINKED ISSUE #387: BFCL strict accuracy is gated by entity-argument grounding, not tool selection
## Motivation

On-device BFCL on Jetson Orin Nano (see #376 comparison) shows strict tool-call accuracy is **bottlenecked by entity-argument extraction, not tool selection** — and this holds across models, so it is not fixable by swapping the LLM:

| metric | Qwen3-4B @4096 | nemotron-4b @4096 |
| --- | --- | --- |
| tool-name | 0.97 | 0.80 |
| **argument** | **0.167** | **0.133** |
| strict | 0.167 | 0.133 |

The models pick the right *tool* 80-97% of the time but get the *entity argument* right only ~13-17%. Typical failure (from the scorer diagnostics):

```
tool[0] argument mismatch: $.entity expected "kitchen lights", got "lights"
```

The model drops the room qualifier. This is the single highest-leverage lever for M1 BFCL/strict accuracy — and it lines up with the M1 thesis: **accuracy from deterministic device-state grounding, not bigger models.**

## Two distinct problems

**1. The metric is too strict (measurement).** `crates/genie-core/src/eval/bfcl.rs` `argument_match` is a raw exact JSON match on the entity *string*. But at runtime `home_control` already grounds the entity — `resolve_device_alias` (`tools/dispatch.rs:1149`) plus the HA provider's alias scoring (`build_aliases` / `best_alias_score`, `ha/provider.rs`) against the device graph. So a predicted `"lights"` may resolve to exactly the right device, yet BFCL marks it wrong. The score under-reports real capability.

**2. Real extraction/grounding gap (capability).** Even with resolution, a bare `"lights"` is ambiguous without room/area context. Partial and room-qualified entity names need disambiguation against the live device graph + current room.

## Proposal / scope

- [ ] Add a **grounded argument comparison** to the BFCL scorer: resolve both predicted and expected entity args against a known alias/device set (or the device graph) before comparing, as an option alongside the existing raw exact-match. Report the delta — this likely recovers a large chunk of the 13-17%.
- [ ] Strengthen **runtime entity resolution** for partial / room-qualified names using room/area context + `best_alias_score` (`resolve_target_in_graph`, `device_aliases`).
- [ ] Add HA-Intents BFCL fixtures that specifically exercise partial / room-qualified entity names, and re-score Qwen3-4B @ 4096.

## Out of scope

- Model swaps — #376 concluded Qwen3-4B is the best 4B tested; the gap is arguments, not the model.

## References

- BFCL comparison + bottleneck data: #376 (closed)
- Routing accuracy: #379
- Scorer: `crates/genie-core/src/eval/bfcl.rs` (`argument_match`)
- Runtime grounding: `tools/dispatch.rs` (`resolve_device_alias`), `ha/provider.rs` (`resolve_target_in_graph`, `build_aliases`, `best_alias_score`), `memory/mod.rs` (`device_aliases`)

ai-hpc:
Closes #387 (measurement half) — opening for review, **not** auto-merging because it touches how the M1 scorecard is computed (see "Decision needed" below).

## What & why

On-device BFCL shows strict accuracy is bottlenecked by **entity-argument matching, not tool selection** (tool-name ~0.97 vs argument ~0.17), and it holds across models — so it is not fixable by swapping the LLM (Qwen3-4B stays the pinned default, per #376).

Root cause (from the 208-case HA-Intents fixtures): the reference home is kitchen-centric with only 8 distinct entity strings, and **98/179 (55%)** of cases have a prompt that lacks the gold entity's room qualifier:

```
prompt "turn off lights"  ->  gold entity "kitchen lights"
```

The model emitting `"lights"` is *faithful to the user* — and the agent already resolves it to the kitchen light at runtime via `resolve_target_in_graph` / `best_alias_score`. But the BFCL scorer exact-matches the **raw pre-resolution string**, so it marks a tool call wrong that the agent would execute correctly. That gap is the whole tool-name-vs-argument delta.

## Change (additive — headline metric untouched)

- **Grounded metric:** resolve both predicted and expected entity args through the *same* runtime resolver against a HA-Intents reference `HomeGraph`, and compare the resolved **`entity_ids` sets** instead of raw strings. `"lights"` ≡ `"kitchen lights"` ≡ `"kitchen light"` ≡ `"all lights"` → all `{light.kitchen}`; `"living room blinds"` stays distinct.
- New fields: `BfclCaseScore.grounded_argument_match`; `BfclReport.grounded_argument_accuracy` / `grounded_strict_accuracy` (+ match counts). Existing `strict_accuracy` / `argument_accuracy` are **unchanged**.
- `bfcl-score` / `bfcl-score-llm` print the grounded numbers alongside.
- Drive-by fix: `infer_domain` was missing `"fan"`, so fan queries didn't resolve by domain *at runtime* (a real agent gap, not just scoring). Added it.

Faithful reuse, not a fork: scoring runs the exact `resolve_target_in_graph` path the agent uses, so the metric reflects what the agent actually acts on.

## Results

- Unit tests: `"lights"`/`"kitchen lights"`/`"kitchen light"`/`"all lights"` unify; underspecified-entity case is `argument_match=false` but `grounded_argument_match=true`. 12/12 bfcl tests pass; existing `strict_accuracy == 1.0` test unaffected.
- Local quick-router (208 cases): argument `5.77%` → grounded `9.62%`.
- **Next:** re-score Qwen3-4B @ 4096 on the Jetson — the LLM emits underspecified entities far more than the deterministic router, so the lift there should be larger (raw argument was ~0.167). Will post numbers on #387.

## Decision needed

This reports grounded accuracy as an **additive column** and leaves the headline M1 `strict_accuracy` as raw exact-match. Once we have the Jetson Qwen delta, do we **promote** grounded matching to the headline M1 scorecard number, or keep both columns? That's a scorecard-semantics call I left to you rather than bake in.

## Reference home fixture

The `bfcl_reference_home()` fixture is dataset-specific (mirrors HA-Intents). Generalizing it (deriving the graph from each imported dataset) is noted as follow-up.

## Real Behavior Proof

**What I ran** (from `/home/speedy/genie-claw`):
- `cargo test -p genie-core --lib eval::bfcl` → `12 passed; 0 failed` (incl. the two new tests: entity-variant unification + grounded credit for an underspecified entity).
- `cargo test -p genie-core --lib infer_domain` → `1 passed` (fan addition doesn't regress domain inference).
- `RUSTFLAGS="-D warnings" cargo clippy -p genie-core -p genie-ctl --all-targets` → clean.
- `cargo run -p genie-ctl -- bfcl-score --cases tests/bfcl/local/ha_home_cases.jsonl --predictions tests/bfcl/local/ha_home_predictions.jsonl`

**What I observed** — the new grounded columns appear and lift the score without touching the raw headline:
```
cases:               208
argument_accuracy:   5.77%
grounded_arg_acc:    9.62%
strict_accuracy:     5.77%
grounded_strict_acc: 9.62%
```
The grounded metric credits cases where the model emits `"lights"` and the gold is `"kitchen lights"` (both resolve to `{light.kitchen}`), while `argument_accuracy`/`strict_accuracy` are unchanged. Next: re-score Qwen3-4B @ 4096 on the Jetson and post the LLM-side delta on #387.

- [x] I ran the change and verified the described behavior locally.

### DIFF
--- a/crates/genie-core/src/eval/bfcl.rs
+++ b/crates/genie-core/src/eval/bfcl.rs
@@ -18,8 +18,9 @@ use std::path::Path;
 use std::sync::Arc;
 
 use crate::ha::{
-    ActionResult, DeviceRef, HomeAction, HomeActionKind, HomeAutomationProvider, HomeGraph,
-    HomeState, HomeTarget, HomeTargetKind, IntegrationHealth, SceneRef,
+    ActionResult, AreaRef, DeviceRef, EntityRef, HomeAction, HomeActionKind, HomeAssistantProvider,
+    HomeAutomationProvider, HomeGraph, HomeState, HomeTarget, HomeTargetKind, IntegrationHealth,
+    SceneRef,
 };
 use crate::tools::dispatch::{ToolDef, ToolDispatcher};
 
@@ -76,6 +77,11 @@ pub struct BfclCaseScore {
     pub parse_success: bool,
     pub tool_name_match: bool,
     pub argument_match: bool,
+    /// Like `argument_match`, but entity-like string arguments are first
+    /// canonicalized through the runtime resolver against the BFCL reference
+    /// home, so that e.g. "lights" and "kitchen lights" compare equal when they
+    /// resolve to the same physical entities.
+    pub grounded_argument_match: bool,
     pub strict_match: bool,
     pub expected_tool_calls: Vec<ExpectedToolCall>,
     pub actual_tool_calls: Vec<ToolCall>,
@@ -90,13 +96,17 @@ pub struct BfclReport {
     pub parsed_cases: usize,
     pub tool_name_matches: usize,
     pub argument_matches: usize,
+    pub grounded_argument_matches: usize,
     pub strict_matches: usize,
+    pub grounded_strict_matches: usize,
     pub missing_predictions: usize,
     pub failure_count: usize,
     pub parse_accuracy: f64,
     pub tool_name_accuracy: f64,
     pub argument_accuracy: f64,
+    pub grounded_argument_accuracy: f64,
     pub strict_accuracy: f64,
+    pub grounded_strict_accuracy: f64,
     pub case_scores: Vec<BfclCaseScore>,
 }
 
@@ -137,9 +147,19 @@ pub fn score_cases(cases: &[BfclCase], predictions: &[BfclPrediction]) -> BfclRe
         .iter()
         .filter(|score| score.argument_match)
         .count();
+    let grounded_argument_matches = case_scores
+        .iter()
+        .filter(|score| score.grounded_argument_match)
+        .count();
     let strict_matches = case_scores
         .iter()
         .filter(|score| score.strict_match)
+        .count();
+    let grounded_strict_matches = case_scores
+        .iter()
+        .filter(|score| {
+            score.parse_success && score.tool_name_match && score.grounded_argument_match
+        })
         .count();
     let missing_predictions = case_scores
         .iter()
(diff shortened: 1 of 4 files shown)


ai-hpc:
On-device validation (full 208, Qwen3-4B @4096): raw strict **20.19%** → grounded **70.19%**, but auditing the 104 flips found **23 wrong-room false credits** (model hallucinates `upstairs_lights`/`living_room_light`; the minimal reference home collapses them to the only kitchen entity). Defensible grounded ≈ **59.13%**. Before this grounded column is headline-trustworthy, `bfcl_reference_home()`/the grounding needs a guard so an unresolvable room qualifier does **not** fall back to the whole-home domain group. Full decomposition on #387. Keeping this PR as the additive infrastructure; the fidelity fix is the follow-up.

PR #399, opened 2026-06-06
feat(bfcl): ground predict prompt in home device catalog — raw 20%→51%, grounded 72%→83% on Jetson

ai-hpc:
Grounds the BFCL predict prompt in the home's **device catalog** — the single biggest on-device accuracy lever found so far.

## Why

The predict prompt gave the model the tool schema but **not the home's device list**, so "turn off lights" forced it to guess an entity name ("lights", "upstairs lights") it could not see. A real on-device agent always knows its devices; withholding them made the eval harder than reality and is the root of low entity accuracy + wrong-room hallucinations. This injects the reference home's devices (compact, lowercase to match the dataset's canonical surface forms) and instructs the model to use those exact names, map vague references to them, never invent rooms, and use the schema's exact action values.

## Result — on-device, Qwen3-4B @ 4096, full 208 HA-Intents cases

| metric | baseline | this PR | Δ |
|---|---|---|---|
| tool-name | 98.08% | 97.12% | −1.0 |
| **raw strict** | **20.19%** | **50.96%** | **+30.8pp (2.5×)** |
| **grounded strict** | **72.12%** | **82.69%** | **+10.6pp** |

The headline raw number more than doubles; grounded reaches ~83%. Accuracy from deterministic device-state grounding, not model scale.

## Design notes (honest)

- **Compact + lowercase by necessity.** An earlier verbose, display-cased version *raised grounded* on a 40-case subset (82.5→87.5%) but **tanked raw** (model emitted "Kitchen Lights" vs gold "kitchen lights") and the long prompt blew past the runtime's prefix-cache window (60s prefill/case). Lowercasing fixed raw; compacting kept the prompt short.
- **Latency caveat.** The catalog lives in the *stable system prompt*, so with prefix caching it costs ~0 at runtime. Without reuse (as the eval currently runs), each call re-prefills ~509 tokens (~16s). Follow-up: wire/verify prefix-cache reuse so a live command stays ~1–3s.
- **Production follow-up.** This grounds the BFCL eval prompt (genie-ctl); the same device-catalog injection should land in the real agent prompt assembly (genie-core, sourced from the live HA device graph) so users get the same gain.

## Real Behavior Proof

**What I ran:** cross-built `genie-ctl` for aarch64 on the build host, deployed to the Jetson Orin Nano, ran `bfcl-predict-llm` (Qwen3-4B via genie-ai-runtime, 208 cases) with the new prompt, then `bfcl-score` against the baseline predictions.
**What I observed:** the table above — raw strict 20.19% → **50.96%**, grounded 72.12% → **82.69%**, on-device.

- [x] Verified end-to-end on Jetson Orin Nano 8GB hardware (prediction + scoring).

### DIFF
--- a/crates/genie-core/src/eval/bfcl.rs
+++ b/crates/genie-core/src/eval/bfcl.rs
@@ -532,6 +532,22 @@ fn bfcl_reference_home() -> HomeGraph {
     }
 }
 
+/// Human-readable device catalog for the BFCL reference home: one
+/// `"<Device Name> (<Area>)"` line per controllable entity. A real on-device
+/// agent always knows its home's devices; feeding this catalog into the eval
+/// prompt grounds entity arguments in real device state (the GeniePod thesis)
+/// instead of forcing the model to guess names it cannot see.
+pub fn bfcl_reference_home_device_catalog() -> Vec<String> {
+    bfcl_reference_home()
+        .entities
+        .iter()
+        .map(|entity| match entity.area.as_deref() {
+            Some(area) => format!("{} ({})", entity.name, area),
+            None => entity.name.clone(),
+        })
+        .collect()
+}
+
 /// Deep-clone `args`, replacing any entity-like string argument with a stable
 /// canonical token derived from the entities the runtime resolver would act on.
 ///
--- a/crates/genie-core/src/eval/bfcl.rs
+++ b/crates/genie-core/src/eval/bfcl.rs
@@ -532,19 +532,18 @@ fn bfcl_reference_home() -> HomeGraph {
     }
 }
 
-/// Human-readable device catalog for the BFCL reference home: one
-/// `"<Device Name> (<Area>)"` line per controllable entity. A real on-device
-/// agent always knows its home's devices; feeding this catalog into the eval
-/// prompt grounds entity arguments in real device state (the GeniePod thesis)
-/// instead of forcing the model to guess names it cannot see.
+/// Controllable device names for the BFCL reference home, lowercased to match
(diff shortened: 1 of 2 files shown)


ai-hpc:
## On-device validation with the prefix-cache runtime (genie-ai-runtime #85)

Re-ran the full 208-case BFCL on the Jetson Orin Nano with this PR's device-grounded prompt against the **updated genie-ai-runtime** (in-memory prefix cache now merged + live):

| run | raw strict | grounded | wall-clock | prefill/case |
|---|---|---|---|---|
| baseline prompt | 20.19% | 72.12% | ~48 min | ~16s |
| this PR (no prefix cache) | 50.96% | 82.69% | ~62 min | ~16s |
| **this PR + prefix cache** | **51.44%** | **83.65%** | **20.2 min** | **~1s** (485 tok reused) |

- **Accuracy:** raw strict **20.19% → 51.44%** (2.5×), grounded **72.12% → 83.65%** — same Qwen3-4B, same 4096 context, no model scale increase. Pure deterministic device-state grounding.
- **Latency is no longer a cost:** the grounding prompt lives in the stable system prefix, so the runtime computes it once and reuses it — per-case prefill **16s → ~1s** (`kv_reused=485`), full eval **62 → 20 min**.
- **Reuse doesn't corrupt output:** accuracy is identical to the no-cache run (51.44% vs 50.96% is run-to-run noise), confirming the KV reuse is correct.

Resolves the accuracy-vs-latency tension on-device: the device catalog that lifts accuracy is effectively free at runtime once prefix-cached.

PR #362, opened 2026-06-01
fix(tools): validate memory_recall query at execution boundary

kiannidev:
Fixes #361

## Summary

`memory_recall` marked `query` as required in `ToolDef`, but missing arguments silently defaulted to broad `"user"` recall via `memory_query()`. This PR adds `parse_memory_recall_query()` with non-empty string validation (including legacy `topic`/`what` aliases) and rejects invalid calls before memory lookup.

## Changes

- Add `parse_memory_recall_query()` and `normalize_memory_recall_query()`.
- Validate query before locking memory in `exec_memory_recall()`.
- Remove silent `unwrap_or("user")` default.
- Add `memory_recall_rejects_invalid_arguments_and_audits` integration test.
- Add `memory_recall_accepts_topic_alias_after_schema_validation` unit test.

## Real Behavior Proof

- [x] I have built and run the affected code locally (or noted why I could not).
- [ ] I have verified the change end-to-end on Jetson hardware.
- [x] I have NOT verified on Jetson hardware, and I explain the equivalent verification path or validation gap below.

Tested profile / hardware (check all that apply):

- [ ] `jetson`
- [ ] `raspberry_pi`
- [ ] `portable_sbc`
- [x] `laptop`
- [ ] `mac`
- [ ] CI-only / docs-only
- [ ] Not run locally

### What I ran

On x86_64 Linux dev host (`laptop` profile), branch `fix/issue-361-memory-recall-query-validation`:

```bash
cd genie-claw
cargo fmt -p genie-core
cargo test -p genie-core memory_recall
cargo test -p genie-core --test tool_gate_integration_test memory_recall_rejects_invalid_arguments_and_audits
cargo test -p genie-core --test tool_gate_integration_test
cargo clippy -p genie-core -- -D warnings
```

### What I observed

- `memory_recall_rejects_invalid_arguments_and_audits` passes for `{}`, empty query, and non-string query.
- Existing `memory_recall_*` unit tests pass; `topic` alias still works after validation.
- Full `tool_gate_integration_test` suite passes; clippy clean.

Jetson validation gap: schema enforcement verified via unit/integration tests on dev host.

## Test plan

- [x] `cargo test -p genie-core memory_recall`
- [x] `cargo test -p genie-core --test tool_gate_integration_test memory_recall_rejects_invalid_arguments_and_audits`
- [x] `cargo test -p genie-core --test tool_gate_integration_test`
- [x] `cargo clippy -p genie-core -- -D warnings`

## Notes for reviewers

[...]

PR #413, opened 2026-06-13
fix(tools): reject invalid memory_forget query at execution boundary

jaso0n0818:
## Summary

`memory_forget` validated its `query` argument too leniently and could run a delete on invalid input instead of rejecting it. This hardens the argument at the execution boundary, matching the read-side sibling `exec_memory_recall` (PR #362) and the recent `play_media` / `calculate` / `set_timer` argument-validation fixes.

## Changes

- Add `parse_memory_forget_query`, which trims the `query` argument and rejects missing / empty / whitespace-only / non-string values with `memory_forget requires non-empty string argument 'query'` — the same idiom as `parse_play_media_query` and `parse_memory_recall_query`.
- Call it at the top of `exec_memory_forget`, before locking the memory backend, replacing `args.get("query").and_then(|v| v.as_str()).unwrap_or("")` and the soft `"Please specify what to forget."` early return. Two inputs previously slipped through on this destructive path:
  - a non-string `query` (e.g. `123`) was silently coerced to `""` and returned a soft success instead of being rejected and audited;
  - a whitespace-only `query` (e.g. `"   "`) passed the `is_empty()` check and reached `mem.search("   ", 10)`.
- Add a tool-gate integration test, `memory_forget_rejects_invalid_arguments_and_audits`, asserting all four invalid shapes (`{}`, `""`, `"   "`, `123`) fail and are tool-audited with `success=false`.

No behavior change for valid calls: a real `query` still trims and forgets exactly as before.

## Real Behavior Proof

- [ ] `jetson`
- [ ] `raspberry_pi`
- [ ] `portable_sbc`
- [x] `laptop`
- [ ] `mac`
- [ ] CI-only / docs-only
- [ ] Not run locally

### What I ran

`x86_64-unknown-linux-gnu` laptop, Rust stable 1.96.0, from the repo root:

```
cargo fmt --all -- --check
cargo test -p genie-core --test tool_gate_integration_test --locked
cargo test  -p genie-core -p genie-ctl --no-default-features --all-targets --locked
```

This change is pure argument parsing (`serde_json` + `anyhow`) with no Jetson-, ALSA-, or hardware-specific code path, so the affected boundary is fully exercised by the `tool_gate_integration_test` dispatcher tests. The `aarch64-unknown-linux-gnu` release build is left to the `Cross-compile (aarch64 / Jetson)` CI job; nothing in the diff is target-conditional.

### What I observed

- `cargo fmt --all -- --check`: clean.
- `cargo clippy --workspace --all-targets --locked -- -D warnings`: clean (0 warnings).
- `tool_gate_integration_test`: **17 passed; 0 failed**, including the new `memory_forget_rejects_invalid_arguments_and_audits` and the unchanged `memory_recall_rejects_invalid_arguments_and_audits` (no regression). Each of the four invalid `memory_forget` calls returned `success=false` and produced one tool-audit event with `"tool":"memory_forget"`, `"success":false`; the seeded `"User's name is Jared"` row was never reached by the delete path.
- `--no-default-features` clippy: clean; `--no-default-features` tests: **39 passed; 0 failed**.

## Test plan

1. `cargo test -p genie-core --test tool_gate_integration_test memory_forget`
2. Optionally, on a device with memory configured, call the `memory_forget` tool with `{}`, `{"query": "   "}`, and `{"query": 123}` from the dashboard and confirm each is rejected and shows up in `runtime/tool-audit.jsonl` with `success:false`, while a valid `{"query": "shopping"}` still forgets matching memories.

## Notes for reviewers

Scope is intentionally one tool. This mirrors `exec_memory_recall`'s validation so the read and delete sides of the memory tools reject the same argument shapes; it does not touch the `MemoryReadContext` delete-gating (PR #201) that already sits below it.

### DIFF
--- a/crates/genie-core/src/tools/dispatch.rs
+++ b/crates/genie-core/src/tools/dispatch.rs
@@ -127,6 +127,14 @@ fn normalize_memory_recall_query(raw: &str) -> String {
     }
 }
 
+fn parse_memory_forget_query(args: &serde_json::Value) -> Result<&str> {
+    args.get("query")
+        .and_then(|v| v.as_str())
+        .map(str::trim)
+        .filter(|value| !value.is_empty())
+        .ok_or_else(|| anyhow::anyhow!("memory_forget requires non-empty string argument 'query'"))
+}
+
 fn parse_calculate_expression(args: &serde_json::Value) -> Result<&str> {
     args.get("expression")
         .and_then(|v| v.as_str())
@@ -1375,6 +1383,14 @@ impl ToolDispatcher {
         args: &serde_json::Value,
         exec_ctx: ToolExecutionContext,
     ) -> Result<String> {
+        // Validate the argument at the execution boundary, before touching the
+        // memory backend, the same way exec_memory_recall does (PR #362). The
+        // previous `unwrap_or("")` silently coerced a missing or non-string
+        // `query` into "", and a whitespace-only query slipped past the
+        // `is_empty()` check straight into `mem.search("   ", ...)` — running a
+        // destructive forget on garbage input instead of being rejected and
+        // audited like its read-side sibling.
+        let query = parse_memory_forget_query(args)?;
         let mem = self
             .memory
             .as_ref()
@@ -1382,11 +1398,6 @@ impl ToolDispatcher {
         let mem = mem
             .lock()
             .map_err(|e| anyhow::anyhow!("memory lock: {}", e))?;
-        let query = args.get("query").and_then(|v| v.as_str()).unwrap_or("");
-
-        if query.is_empty() {
-            return Ok("Please specify what to forget.".to_string());
-        }
 
         // Gate deletes through the same MemoryReadContext that exec_memory_recall
         // uses. Without it, an LLM that cannot READ a person-scoped row could
--- a/crates/genie-core/tests/tool_gate_integration_test.rs
+++ b/crates/genie-core/tests/tool_gate_integration_test.rs
@@ -743,6 +743,82 @@ async fn memory_recall_rejects_invalid_arguments_and_audits() {
 }
 
 #[tokio::test]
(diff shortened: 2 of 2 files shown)


jaso0n0818:
Rebased onto latest `main` (fork merge-upstream + force-push). Compare is now **ahead 1 / behind 0**. Local `cargo test -p genie-core --test tool_gate_integration_test` passes on this branch. Ready for maintainer review.

ai-hpc (approved):
The strongest of the recent typed-tool fixes — `memory_forget` is *destructive*, and this closes a real hazard: a whitespace-only `query` (`"   "`) previously slipped past the `is_empty()` check straight into `mem.search("   ", …)`, i.e. a forget running on garbage input. `parse_memory_forget_query()` trims before the empty check (so `"   "` is rejected) and validates at the execution boundary, before acquiring the memory lock — matching the read-side `exec_memory_recall` (#362). Clean and consistent with the `get_weather` / `web_search` / `play_media` / `calculate` siblings.

**Verified on-device (Jetson Orin Nano, aarch64):** built genie-core from this head (`55385f7`) and ran the full `tool_gate_integration_test` suite — **18 passed, 0 failed**, including `memory_forget_rejects_invalid_arguments_and_audits` (all four cases: missing / empty / whitespace-only / non-string). Did this because GitHub CI is gated — see below.

Two notes:
- **CI gate (maintainer action):** only the PR-body checklist has run; the CI + aarch64 cross-compile workflows are `action_required` (first-time contributor), which is why merge is `BLOCKED`. They need an "Approve and run workflows" before the gate clears — the on-device run above stands in until then.
- **Minor (non-blocking):** the test comment says "the stored memory must survive every rejected call," but it only asserts the audit events — it never re-queries to confirm the entry persists. A `memory.search(...)`-still-present assertion after the rejected calls would make that guarantee explicit.

LGTM.

PR #424, opened 2026-06-20
fix(tools): reject malformed optional get_weather forecast and web_search limit at execution boundary

### LINKED ISSUE #422: [bug] tools/dispatch: get_weather forecast / web_search limit silently coerce malformed optional args
## Summary

Two optional tool arguments accept a **provided-but-malformed value** and silently coerce it to a default instead of rejecting it at the execution boundary:

- `get_weather.forecast` — a non-boolean value is dropped to `false`. A small model emitting `"forecast": "true"` (a stringified bool, a form small models routinely emit) returns **current weather when the user asked for the 7-day forecast**.
- `web_search.limit` — a non-integer value is dropped to the default `3`. A `"limit": "5"`, `2.5`, or `-1` quietly returns a **different result count than the caller asked for**.

This is the same silent-coercion class already closed for the **required** args of these tools (#412 web_search query, #409 get_weather location) and for `home_control`'s optional `value` in **#414** — now on the *optional* args of the same two tools. #414 fixed only the numeric `value`; it did nothing for these.

## Steps to reproduce

Traced in the dispatcher source against `main` (`6ea5197`). Both values are read with `and_then(as_*).unwrap_or(default)`, which turns any wrong-typed value into the default rather than an error:

```
get_weather  {"location":"Denver","forecast":"true"}  => returns CURRENT weather (forecast silently false)
get_weather  {"location":"Denver","forecast":1}       => returns CURRENT weather
web_search   {"query":"...","limit":"5"}              => returns 3 results (limit silently default)
web_search   {"query":"...","limit":2.5}              => returns 3 results
web_search   {"query":"...","limit":-1}               => returns 3 results
```

Root cause in `crates/genie-core/src/tools/dispatch.rs`:

```rust
// exec_weather
let forecast = args.get("forecast").and_then(|v| v.as_bool()).unwrap_or(false);

// exec_web_search
let limit = args.get("limit").and_then(|v| v.as_u64()).unwrap_or(3).clamp(1, 5) as usize;
```

`as_bool()` / `as_u64()` return `None` for any wrong-typed value (string, float, negative), and `unwrap_or(default)` then erases the distinction between "absent" and "supplied but malformed".

## Expected vs actual

- **Actual:** a provided malformed `forecast` / `limit` is silently coerced to the default; the user's intent (forecast vs current; N results) is lost with no signal, and nothing is audited as a failure.
- **Expected:** an *absent or null* value stays the optional no-op default, but a *provided* value of the wrong type is **rejected at the dispatch boundary** (`get_weather 'forecast' must be a boolean when provided`, `web_search 'limit' must be an integer when provided`) and audited as a failure — consistent with #414's treatment of `home_control` `value`.

## Suggested fix

Add boundary parse helpers mirroring `parse_home_control_args`:

- `parse_get_weather_forecast`: `None`/`null` → `false`; a real bool → its value; anything else → reject.
- `parse_web_search_limit`: `None`/`null` → `3`; a valid integer → clamp into `1..=5`; a non-integer → reject.

A valid integer outside `1..=5` still clamps (preserving prior lenient behavior for in-type values); only wrong-typed values are rejected.

> Adjacent (optional, separate, deliberately out of scope here): `web_search.fresh` uses the same `as_bool().unwrap_or(false)` pattern, and the `/api/web-search` HTTP endpoint in `server.rs` coerces `limit`/`fresh` the same way. Both are different surfaces from the tool execution boundary and can follow as their own change.

## Scope / M1

Tool-call correctness + BFCL strict-arg scoring on the 4096-token Jetson path (M1). Same silent-coercion class the reliability series has been closing (#409, #412, #414), now on the optional args of `get_weather` and `web_search`. Found via dispatcher source review.

jakearmstrong59:
## Summary

`get_weather.forecast` and `web_search.limit` silently coerced a provided-but-malformed value to a default instead of rejecting it: a stringified `"forecast": "true"` returned current weather when the user asked for the 7-day forecast, and a `"limit": "5"` / `2.5` / `-1` quietly fell back to 3 results. This rejects wrong-typed optional values at the dispatch boundary and audits them, the same way #414 handles `home_control`'s `value`; absent/null stays the optional default and a valid `limit` still clamps to `1..=5`. Closes #422.

## Changes

- `crates/genie-core/src/tools/dispatch.rs`: added `parse_get_weather_forecast` (None/null → false; real bool → value; else reject) and `parse_web_search_limit` (None/null → 3; valid integer → clamp `1..=5`; non-integer → reject), wired into `exec_weather` / `exec_web_search`.
- `crates/genie-core/src/tools/dispatch.rs`: two unit tests (`get_weather_forecast_must_be_boolean_when_provided`, `web_search_limit_must_be_integer_when_provided`) modeled on `home_control_value_must_be_numeric_when_provided`.
- `crates/genie-core/tests/tool_gate_integration_test.rs`: extended `web_search_rejects_invalid_arguments_and_audits` and `get_weather_rejects_invalid_arguments_and_audits` with the malformed-optional-arg cases (each rejection is audited as a failure).
- No internal caller is affected: `quick.rs`, `genie-ctl`, and the voice path already pass `forecast` as a bool and `limit` as an integer. Out of scope (separate surfaces): `web_search.fresh` and the `/api/web-search` HTTP endpoint in `server.rs`.

## Real Behavior Proof

- [x] `jetson`
- [ ] `raspberry_pi`
- [ ] `portable_sbc`
- [x] `laptop`
- [ ] `mac`
- [ ] CI-only / docs-only
- [ ] Not run locally

### What I ran

Built genie-core from source and ran the affected tests **on the target Jetson**: NVIDIA Jetson Orin Nano (Engineering Reference Developer Kit Super, 8 GB), L4T R36.4.7, `Linux 5.15.148-tegra aarch64`, `rustc 1.95.0`, debug profile. This change is pure tool-dispatch argument validation (JSON → typed value) with no model, voice, audio, Home Assistant, or dashboard subsystem, so the dispatcher unit tests and the tool-gate integration tests ARE the affected code path — running them on the Orin Nano exercises the fix on the real aarch64/L4T target. Cloned the PR branch fresh (`11a795f`) and built from scratch on the device:

```
# on the Jetson Orin Nano — aarch64, L4T R36.4.7
cargo test -p genie-core --lib -- \
  get_weather_forecast_must_be_boolean_when_provided \
  web_search_limit_must_be_integer_when_provided

cargo test -p genie-core --test tool_gate_integration_test -- \
  web_search_rejects_invalid_arguments_and_audits \
  get_weather_rejects_invalid_arguments_and_audits
```

Also on an x86_64 dev box (`rustc 1.95.0`): the same tests plus `cargo fmt -p genie-core -- --check` and `cargo clippy -p genie-core --tests` (both clean).

### What I observed

All four tests pass on the Jetson Orin Nano (genie-core built from scratch on-device: lib 2m13s, integration test binary 1m08s):

```
# unit — target/debug/deps/genie_core-19794b9c31c74abc
running 2 tests
test tools::dispatch::tests::get_weather_forecast_must_be_boolean_when_provided ... ok
test tools::dispatch::tests::web_search_limit_must_be_integer_when_provided ... ok
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 718 filtered out

# integration — tests/tool_gate_integration_test.rs
running 2 tests
test get_weather_rejects_invalid_arguments_and_audits ... ok
test web_search_rejects_invalid_arguments_and_audits ... ok
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 17 filtered out
```

The integration tests assert each malformed optional arg is rejected at the dispatch boundary AND written to the tool-audit log as `success: false`. Behavior at the boundary:

| Call | Before | After |
| --- | --- | --- |
| `get_weather {"location":"Denver","forecast":"true"}` | current weather (silent) | rejected + audited |
| `get_weather {"location":"Denver","forecast":true}` | forecast | forecast (unchanged) |
| `web_search {"query":"x","limit":"5"}` | 3 results (silent) | rejected + audited |
| `web_search {"query":"x","limit":99}` | 5 results (clamped) | 5 results (clamped, unchanged) |

Scope note: this is the on-device test run on the real aarch64/L4T target. The change has no model/voice/HA/dashboard surface, so there is no separate running-service path to exercise beyond these tests.

## Test plan

- `cargo test -p genie-core --lib -- get_weather_forecast_must_be_boolean_when_provided web_search_limit_must_be_integer_when_provided`
- `cargo test -p genie-core --test tool_gate_integration_test -- web_search_rejects_invalid_arguments_and_audits get_weather_rejects_invalid_arguments_and_audits`
- Optional manual: issue `get_weather` with `"forecast":"true"` and `web_search` with `"limit":"5"` and confirm each is rejected with the boundary error and written to the tool-audit log as `success:false`.

### DIFF
--- a/crates/genie-core/src/tools/dispatch.rs
+++ b/crates/genie-core/src/tools/dispatch.rs
@@ -173,12 +173,49 @@ fn parse_web_search_query(args: &serde_json::Value) -> Result<&str> {
         .ok_or_else(|| anyhow::anyhow!("web_search requires non-empty string argument 'query'"))
 }
 
+/// `limit` stays optional — an absent or null value defaults to 3 results. But a
+/// *provided* value must be a non-negative integer. The old
+/// `args.get("limit").and_then(|v| v.as_u64()).unwrap_or(3)` silently coerced a
+/// malformed limit (a stringified `"5"`, a float `2.5`, or a negative) to the
+/// default 3, quietly returning a different result count than the caller asked
+/// for. Reject the malformed value at the boundary the same way home_control
+/// rejects a non-numeric `value` (PR #414); a valid integer outside 1..=5 still
+/// clamps into range rather than erroring.
+fn parse_web_search_limit(args: &serde_json::Value) -> Result<usize> {
+    match args.get("limit") {
+        None | Some(serde_json::Value::Null) => Ok(3),
+        Some(provided) => {
+            let limit = provided.as_u64().ok_or_else(|| {
+                anyhow::anyhow!("web_search 'limit' must be an integer when provided")
+            })?;
+            Ok(limit.clamp(1, 5) as usize)
+        }
+    }
+}
+
 fn parse_get_weather_location(args: &serde_json::Value) -> Result<&str> {
     args.get("location")
         .and_then(|v| v.as_str())
         .map(str::trim)
         .filter(|value| !value.is_empty())
         .ok_or_else(|| anyhow::anyhow!("get_weather requires non-empty string argument 'location'"))
+}
+
+/// `forecast` stays optional — an absent or null value means current weather
+/// (the no-op default). But a *provided* value must be a real boolean. The old
+/// `args.get("forecast").and_then(|v| v.as_bool()).unwrap_or(false)` silently
+/// coerced a stringified `"true"` (a form small models routinely emit) to
+/// `false`, returning current weather when the user asked for the 7-day
+/// forecast. Reject the malformed value at the boundary the same way
+/// home_control rejects a non-numeric `value` (PR #414).
+fn parse_get_weather_forecast(args: &serde_json::Value) -> Result<bool> {
+    match args.get("forecast") {
+        None | Some(serde_json::Value::Null) => Ok(false),
+        Some(serde_json::Value::Bool(value)) => Ok(*value),
+        Some(_) => Err(anyhow::anyhow!(
+            "get_weather 'forecast' must be a boolean when provided"
+        )),
+    }
 }
 
 /// Tool definition for LLM function calling.
@@ -2107,10 +2144,7 @@ fn exec_calculate(args: &serde_json::Value) -> Result<String> {
 
 async fn exec_weather(args: &serde_json::Value) -> Result<String> {
     let location = parse_get_weather_location(args)?;
-    let forecast = args
-        .get("forecast")
-        .and_then(|v| v.as_bool())
-        .unwrap_or(false);
+    let forecast = parse_get_weather_forecast(args)?;
 
     if forecast {
         super::weather::get_forecast(location).await
@@ -2121,11 +2155,7 @@ async fn exec_weather(args: &serde_json::Value) -> Result<String> {
 
 async fn exec_web_search(args: &serde_json::Value, config: &WebSearchConfig) -> Result<String> {
     let query = parse_web_search_query(args)?;
-    let limit = args
-        .get("limit")
-        .and_then(|v| v.as_u64())
-        .unwrap_or(3)
-        .clamp(1, 5) as usize;
+    let limit = parse_web_search_limit(args)?;
     let fresh = args
         .get("fresh")
         .or_else(|| args.get("cache_bypass"))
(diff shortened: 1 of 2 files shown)


matedev01:
**Verified on-device (Jetson Orin Nano, aarch64):** built this head (`11a795f`) and ran `cargo test -p genie-core` — all green (lib **720 passed, 0 failed** + integration suites 0 failed), including the two new parser tests (`get_weather_forecast_must_be_boolean_when_provided`, `web_search_limit_must_be_integer_when_provided`) and `routes_forecast_with_explicit_location`.

PR #425, opened 2026-06-20
fix(tools): reject missing value for home_control set_brightness/set_temperature at execution boundary

### LINKED ISSUE #421: [bug] tools/dispatch: home_control set_brightness/set_temperature apply a silent default when value is missing
## Summary

The two value-requiring `home_control` actions — `set_brightness` and `set_temperature` — accept a call with **no `value`** and silently substitute a hardcoded default (brightness → **50**, temperature → **20**), then report success. An under-specified request like "set the thermostat" or "change the brightness" (no number) is **executed against the device at an arbitrary default** instead of being rejected so the agent can ask for the missing value.

This is the same silent-default class already fixed for **#359** (set_timer → 60s), **#392** (calculate → 0), and **#391** (play_media), now on the **actuation** path. It is **not** covered by #414, which rejects a *non-numeric* value but does nothing for a *missing* one.

## Steps to reproduce

Live-probed against `main` (`e4430f6`) on a Jetson Orin Nano (aarch64) via the `tool_gate_integration_test` harness (`execute_with_context`, `RequestOrigin::Dashboard`):

```
home_control {"entity":"kitchen light","action":"set_brightness"}    => success=true  "Executed SetBrightness"
home_control {"entity":"thermostat",  "action":"set_temperature"}    => success=true  "Executed SetTemperature"
```

Root cause in `crates/genie-core/src/ha/provider.rs`:

```rust
HomeActionKind::SetBrightness => {
    let brightness = normalize_brightness(action.value.unwrap_or(50.0));   // missing -> 50
    ...
}
HomeActionKind::SetTemperature => {
    let temp = action.value.unwrap_or(20.0);                              // missing -> 20
    ...
}
```

`parse_home_control_args` returns `value: None` when the key is absent, `home::control` passes `None` straight through, and the provider's `.unwrap_or(...)` turns "no value given" into a concrete actuation.

## Expected vs actual

- **Actual:** missing value → silent default (brightness 50 / temp 20) → success; the user's under-specified request is executed at an arbitrary value with no signal that a value was never supplied.
- **Expected:** a value-requiring action (`set_brightness`, `set_temperature`) with no numeric `value` should be **rejected at the dispatch boundary** (e.g., `set_brightness requires numeric argument 'value'`) and audited as a failure, so the agent asks for the number instead of guessing — consistent with #359 / #392 / #391.

## Suggested fix

In `parse_home_control_args` (or the exec boundary): when `action ∈ {set_brightness, set_temperature}`, require a numeric `value`; reject + audit when absent or non-numeric, and drop the provider-side `.unwrap_or(50.0)` / `.unwrap_or(20.0)` fallbacks. Complements #414's non-numeric rejection.

> Adjacent (optional, separate): `set_temperature` passes `value` to `climate.set_temperature` with no range check, so `value: 9999` / `-5` are accepted as-is (brightness is clamped via `normalize_brightness`, temperature is not).

## Scope / M1

Actuation correctness + BFCL strict-arg scoring on the 4096-token Jetson path (M1). Same silent-default class the reliability series has been closing, on the highest-stakes (actuation) tool. Found via live on-device edge-case probing.

jakearmstrong59:
## Summary

`set_brightness` and `set_temperature` actuate a numeric setpoint, but a `home_control` call with **no `value`** was silently defaulted by the provider (brightness → 50, temperature → 20) and reported success — executing an arbitrary setpoint the user never asked for. An under-specified request like "set the thermostat" got actuated at a guess instead of being rejected so the agent can ask for the number. This rejects a missing `value` for those two actions at the dispatch boundary and audits it as a failure, complementing #414 (which only rejects a *non-numeric provided* value). Closes #421.

## Changes

- `crates/genie-core/src/tools/dispatch.rs`: `parse_home_control_args` now rejects `set_brightness`/`set_temperature` when `value` is absent or null (`home_control '<action>' requires a numeric argument 'value'`), via a small `action_requires_value` helper. Absent/null stays a no-op `None` for every other action (turn_on, toggle, open, lock, activate, …).
- `crates/genie-core/src/ha/provider.rs`: dropped the silent `action.value.unwrap_or(50.0)` / `unwrap_or(20.0)` fallbacks — a value-less setpoint now errors (`ok_or_else`) as defense in depth rather than actuating a hardcoded default.
- `crates/genie-core/src/tools/dispatch.rs`: unit test `home_control_set_actions_require_value` (absent/null rejected for both actions; value-bearing and non-value actions still parse).
- `crates/genie-core/tests/tool_gate_integration_test.rs`: extended `home_control_rejects_invalid_arguments_and_audits` with the two missing-value cases (rejected at the boundary, never reach the provider, audited `success: false`).

## Real Behavior Proof

- [x] `jetson`
- [ ] `raspberry_pi`
- [ ] `portable_sbc`
- [x] `laptop`
- [ ] `mac`
- [ ] CI-only / docs-only
- [ ] Not run locally

### What I ran

Built genie-core from source and ran the affected tests **on the target Jetson**: NVIDIA Jetson Orin Nano (Engineering Reference Developer Kit Super, 8 GB), L4T R36.4.7, `Linux 5.15.148-tegra aarch64`, `rustc 1.95.0`, debug profile. The change is pure tool-dispatch argument validation plus a provider guard, with no model/voice/HA/dashboard subsystem, so the dispatcher unit tests and the tool-gate integration test ARE the affected code path. Applied this change onto a fresh upstream-`main` clone (base `6ea5197`) and built from scratch on the device:

```
# on the Jetson Orin Nano — aarch64, L4T R36.4.7
cargo test -p genie-core --lib -- \
  home_control_set_actions_require_value \
  home_control_value_must_be_numeric_when_provided

cargo test -p genie-core --test tool_gate_integration_test -- \
  home_control_rejects_invalid_arguments_and_audits
```

Also on an x86_64 dev box (`rustc 1.95.0`): the same tests plus `cargo fmt -p genie-core -- --check` and `cargo clippy -p genie-core --tests` (both clean), and the provider tests (`ha::provider::tests`, no regression).

### What I observed

All tests pass on the Jetson Orin Nano (genie-core built from scratch on-device: lib 1m50s, integration test binary 51s):

```
# unit — target/debug/deps/genie_core-19794b9c31c74abc
running 2 tests
test tools::dispatch::tests::home_control_value_must_be_numeric_when_provided ... ok
test tools::dispatch::tests::home_control_set_actions_require_value ... ok
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 717 filtered out

# integration — tests/tool_gate_integration_test.rs
running 1 test
test home_control_rejects_invalid_arguments_and_audits ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 18 filtered out
```

The integration test asserts the missing-value calls are rejected at the boundary, never reach the home provider (`executed` stays empty), and are written to the tool-audit log as `success: false`. Behavior at the boundary:

| Call | Before | After |
| --- | --- | --- |
| `home_control {"entity":"kitchen light","action":"set_brightness"}` | success — brightness silently 50 | rejected + audited |
| `home_control {"entity":"thermostat","action":"set_temperature"}` | success — temperature silently 20 | rejected + audited |
| `home_control {"entity":"kitchen light","action":"set_brightness","value":60}` | brightness 60 | brightness 60 (unchanged) |
| `home_control {"entity":"kitchen light","action":"turn_on"}` | turn_on | turn_on (unchanged — no value needed) |

### Test plan

- `cargo test -p genie-core --lib -- home_control_set_actions_require_value home_control_value_must_be_numeric_when_provided`
- `cargo test -p genie-core --test tool_gate_integration_test -- home_control_rejects_invalid_arguments_and_audits`
- Optional manual: issue `home_control {"entity":"<light>","action":"set_brightness"}` with no `value` and confirm it is rejected with `home_control 'set_brightness' requires a numeric argument 'value'` and audited `success:false`, instead of dimming to 50%.

### Notes for reviewers

- Two layers: the boundary rejection (user-facing + audited) is the fix; dropping the provider `unwrap_or` defaults is defense in depth so no future caller can reach a silent default.
- Out of scope (noted in the issue as a separate follow-up): `set_temperature` still has no range check on a *provided* value (e.g. `9999` / `-5`); brightness is clamped via `normalize_brightness`, temperature is not.

### DIFF
--- a/crates/genie-core/src/ha/provider.rs
+++ b/crates/genie-core/src/ha/provider.rs
@@ -721,7 +721,13 @@ impl HomeAutomationProvider for HomeAssistantProvider {
             ),
             HomeActionKind::SetBrightness => {
                 ensure_domain(&domain, &["light"])?;
-                let brightness = normalize_brightness(action.value.unwrap_or(50.0));
+                // No silent default: a value-less set_brightness is rejected at
+                // the dispatch boundary (issue #421); error here too as defense
+                // in depth rather than actuating an arbitrary brightness.
+                let brightness =
+                    normalize_brightness(action.value.ok_or_else(|| {
+                        anyhow::anyhow!("set_brightness requires a numeric 'value'")
+                    })?);
                 (
                     "light".into(),
                     "turn_on",
@@ -730,7 +736,12 @@ impl HomeAutomationProvider for HomeAssistantProvider {
             }
             HomeActionKind::SetTemperature => {
                 ensure_domain(&domain, &["climate"])?;
-                let temp = action.value.unwrap_or(20.0);
+                // No silent default: a value-less set_temperature is rejected at
+                // the dispatch boundary (issue #421); error here too as defense
+                // in depth rather than actuating an arbitrary temperature.
+                let temp = action
+                    .value
+                    .ok_or_else(|| anyhow::anyhow!("set_temperature requires a numeric 'value'"))?;
                 (
                     "climate".into(),
                     "set_temperature",
--- a/crates/genie-core/src/tools/dispatch.rs
+++ b/crates/genie-core/src/tools/dispatch.rs
@@ -65,7 +65,23 @@ fn parse_home_control_args(args: &serde_json::Value) -> Result<(&str, &str, Opti
             anyhow::anyhow!("home_control 'value' must be a number when provided")
         })?),
     };
+    // set_brightness / set_temperature actuate a numeric setpoint, so a call with
+    // no `value` is under-specified. The provider used to substitute a hardcoded
+    // default (brightness 50 / temperature 20) and report success, actuating a
+    // setpoint the user never asked for. Reject the missing value at the boundary
+    // so the agent asks for the number instead of guessing — the same boundary
+    // #414 uses to reject a *non-numeric* value. (issue #421)
+    if value.is_none() && action_requires_value(action) {
+        anyhow::bail!("home_control '{action}' requires a numeric argument 'value'");
+    }
     Ok((entity, action, value))
+}
+
+/// Actions that actuate a numeric setpoint and therefore require a `value`.
+/// Every other action (turn_on, turn_off, toggle, open, close, lock, unlock,
+/// activate) is a no-op for `value` and leaves it `None`.
+fn action_requires_value(action: &str) -> bool {
+    matches!(action, "set_brightness" | "set_temperature")
 }
 
 /// Canonicalize a model-emitted action verb to one of [`HOME_CONTROL_ACTIONS`].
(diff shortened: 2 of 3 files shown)


ai-hpc:
Pushed two **test-only** commits on top of your fix (`28abbe8`, `be7c9ea`) — the fix itself is unchanged, just added coverage:

- **`action_requires_value_only_for_setpoint_actions`** (unit) — pins the contract exhaustively: only `set_brightness` / `set_temperature` require a `value`; the other eight actions (`turn_on` / `turn_off` / `toggle` / `open` / `close` / `lock` / `unlock` / `activate`) don't. The existing test exercised three of those; this guards against a future action being mis-classified.
- **`home_control_set_brightness_with_value_actuates`** (integration) — a valid value-bearing `set_brightness` still actuates end-to-end through the gate (success + `SetBrightness` executed + audited), so the new missing-value rejection can't over-fire on a valid setpoint.

ai-hpc:
Thanks @jakearmstrong59 — clean fix. Pushed two test commits (`28abbe8`, `be7c9ea`); when you get a sec, please verify they match your intent.

Issue #402, opened 2026-06-06
Contribution scope: performance (rewarded via gittensor) · real-HA tool-dispatch fixes (else closed)


ai-hpc:
## What we accept (and what we don't)

To keep review quality high while M1 is active, a PR is accepted **only** if it lands in one of these two buckets, with reproducible on-device proof. Anything else is closed.

> 💎 **Performance PRs are rewarded.** Land a performance-improvement PR that meets the rules below — measurable Jetson win, reproducible before→after proof — and you're eligible for a reward through [gittensor](https://gittensor.io/), the Bittensor subnet that pays out for merged open-source contributions.

### 1. Performance improvement
Measurable latency / throughput / memory wins on Jetson Orin Nano 8 GB, with before→after numbers.
- Example: [genie-ai-runtime#85](https://github.com/GeniePod/genie-ai-runtime/pull/85) — in-memory KV prefix cache, **~13× faster prefill** (16s → ~1s per command); cut the BFCL eval from ~62 min to ~20 min.

### 2. Tool-dispatch / real-Home-Assistant error fix
Correctness fixes in tool routing or home actuation, **reproduced against a real Home Assistant**. A runnable sample HA config is provided at [`deploy/homeassistant/`](https://github.com/GeniePod/genie-claw/tree/main/deploy/homeassistant) — bring it up, reproduce the failure, and prove the fix by showing the entity state change confirmed via the HA API.
- Examples: [#400](https://github.com/GeniePod/genie-claw/pull/400) — canonicalize `home_control` action synonyms (model emits `"turn off"`, runtime required `"turn_off"`), proven by a live `off → on` actuation; [#380](https://github.com/GeniePod/genie-claw/pull/380) — stop leaking unparsed tool-call JSON to the user.

---

**Required:** every such PR includes a **Real Behavior Proof** — what you ran, on what hardware, what changed. For HA fixes: live-HA before/after, confirmed via the API.

**Closed on sight:** anything outside these two buckets, prompt growth without a measured accuracy gain, generated/demo-only code, duplicate or churn PRs, or changes without reproducible proof.

This mirrors the README's [Accepted contribution scope](https://github.com/GeniePod/genie-claw#accepted-contribution-scope) (PR #401). Questions / better-collaboration ideas → Discord.

matedev01:
[...]

PR #431, opened 2026-06-20
perf(memory): batch recall-tracking writes into one transaction

jakearmstrong59:
## Summary

Every memory recall updates recall tracking (`accessed_ms`, `recall_count`, `max_score`, `query_hashes`) for each returned hit. That write was issued as **one auto-committed `UPDATE` per hit**, so a single `search` / `semantic_search` performed up to `limit` separate write transactions — i.e. `limit` fsyncs. On the Jetson Orin Nano's eMMC/SD storage those fsyncs dominate recall latency. Batching them into a single transaction collapses `limit` fsyncs into one. The per-row `UPDATE`s are byte-for-byte unchanged, so recall results are identical. Measured **~2.1× faster semantic recall (median 46.8 ms → 21.9 ms per recall)** on the Orin Nano, with tail latency cut from 57 ms to 22 ms. Contributes under #402 (performance bucket).

## Changes

- `crates/genie-core/src/memory/mod.rs`: new `record_recalls` wraps the per-result-set recall-tracking writes in one `unchecked_transaction` (one commit/fsync for the whole result set). `search`, `semantic_search`, and the `search_like_fallback` LIKE path now build a `(id, score)` list and call it instead of looping `update_recall_tracking` once per hit. Tracking stays best-effort — a failure logs and rolls back atomically without failing the recall (same as the previous per-row handling); `update_recall_tracking` itself is unchanged.
- `crates/genie-core/tests/semantic_recall_bench.rs`: ignored standalone benchmark over `Memory::semantic_search` for reproducible before→after timing.

## Real Behavior Proof

- [x] `jetson`
- [ ] `raspberry_pi`
- [ ] `portable_sbc`
- [x] `laptop`
- [ ] `mac`
- [ ] CI-only / docs-only
- [ ] Not run locally

### What I ran

On the target **Jetson Orin Nano** (Engineering Reference Developer Kit Super, 8 GB), L4T R36.4.7, `Linux 5.15.148-tegra aarch64`, `rustc 1.95.0`, **release** profile (`opt-level="z"`, LTO disabled only to speed the build). Applied the change and the benchmark onto a fresh upstream-`main` clone (base `d11250e`), built in a tmpfs target dir, and ran the ignored benchmark — which times `semantic_search` over 1000 embedded household memories (100 recalls, `limit` 10). The benchmark database lives on the SD card (`/tmp`), so real fsync cost is measured. Ran 3× per side:

```
# before = main (one auto-committed UPDATE per hit); after = this change (one transaction)
cargo test -p genie-core --release --test semantic_recall_bench -- --ignored --nocapture
```

Also on an x86_64 dev box (`rustc 1.95.0`): full `memory::` test suite (161 tests) green, `cargo fmt -p genie-core -- --check` and `cargo clippy -p genie-core --tests` clean. (Local SSD fsync is ~free, so the win is Jetson-storage-specific — hence the on-device numbers below.)

### What I observed

`semantic_search`, 1000 embedded memories, 100 recalls per run:

| run | before (per-hit autocommit) | after (one transaction) |
| --- | --- | --- |
| 1 | 33.7 ms/recall | 21.9 ms/recall |
| 2 | 57.1 ms/recall | 22.0 ms/recall |
| 3 | 46.8 ms/recall | 21.9 ms/recall |
| **median** | **46.8 ms** | **21.9 ms** |

- **~2.1× faster** per recall at the median; worst-case (tail) recall drops from 57 ms to 22 ms.
- The "after" numbers are stable to ±0.1 ms, while "before" swings 33–57 ms — the expected signature of collapsing many SD-card fsyncs (variable) into one.
- Recall results are unchanged (the per-row `UPDATE`s are identical); the 161-test `memory::` suite — including all `semantic_search_links_*` recall-quality tests — passes on both x86_64 and aarch64.

### Test plan

- `cargo test -p genie-core --lib memory::` (recall behavior unchanged)
- `cargo test -p genie-core --release --test semantic_recall_bench -- --ignored --nocapture` on `main` vs this branch to reproduce the table above.

### Notes for reviewers

- The benchmark exercises `semantic_search` only; per real turn the saving is larger because the lexical `search` path runs too and previously also issued one fsync per hit — both now batch.
- The remaining ~22 ms is the 1000-row embedding fetch + cosine, not fsync. A follow-up could cut that with a two-pass scan (score by id, then fetch full entries for the top-k only); kept out of scope here to keep this PR a single, measured change.

### DIFF
--- a/crates/genie-core/src/memory/mod.rs
+++ b/crates/genie-core/src/memory/mod.rs
@@ -832,16 +832,13 @@ impl Memory {
         scored.sort_by(|a, b| b.1.partial_cmp(&a.1).unwrap_or(std::cmp::Ordering::Equal));
         scored.truncate(limit);
 
-        // Update recall tracking for returned results.
-        for (entry, score) in &scored {
-            if let Err(error) = self.update_recall_tracking(entry.id, now, *score, &query_hash) {
-                tracing::error!(
-                    memory_id = entry.id,
-                    error = %error,
-                    "memory recall tracking update failed"
-                );
-            }
-        }
+        // Update recall tracking for returned results in one transaction (one
+        // fsync for the whole result set instead of one per hit).
+        let recalls: Vec<(i64, f64)> = scored
+            .iter()
+            .map(|(entry, score)| (entry.id, *score))
+            .collect();
+        self.record_recalls(&recalls, now, &query_hash);
 
         Ok(scored.into_iter().map(|(e, _)| e).collect())
     }
@@ -893,16 +890,11 @@ impl Memory {
         });
         entries.truncate(limit);
 
-        for entry in &entries {
-            let score = lexical_overlap_score(query, &entry.content);
-            if let Err(error) = self.update_recall_tracking(entry.id, now, score, query_hash) {
-                tracing::error!(
-                    memory_id = entry.id,
-                    error = %error,
-                    "memory recall tracking update failed"
-                );
-            }
-        }
+        let recalls: Vec<(i64, f64)> = entries
+            .iter()
+            .map(|entry| (entry.id, lexical_overlap_score(query, &entry.content)))
+            .collect();
+        self.record_recalls(&recalls, now, query_hash);
 
         Ok(entries)
     }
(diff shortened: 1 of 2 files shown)


matedev01:
Thnaks @jakearmstrong59 really appreciate your help on jetson orin nano, exact target device. If you have a sec, is it possible testing orin nx?
Thanks again.

PR #432, opened 2026-06-20
perf(voice/identity): build the Hann window once per fingerprint, not per band (~3.9x faster band powers)

jaso0n0818:
## Summary

Speaker-identification fingerprinting recomputed the Hann analysis window — a `cos()` per sample — once for every frequency band, and again for every frame. The window depends only on the frame length, which is constant across a fingerprint, so it can be built once and reused. This makes band-power extraction ~3.9x faster with bit-for-bit identical fingerprints.

## Changes

- `extract_fingerprint` builds the Hann window once (`hann_window`) and shares it with every frame's feature extraction.
- `band_powers` now takes the precomputed window, applies it once per frame (`sample * window[idx]`), and reuses the windowed frame across all 8 Goertzel bands.
- `goertzel_power` consumes already-windowed samples; its inner loop is now a pure Goertzel recurrence with no per-sample `cos()`.
- No behavior change: the per-sample product `sample * window` is identical, so band powers — and the resulting speaker fingerprint vector — are bit-for-bit unchanged.

## Real Behavior Proof

- [x] `laptop`

### What I ran

On an x86_64 Linux dev host (no Jetson available):

- `cargo fmt --all -- --check` — clean.
- `cargo clippy -p genie-core --all-targets --locked -- -D warnings` — clean (0 warnings).
- `cargo test -p genie-core --lib voice::identity` — **7 passed**, including `local_biometric_enrolls_and_identifies_matching_voice`, the end-to-end enroll → fingerprint → identify test that exercises this exact code path.
- A standalone `rustc -O` micro-benchmark reproducing `band_powers` old vs. new over a 512-sample frame (32 ms @ 16 kHz), 8 bands, 160 frames/fingerprint, comparing outputs with `f32::to_bits()` equality.

### What I observed

- **Bit-identical:** every band-power output matches the previous implementation bit-for-bit (`to_bits()` equality), so speaker fingerprints and match scores are unchanged. The enroll/identify round-trip unit test confirms identical matching behavior end-to-end.
- **Timing** (x86 laptop, mean over repeated runs of 160 frames/fingerprint × 2000 iterations):
  - old (window recomputed per band): **~9.3 ms / fingerprint**
  - new (window built once per fingerprint): **~2.4 ms / fingerprint**
  - **≈3.9x faster** band-power extraction.

This is the speaker-identification path invoked per voice interaction (`voice_loop.rs` → `SpeakerIdentity::identify` → `extract_fingerprint`). I could not test on Jetson hardware; the change is pure CPU arithmetic with provably bit-identical output, validated by the unit tests and the micro-benchmark above. The absolute win is larger on the Jetson's slower cores where the redundant `cos()` work dominates.

## Test plan

- `cargo test -p genie-core --lib voice::identity`
- Enroll a speaker via `genie-ctl` and identify against the same and a different voice; fingerprints and match scores are unchanged versus `main`.

## Notes for reviewers

Pure performance change, no behavior change. The window coefficient `0.5 - 0.5*cos(2*pi*idx/(len-1))` is unchanged — it is just computed once now instead of `8 bands x N frames` times. `band_powers` multiplies it in once per frame and each band's Goertzel recurrence runs over the shared windowed buffer.

### DIFF
--- a/crates/genie-core/src/voice/identity.rs
+++ b/crates/genie-core/src/voice/identity.rs
@@ -412,12 +412,16 @@ fn extract_fingerprint(audio: &WavAudio) -> anyhow::Result<Vec<f32>> {
 
     let frame_len = ((audio.sample_rate as f32 * 0.032).round() as usize).clamp(256, 2048);
     let hop = (frame_len / 2).max(128);
+    // Every frame has the same length, so the Hann window is identical across
+    // them — build it once and share it with every frame's feature extraction.
+    let window = hann_window(frame_len);
     let mut features = Vec::new();
     let mut cursor = 0usize;
     while cursor + frame_len <= voiced.len() && features.len() < 160 {
         features.push(frame_features(
             &voiced[cursor..cursor + frame_len],
             audio.sample_rate,
+            &window,
         ));
         cursor += hop;
     }
@@ -467,10 +471,10 @@ fn voiced_samples(samples: &[f32]) -> Vec<f32> {
         .collect()
 }
 
-fn frame_features(frame: &[f32], sample_rate: u32) -> Vec<f32> {
+fn frame_features(frame: &[f32], sample_rate: u32, window: &[f32]) -> Vec<f32> {
     let rms = (frame.iter().map(|s| s * s).sum::<f32>() / frame.len() as f32).sqrt();
     let zcr = zero_crossing_rate(frame);
-    let bands = band_powers(frame, sample_rate);
+    let bands = band_powers(frame, sample_rate, window);
     let total_power = bands.iter().sum::<f32>().max(1e-9);
 
     let mut out = Vec::with_capacity(3 + bands.len());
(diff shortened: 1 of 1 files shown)


matedev01:
**Verified on-device (Jetson Orin Nano, aarch64):** built head `16e5bee` and ran `cargo test -p genie-core --lib` — **729 passed, 0 failed**, including `local_biometric_enrolls_and_identifies_matching_voice` (the enroll/identify roundtrip — it would break if the window refactor shifted fingerprints) plus all other `voice::identity` tests. CI is fully green too (aarch64 release build + `cargo test`).

matedev01:
micro-benchmarked `band_powers` directly on the Orin (both paths reimplemented faithfully, with the window precomputed for the new path exactly as `extract_fingerprint` amortizes it). Output is **bit-for-bit identical** (`to_bits()`-equal) at 16 / 22.05 / 48 kHz, and the new path is **~2.5–2.8× faster** (codegen-flag dependent), consistent across all three frame sizes:

```
sr=16000  frame_len=512   OLD 47.8us  NEW 16.8us   2.8x   (2.5x with target-cpu=native)
sr=22050  frame_len=706   OLD 65.6us  NEW 23.2us   2.8x
sr=48000  frame_len=1536  OLD 141.6us NEW 50.2us   2.8x
```

PR #436, opened 2026-06-21
feat(llm/privacy-proxy): add opt-in cloud escalation with local_only …

### LINKED ISSUE #418: [feature] Optional privacy-preserving cloud escalation via PrivacyProxy v0.1 (on-device anonymising gateway)
## Summary

Proposal: let genie-claw **optionally escalate hard tasks to a strong *free* cloud model without breaking its privacy-first guarantee**, by routing through [PrivacyProxy](https://github.com/ai-hpc/PrivacyProxy) — a small on-device gateway (Rust, MIT) that anonymises confidential data locally, sends only placeholders to the cloud, and restores the real values in the response. Just tagged [v0.1.0](https://github.com/ai-hpc/PrivacyProxy/releases/tag/v0.1.0).

_Disclosure: I'm the author of PrivacyProxy, and its local-memory layer was adapted from genie-claw's design — so this is meant as a native fit, not a bolt-on._

## Why it fits genie-claw

genie-claw is local-first and **limited-context** by design — small on-device models (Gemma 4, Falcon-H1, Qwen3.5-4B @ 4–16k, per #410 / #415). Right default for latency + privacy, but some household tasks exceed what a 4B model at 4k can do. Today that forces a choice: degrade locally, or send raw household data to a cloud model and lose privacy-first.

PrivacyProxy is a third option — a **privacy-preserving escalation path**. The cloud model reasons over `__PERSON_1__`, `__DEVICE_1__`, `__LOC_1__`, never `Alex` / the nursery Nest / your address, and genie-claw gets the real answer back. Privacy-first holds even when the reasoning is rented.

## The design is already aligned

PrivacyProxy's `egress_policy` (`local_only` | `anonymized`) was adapted directly from genie-claw's memory **`spoken_policy`** — the same cloud/share-boundary idea — so the integration is conceptually native:
- a memory/datum marked **`local_only`** is *never* sent, even to the anonymising gateway (it only seeds the local detector, so its terms get masked everywhere);
- **`anonymized`** data may cross, but only as placeholders.

## Concrete integration (small, opt-in)

PrivacyProxy is a drop-in OpenAI **and** Anthropic-compatible HTTP endpoint, so genie-claw needs no protocol change:

1. Add an optional `cloud_escalation` backend in the harness pointing at `http://127.0.0.1:8080/v1` (PrivacyProxy), used only on local decline / low confidence / context overflow.
2. Seed PrivacyProxy's vocabulary from genie-claw's `local_only` household memory (member names, device/room labels) so household identifiers are masked deterministically.
3. Keep the policy gate: only `anonymized`-policy context is eligible for escalation; `local_only` stays on-device.

Net: genie-claw stays the privacy-first harness; PrivacyProxy is the enforced cloud boundary when you opt in.

Notes:
- **Clean process boundary** — the two talk over localhost HTTP as separate processes, so AGPL-3.0 / MIT don't entangle.
- PrivacyProxy ships a `pp-eval` harness measuring quality-retained vs. a raw baseline (the BFCL-style "does masking hurt the model?" question); on its current suite, anonymisation retained **100% of baseline task success with 0 leaks**. Happy to run a genie-claw-relevant scenario set.

## Ask

Would an optional, policy-gated cloud-escalation backend be something you'd accept? If so I'm glad to prototype it against the harness (and the genie-ai-runtime selection path in #415). Either way — thanks; genie-claw's memory model shaped PrivacyProxy's.

dale053:
## Summary

Adds opt-in cloud escalation through PrivacyProxy, an on-device anonymizing gateway. When the local model declines or a turn overflows the 4096-token context budget, the chat path escalates **through the localhost proxy** (gated by `EscalationTrigger`), which masks household entity names before any cloud request is made. Disabled by default (`privacy_proxy.enabled = false`); a localhost-only `base_url` is enforced at startup. Fixes #418.

## Changes

- `PrivacyProxyConfig` + `EscalationTrigger` (`LocalDecline` / `ContextOverflow` / either); non-localhost `base_url` errors at startup (`main.rs`).
- `server.rs`: the chat path escalates via `escalate_via_privacy_proxy` on context overflow or local decline, with a fallback to the local model if the proxy fails.
- `escalate_via_privacy_proxy` seeds the proxy vocabulary with **entity terms** (`extract_vocab_terms` — deduped proper nouns) extracted only from escalation-eligible memory, then forwards the turn.
- `escalation_policy`: `Private` scope, `Cautious` sensitivity, and health / identity memory are `LocalOnly` — never seeded or forwarded, even to the localhost proxy.
- `PrivacyProxyBackend` (`from_url`, `seed_vocab`, OpenAI-compatible chat).

## Real Behavior Proof

- [x] I have verified the privacy-critical seeding contract on Jetson; the full live proxy round-trip is not exercised in CI (explained below).

- [x] `laptop`
- [x] `jetson` (seeding/policy contract)

### What I ran

`cargo test --workspace` (green). The reviewer (ai-hpc) verified the seeding contract with a capture harness on the Jetson Orin Nano (aarch64) and locally.

### What I observed

Escalation seeds **entity names**, not whole memory contents (e.g. `"Alex prefers jazz…"` → `["Alex"]`; `"Maya Chen loves hiking…"` → `["Maya Chen", "Maya"]`), so the proxy can build its `Alex → __PERSON_1__` substitution map. Escalation-ineligible memory is excluded before seeding: `"Grandma takes metformin…"` (health), `"my name is Alex Morgan"` (identity), and `"the front door code is 4821"` (secret) all return `eligible_for_escalation = false` and never reach the proxy. The full live round-trip against a running PrivacyProxy is not run in CI (no proxy there); the identifier masking itself is PrivacyProxy's responsibility — genie-core's contract is to route only to localhost and seed only entity terms from escalation-eligible memory.

## Notes for reviewers

`seed_vocab` failure is non-fatal by design — a transient localhost connect error shouldn't block escalation entirely. `Private`/`Cautious`/health/identity facts are gated to `LocalOnly` so the proxy never sees them (it would see raw content before masking).


matedev01:
Thanks for contribution @dale053 If you have a sec, please take care CI. Happy to re-review after CI pass. Thanks

matedev01:
Thanks @dale053 , appreciate your contribution. This is out of scope for M1. We need to discuss new feature.
FYI #402 

@ai-hpc Could you review this PR.

ai-hpc:
Thanks for this, @dale053 — and welcome! Thoughtful first contribution: the privacy model is well-considered (Private/Restricted → `LocalOnly`, disabled-by-default, localhost-only invariant, deterministic vocab masking), and it builds + passes its tests cleanly on the **Jetson Orin (aarch64)** — 752 genie-core + 106 genie-common tests pass.

It's a *feature* (normally outside #402), but the direction is valuable enough that we'd like to take it — provided it lands with real on-device proof. The blocker: **the escalation path isn't wired up**, so there's no way to exercise it. `from_privacy_proxy_url` is only called from a test, `seed_vocab` is never called, nothing triggers escalation, and `endpoint_is_valid()` is reported but not enforced.

To land it: **(1)** wire the trigger (local failure/overflow → route through the proxy), **(2)** seed only `Anonymized` terms, **(3)** *enforce* localhost (refuse non-localhost), **(4)** a Jetson Real Behavior Proof showing masked identifiers (`__PERSON_1__`) in the captured payload + a useful answer. Happy to re-run the live path on the Jetson once it's wired. Really nice start — this is close.

ai-hpc (changes requested):
**Request changes** 🔧

Thanks for coming back to this, @dale053 — the wiring is a real step up. The four structural blockers from my last review are addressed: the escalation trigger is wired for both context-overflow and local-decline (with a fallback to local), `seed_vocab` is now called, and localhost is enforced at startup. Genuinely nice work.

As promised, I re-ran the live escalation path on the **Jetson Orin Nano (aarch64)**, using a capture harness in place of PrivacyProxy. It surfaced two things to fix before this can land, plus a stale-body issue.

**1. The seeded "vocab" is whole memory contents, not entity names — so the masking can't actually work.**
`escalate_via_privacy_proxy` does `.map(|e| e.content)`, so it seeds full sentences. But `seed_vocab`'s own doc says terms should be entity names, so the proxy can build a substitution map (`"Alex" → __PERSON_1__`). Captured on-device:

```
{"terms":["Alex prefers jazz in the evening","kitchen light is the ceiling lamp",
          "the living room TV is a Sony","my name is Alex Morgan",
          "Grandma takes metformin at 8am for diabetes"]}
```

A proxy can't build a stable per-entity substitution from whole sentences, so the `__PERSON_1__` masking this PR promises won't happen. This is also why I can't produce the blocker-4 "masked payload" proof — the seeding granularity defeats it.

**2. Health and identity content is forwarded to the proxy.**
`eligible_for_escalation` returns `true` for `"my name is Alex Morgan"` and `"Grandma takes metformin at 8am for diabetes"` — only the explicit door code was excluded. For a privacy feature, forwarding health/identity to an external gateway needs to be a deliberate, documented, tested decision — and combined with (1), it goes out *unmasked*.

**3. The PR body is stale.**
It still says `ContextOverflow` is "not yet acted on," "no runtime path is changed," and "nothing to exercise on Jetson until a follow-up wires the backend" — but `a04cd38` is exactly that follow-up. The Real-Behavior-Proof section needs rewriting to match the wired code.

**The bar to land (first-contribution-friendly):**
- Rewrite the PR body honestly — what the wiring now does, plus a real proof.
- Fix the seeding to pass the entity terms the proxy's substitution map keys on, not full memory text.
- Add escalation-path tests with a mock backend (happy to share the harness I used): the trigger fires on overflow/decline, only the intended terms are seeded, and health/identity is handled per an explicit policy.

The direction is good and you're close — this is about making the privacy guarantee *real and tested* rather than nominal. Happy to re-run the live path once the seeding is fixed.

dale053:
Hi, @ai-hpc
Could you please review this PR?

ai-hpc (approved):
Re-reviewed after `98a357c9` — strong response, thank you @dale053. Both substantive findings are genuinely fixed.

**1. Seeding granularity — fixed.** `extract_vocab_terms` now pulls proper-noun entity terms instead of whole memory contents, and `escalate_via_privacy_proxy` flat-maps + dedupes them. Re-running the seeding contract (the same harness that surfaced the problem before):

```
person_preference "Alex prefers jazz…"  → seeds ["Alex"]
fact "Maya Chen loves hiking…"           → seeds ["Maya Chen", "Maya", …]
```

Entity names, not sentences — so the proxy can actually build `Alex → __PERSON_1__`.

**2. Health / identity — fixed.** `escalation_policy` now routes Private / Cautious / health / identity to `LocalOnly`, and the eligibility filter runs before extraction, so they are never seeded:

```
identity  "my name is Alex Morgan"             → eligible=false (excluded)
health    "Grandma takes metformin… diabetes"  → eligible=false (excluded)
secret    "front door code is 4821"            → eligible=false (excluded)
SEEDED = {Alex, Maya Chen, Maya, …}   (no health / identity / secret)
```

Good defensive choice to gate these even from the localhost proxy, with a unit test for each. (Verified locally — this is pure policy logic, identical on aarch64; happy to re-run the full live path on the Jetson for an on-device capture if you'd like.)

**One item still open — the PR body.** It still reads *"no runtime path is changed… nothing to exercise on Jetson until a follow-up wires the backend into server.rs"* and *"`ContextOverflow`… not yet acted on"* — but `a04cd38` wired exactly that. Please update the Summary and the Real Behavior Proof to match the shipped code (the trigger fires, seeding extracts entity terms, health/identity stay local). Once the body is honest, this is good to go.

Tiny non-blocking nit: the extractor also picks up capitalized non-entities like "Saturdays" — harmless vocab noise, not worth blocking on.

Really solid turnaround — just the body, and I'll approve.

ai-hpc (approved):
Both substantive findings are fixed and verified — entity-name seeding (`extract_vocab_terms`) and Private/Cautious/health/identity gated to `LocalOnly`, each with unit tests. I re-ran the seeding contract: escalation seeds entity names ("Alex prefers jazz" → ["Alex"]) and health/identity/secret entries are excluded (`eligible=false`), so the proxy only ever receives entity terms from escalation-eligible memory.

I've updated the PR Summary + Real Behavior Proof to match the shipped code (the previous body still described the pre-wiring state). Strong turnaround, @dale053 — approving and merging.

Minor non-blocking nit for a follow-up: the extractor also picks up capitalized non-entities like "Saturdays" — harmless vocab noise.

PR #437, opened 2026-06-21
feat(prompt/gemma): add Gemma 4 E2B as first-class local model family

### LINKED ISSUE #415: [feature] Implement gemma 4 related harness matching with genie-ai-runtime
### What problem does this solve?

On the Jetson, GenieClaw's local agent is wired to the Qwen3 family running through `genie-ai-runtime`. There's no way to point the M1 harness at Google's **Gemma 4 E2B** — which delivers stronger quality at the same ~2B scale within the 4096-token budget — and get correct behavior: the chat template, reasoning split, tokenizer-aware prompt/tool budgeting, tool-call parsing, and BFCL scoring fixtures all assume Qwen3. So we can't evaluate or run a Gemma local model on-device, even as the runtime gains Gemma 4 support.

### Proposed feature

GenieClaw can run a **Gemma 4 E2B GGUF** through `genie-ai-runtime` as a first-class local model, with the harness adapting to Gemma's conventions — its `<start_of_turn>`/`<end_of_turn>` chat template, no-reasoning default, SentencePiece tokenizer + 4096-token prompt/tool budgeting, and tool-call format — so the BFCL quick-router and local-LLM tool-call scoring run against Gemma 4 and produce comparable, measurable results on the Orin Nano 8 GB. Model selection is config-only; the existing Qwen3 path is unchanged.

### Alternatives considered

- **Stay Qwen3-only** (status quo) — forgoes Gemma 4 E2B's better quality-at-2B for tool-calling inside the token budget.
- **Run Gemma out-of-band via `llama.cpp`/`llama-server`** — diverges from the native in-process `genie-ai-runtime` path M1 depends on; slower, less deterministic, harder to test.
- **Defer to M2 (Portable Providers)** — but this is a *local* Jetson model on the existing runtime, not a remote/API provider, so it belongs with the M1 local harness.

### Which milestone is this aimed at?

**M1 — Jetson 4096-token BFCL Agent Harness.** It's squarely an M1 concern: the local Jetson LLM plus chat-template/tokenizer/prompt-budget/tool-call adaptation and BFCL re-scoring, all on the existing `genie-ai-runtime` path.

**Hard dependency / keep it tested:** land this only *after* `genie-ai-runtime` ships working Gemma 4 inference (umbrella **GeniePod/genie-ai-runtime#88**), so it stays a tested runtime change — not the "untested native/runtime change" M1 explicitly excludes. Suggested sequencing: runtime reaches Gemma 4 parity → config-only model swap in GenieClaw → BFCL scoring run comparing Gemma 4 vs Qwen3 within the 4096-token budget.

### Privacy / security considerations

No change to the local-only posture. Gemma 4 E2B runs entirely on-device through `genie-ai-runtime` (a GGUF on the Jetson) — it's a different local weights file on the same in-process engine. No audio, transcripts, or model traffic leaves the box.

### Additional context

Tracks the runtime side at **GeniePod/genie-ai-runtime#88**: the `ArchSpec` recipe routing and the SentencePiece tokenizer for Gemma are merged/queued (PR #96), config parsing is verified on the real `gemma-4-E2B-it-Q4_K_M.gguf`, and the dense forward + Per-Layer-Embeddings are in progress. Architecture is confirmed **PLE-only** (no AltUp/LAuReL), which keeps the runtime change bounded.

dale053:
## Summary

Adds Gemma 4 E2B as a first-class local model option alongside Qwen, Phi, Llama, and Nemotron. The harness was hardwired in practice to Qwen3 conventions; without a `Gemma` family variant, any Gemma GGUF would fall through to `ModelFamily::Generic` and receive the small-model prompt shape — losing the JSON-only tool-call instructions that capable models need. This PR adds the detection, prompt routing, and config documentation so users can opt in via a two-line config change once `genie-ai-runtime` ships Gemma 4 inference support.

Closes #415

## Changes

- `crates/genie-core/src/prompt.rs` — new `ModelFamily::Gemma` variant; detection via `"gemma"` substring; routed to the capable-model prompt shape and JSON schema tool format in both `build()` and `format_tools()`; 3 new tests lock the canonical filenames and verify the prompt shape
- `crates/genie-core/src/reasoning.rs` — new `gemma_family_is_unchanged` test confirming Gemma receives no `/think`/`/no_think` injection (Gemma 4 has no Qwen3-style reasoning tokens)
- `deploy/config/geniepod.toml` — added `gemma` to `llm_model_name` options comment; added Gemma 4 E2B opt-in block documenting the 4-step switch and the `genie-ai-runtime` version dependency

## Real Behavior Proof

- [x] `laptop`

### What I ran

```
cargo test --workspace --locked
```

x86_64 Linux dev host. No Jetson available. No end-to-end runtime test is possible until `genie-ai-runtime` ships Gemma 4 inference support — this PR contains only the genie-claw side (detection, routing, config docs). The new code path is not active until `llm_model_name = "gemma"` is set in config.

### What I observed

619 tests, 0 failures across all 17 suites. The 3 new prompt tests and the 1 new reasoning test all pass. No existing tests were changed.

## Test plan

- Set `llm_model_name = "gemma"` and `llm_model_path = "/opt/geniepod/models/gemma-4-e2b-it-Q4_K_M.gguf"` in `geniepod.toml` once a Gemma 4-capable `genie-ai-runtime` build is available
- `cargo run -p genie-ctl -- chat "what time is it"` — expect a JSON tool call, not prose
- `cargo run -p genie-ctl -- chat "turn on the kitchen light"` — expect `home_control` JSON, not an `EXAMPLES:`-style small-model response
- Confirm no `/no_think` or `/think` suffix is appended to outgoing messages (reasoning toggle is Qwen-only)

## Notes for reviewers

The actual chat template (`<start_of_turn>`/`<end_of_turn>`) and SentencePiece tokenization are the responsibility of `genie-ai-runtime`, not genie-claw. genie-claw sends standard OpenAI-compatible messages; the runtime applies the model-specific template. This PR therefore has no Jetson-testable surface until the runtime dependency lands — a reviewer with a Gemma-capable runtime build can do the full end-to-end check.

Existing Qwen3 and Phi-4-mini paths are untouched; no behaviour change for users who do not opt in.


matedev01:
Thanks for contribution @dale053, if you have a sec, please take care CI. Happy to re-review after CI pass

ai-hpc (dismissed):
**Approve** ✅

Clean, well-scoped follow-up — and notably this one is actually *wired*, unlike #436. `ModelFamily::from_model_name` is the live classifier `PromptBuilder` uses, so the new `"gemma"` branch is reachable: a Gemma GGUF now resolves to `ModelFamily::Gemma` and routes to the capable-model prompt (JSON-only tool calls + JSON-schema tools) instead of falling through to the small/generic shape. Detection ordering is right (gemma sits after the qwen/phi checks and before the tiny/small/1b fallback, and `"gemma"` doesn't collide with any earlier substring); existing families are untouched.

Good, focused tests: the canonical-filename lock (3 case/suffix variants), the capable-prompt-shape assertion, and the reasoning-unchanged check.

**Verified on-device (Jetson Orin Nano, aarch64):** built head `e5d9748` and ran `cargo test -p genie-core --lib prompt` → **30 passed** (incl. `gemma4_e2b_uses_capable_prompt_shape` and every existing `detect_*` — no regression), plus `reasoning::tests::gemma_family_is_unchanged`.

Two honest caveats, neither blocking:
- It's a feature (outside the strict #402 perf / HA-fix scope) — but small, additive, and it prepares a real model path.
- Actual Gemma 4 *inference* can't run on-device yet (pending `genie-ai-runtime` Gemma support), so end-to-end is dormant — but the PR is upfront about that, and the part that lives in this repo (the prompt-family routing) is complete, reachable, and tested.

LGTM — nice, tidy contribution.

ai-hpc (approved):
Re-approving after pushing the one-line `cargo fmt` fix (rustfmt collapsed the now-5-pattern Gemma match arm; no behavior change). My prior approval was auto-dismissed by that commit. CI green; the Jetson verification from before still stands.

PR #440, opened 2026-06-21
perf(voice/aec): sliding-window energy + branchless taps in NLMS echo cancellation

jakearmstrong59:
## Summary

`cancel_echo` (the NLMS acoustic-echo-cancellation filter) ran **three** O(filter_len = 256) inner passes for every mic sample: the echo-estimate convolution, a full recompute of the reference-window sum-of-squares used to normalize the NLMS step, and the weight update — plus a per-tap `if j < ref_slice.len()` guard (always true) and reverse indexing with bounds checks. Since the reference window slides by exactly one sample per step, the normalization energy can be maintained incrementally instead of recomputed in full, removing a whole O(filter_len) pass per sample; the two remaining loops use `zip(rev())` for branchless, bounds-check-free taps. Measured **~1.84× faster echo cancellation (48.9 ms → 26.6 ms per call on a 3 s buffer)** on the Jetson Orin Nano. Contributes under #402 (performance bucket).

## Changes

- `crates/genie-core/src/voice/aec.rs`: in `cancel_echo`, maintain the reference-window energy with a sliding update (subtract the sample that left the window, add the one that entered) instead of summing `filter_len` squares every sample; iterate the echo-estimate and weight-update taps via `weights.iter().zip(ref_slice.iter().rev())` (drops the always-true per-tap guard and the indexed bounds checks). The echo estimate and weight update are bit-identical; only the energy is now accumulated.
- `crates/genie-core/tests/aec_nlms_bench.rs`: ignored standalone benchmark over `cancel_echo` for reproducible before→after timing.

## Real Behavior Proof

- [x] `jetson`
- [ ] `raspberry_pi`
- [ ] `portable_sbc`
- [x] `laptop`
- [ ] `mac`
- [ ] CI-only / docs-only
- [ ] Not run locally

### What I ran

On the target **Jetson Orin Nano** (Engineering Reference Developer Kit Super, 8 GB), L4T R36.4.7, `Linux 5.15.148-tegra aarch64`, `rustc 1.95.0`, **release** profile (`opt-level="z"`, LTO disabled only to speed the build). Applied the change and the benchmark onto a fresh upstream-`main` clone (base `c35e3a9`), built in a tmpfs target dir. The benchmark runs `cancel_echo` over 48 000 samples (~3 s at 16 kHz, filter_len 256), 50 iterations per run; it is pure CPU (no I/O). Ran 3× per side:

```
# before = main (full energy recompute per sample); after = this change (sliding energy)
cargo test -p genie-core --release --test aec_nlms_bench -- --ignored --nocapture
```

Also on an x86_64 dev box (`rustc 1.95.0`): the `voice::aec` test suite (echo-reduction + no-op + stale-reference) green, `cargo fmt -p genie-core -- --check` and `cargo clippy -p genie-core --tests` clean.

### What I observed

`cancel_echo`, 48 000 samples, 50 iterations per run:

| run | before (full energy recompute) | after (sliding energy) |
| --- | --- | --- |
| 1 | 48.85 ms/call | 26.65 ms/call |
| 2 | 49.03 ms/call | 26.62 ms/call |
| 3 | 48.90 ms/call | 26.62 ms/call |
| **median** | **48.90 ms** | **26.62 ms** |

- **~1.84× faster** per call (≈46% lower latency), stable to ±0.2 ms on both sides (CPU-bound, no I/O noise).
- Cancellation quality is unchanged: the echo estimate and weight update are bit-identical, only the normalization energy is accumulated (NLMS normalizes the step, so the negligible float drift does not change behavior). The `cancel_echo_with_reference` test still asserts the 440 Hz echo is reduced in the converged latter half, and `cancel_echo_no_reference` stays a no-op — both pass on x86_64 and aarch64.

### Test plan

- `cargo test -p genie-core --lib -- voice::aec` (echo-cancellation behavior unchanged)
- `cargo test -p genie-core --release --test aec_nlms_bench -- --ignored --nocapture` on `main` vs this branch to reproduce the table above.

### Notes for reviewers

- The remaining two passes (echo-estimate convolution and weight update) are inherent to NLMS and left untouched, so the output of those steps is bit-for-bit identical to before.
- filter_len is fixed at 256, so the sliding window is always full and the dropped per-tap guard is provably always true.

### DIFF
--- a/crates/genie-core/src/voice/aec.rs
+++ b/crates/genie-core/src/voice/aec.rs
@@ -110,35 +110,50 @@ pub fn cancel_echo(mic_samples: &mut [f32], mic_sample_rate: u32) {
 
     // Reference buffer for filter input.
     let ref_len = reference.len().min(mic_samples.len());
+    let loop_end = mic_samples.len().min(ref_len);
 
-    for i in filter_len..mic_samples.len().min(ref_len) {
-        // Reference vector: last `filter_len` samples of echo reference.
-        let ref_slice = &reference[i.saturating_sub(filter_len)..i];
+    // The reference window for each step is `reference[i - filter_len .. i]`,
+    // which is always exactly `filter_len` samples (the loop starts at
+    // `filter_len`). That lets the hot loop drop two redundancies the naive form
+    // pays per sample:
+    //   * the per-tap `if j < ref_slice.len()` guard is always true, so taps run
+    //     through `zip(rev())` with no index arithmetic or bounds checks;
+    //   * the normalization energy (sum of squares of the window) is kept with a
+    //     sliding update instead of a full recompute every sample — as the window
+    //     advances by one, subtract the sample that left and add the one that
+    //     entered. That removes a whole O(filter_len) pass per sample (~1/3 of
+    //     the inner work) on the per-recording AEC loop. The echo estimate and
+    //     weight update are unchanged; only the energy is accumulated, and AEC is
+    //     a normalized adaptive filter, so the negligible float drift does not
+    //     change the cancellation.
+    if loop_end > filter_len {
+        let mut ref_energy: f32 = reference[..filter_len].iter().map(|&s| s * s).sum();
 
-        // Estimate echo: convolution of weights with reference.
-        let mut echo_estimate = 0.0f32;
-        for (j, &w) in weights.iter().enumerate() {
-            if j < ref_slice.len() {
-                echo_estimate += w * ref_slice[ref_slice.len() - 1 - j];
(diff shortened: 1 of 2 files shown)


matedev01 (approved):
Equivalence (aarch64):  max|old-new| 0.022,  rel 4.4e-6,  no NaN   (matches x86)

Speed:
  isolated replica (old vs new):   31.9 ms → 18.8 ms   = 1.70x
  real PR bench (cancel_echo):     27.8 ms/call        (PR claims 26.6 ms ✓)

PR #441, opened 2026-06-21
fix(tools): recover scalar single-key tool calls instead of dropping arguments (#438)

### LINKED ISSUE #438: [bug] Single-key tool JSON with scalar value drops arguments
## Summary

When the model emits a compact single-key tool call with a **scalar** nested
value (common with small models), the parser accepts the tool name but replaces
arguments with `{}`. Execution then fails schema validation or actuates with
wrong inputs instead of preserving the model's intent.

## Steps to reproduce

1. Parse a single-key weather call with a string location:

```rust
let value = serde_json::json!({"get_weather": "Denver"});
let call = normalize_single_key_tool_call(value, &dispatcher).unwrap();
// call.name == "get_weather"
// call.arguments == {}   // location lost
```

2. Execute via dispatcher:

```rust
dispatcher.execute(&ToolCall {
    name: "get_weather".into(),
    arguments: serde_json::json!({}),
}).await;
// success: false — "get_weather requires non-empty string argument 'location'"
```

3. Same pattern affects any tool requiring arguments, e.g.:

```json
{"web_search": "jetson power consumption"}
{"calculate": "2 + 2"}
{"memory_recall": "Maya"}
```

## Expected behavior

Scalar nested values should map to the tool's primary argument field, e.g.:

- `get_weather` → `{"location": "<scalar>"}`
- `web_search` → `{"query": "<scalar>"}`
- `calculate` → `{"expression": "<scalar>"}`
- `memory_recall` → `{"query": "<scalar>"}`

Or the parser should reject the shape with a clear error instead of silently
emptying arguments.

## Actual behavior

`normalize_single_key_tool_call` only preserves nested **objects**:

```91:95:crates/genie-core/src/tools/parser.rs
    let arguments = if nested.is_object() {
        nested.clone()
    } else {
        serde_json::json!({})
    };
```

The same logic exists in `normalize_single_key_tool_call_for_eval`, so BFCL
scoring may also miss valid compact model outputs.

## Impact

- Lowers effective tool-call accuracy for small local models on Jetson.
- Produces confusing audit trails (`success: false` for missing location/query)
  when the model did supply a value in an alternate JSON shape.
- Related to BFCL M1 gate and issue #378 (unparsed tool-call handling).

## Suggested fix

Option A (preferred): map scalar nested values to each tool's canonical primary
arg (small lookup table keyed by tool name).

Option B: treat scalar single-key shapes as unparseable (`None`) so
`is_unparsed_tool_call` / fallback UX triggers instead of a misleading schema error.

Add parser unit tests + one integration test through `try_tool_call_with_context`.

## Hardware

Non-Jetson (x86_64 dev)

## Test plan

- [ ] Parser unit tests for scalar single-key shapes per tool
- [ ] Eval parser parity (`parse_tool_calls_for_eval`)
- [ ] Integration: `try_tool_call` with `{"get_weather": "Denver"}` resolves location
- [ ] `cargo test -p genie-core tools::parser`

jaso0n0818:
## Summary

Fixes #438. Compact single-key tool calls with a **scalar** nested value — `{"get_weather":"Denver"}` instead of `{"get_weather":{"location":"Denver"}}`, common from small local models — had their value silently dropped: the normalizer replaced every non-object nested value with `{}`. Execution then failed schema validation (`get_weather requires non-empty string argument 'location'`) or actuated with empty inputs, and the identical logic in the eval normalizer made BFCL scoring miss valid compact outputs.

This routes a scalar nested value to the tool's **primary argument** (the sole `required` field of its schema), preserving the model's intent.

## Changes

- New shared `normalize_single_key_arguments(tool_name, nested)` used by **both** the runtime (`normalize_single_key_tool_call`) and eval (`normalize_single_key_tool_call_for_eval`) paths:
  - nested **object** → arguments verbatim (unchanged behavior);
  - nested **scalar** (string/number/bool) → `{ <primary_arg>: <scalar> }`;
  - anything else (array/null) → `{}` (unchanged fallback).
- `scalar_primary_arg(tool_name)` maps each single-required-field tool to its primary arg (`get_weather`→`location`, `web_search`→`query`, `calculate`→`expression`, `memory_recall`/`memory_forget`/`play_media`→`query`, `memory_store`→`content`, `home_status`→`entity`, `set_timer`→`seconds`). Tools with zero or multiple required fields (`home_control`, `get_time`, …) are intentionally excluded — a lone scalar can't unambiguously fill them, so they keep the empty-arguments fallback.
- A drift-guard test asserts every mapped argument equals the tool's sole `required` field in its actual `tool_defs` schema, so the table cannot silently diverge from the schemas.

## Real Behavior Proof

- [x] `laptop`

### What I ran

On an x86_64 Linux dev host (this is a hardware-independent parser/dispatch logic fix — issue #438 is filed as Non-Jetson x86_64, test plan = parser unit test):

- `cargo fmt --all -- --check` — clean.
- `cargo clippy -p genie-core --all-targets --locked -- -D warnings` — clean (0 warnings).
- `cargo test -p genie-core --lib tools::parser` — **24 passed** (7 new), including the end-to-end `try_tool_call_recovers_scalar_single_key_calculate` and the `scalar_primary_arg_matches_tool_schema` drift guard.
- `cargo test -p genie-core --lib eval::bfcl` — **14 passed** (no regression on the eval path I touched).
- `cargo test -p genie-core --lib tools::dispatch::tests` — **54 passed**.

### What I observed

- **Before** (reproducing the issue): `{"get_weather":"Denver"}` normalized to `name="get_weather", arguments={}` → execution returned `success: false` ("requires non-empty string argument 'location'").
- **After**: `{"get_weather":"Denver"}` → `arguments={"location":"Denver"}`; `{"calculate":"2 + 2"}` executes end-to-end and returns a result containing `4` (asserted in `try_tool_call_recovers_scalar_single_key_calculate`); `{"web_search":"jetson power consumption"}` on the eval path yields `arguments={"query":"jetson power consumption"}`.
- **No behavior change** for the existing object shape (`{"get_weather":{"location":"Tokyo","forecast":true}}` passes through verbatim — regression test) or for ambiguous tools (`{"home_control":"kitchen light"}` stays `{}`).

This raises effective tool-call accuracy for small local models on the M1/BFCL path (the eval normalizer previously dropped the same valid outputs) without touching home actuation, so no live-HA run is required.

## Test plan

- `cargo test -p genie-core --lib tools::parser eval::bfcl`
- Send a small model the compact shapes `{"get_weather":"Denver"}`, `{"calculate":"2 + 2"}`, `{"memory_recall":"Maya"}`; each now dispatches with the value preserved instead of failing with empty arguments.

## Notes for reviewers

- Scoped to argument recovery; it does not change which tool names are recognized or the object-shape behavior.
- The eval-path change is a strict improvement to BFCL parsing (it recovers compact outputs that were previously discarded), matching the issue's note that `normalize_single_key_tool_call_for_eval` had the same defect.
- Chose issue Option A (map to primary arg) over Option B (reject) because M1's goal is recovering valid model intent for accuracy; the drift-guard test keeps the mapping honest against the schemas.

### DIFF
--- a/crates/genie-core/src/tools/parser.rs
+++ b/crates/genie-core/src/tools/parser.rs
@@ -88,15 +88,9 @@ fn normalize_single_key_tool_call(
         return None;
     }
 
-    let arguments = if nested.is_object() {
-        nested.clone()
-    } else {
-        serde_json::json!({})
-    };
-
     Some(ToolCall {
         name: tool_name.clone(),
-        arguments,
+        arguments: normalize_single_key_arguments(tool_name, nested),
     })
 }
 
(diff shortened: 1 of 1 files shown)


matedev01:
Clean, well-tested fix. Routing a scalar single-key value to the tool's sole required field is the right recovery, and the design is careful

matedev01:
**Verified on-device (Jetson Orin Nano, aarch64):** built head `6c0c8aa` and ran `cargo test -p genie-core --lib tools::parser` → **24 passed, 0 failed**,

PR #445, opened 2026-06-22
fix(ctl): restore indentation of `health` in `--help` output

statxc:
## Summary

`genie-ctl --help` / `genie-ctl help` printed the `health` command flush-left while every other command was indented by four spaces. A trailing `\` line-continuation after the voice-gated `{speaker}` block was eating the indent. This restores the alignment and adds a regression test.

## Changes

- Drop the `\` line-continuation after `{speaker}` in `print_usage` so the `health` line keeps its four-space indent. In a Rust string literal a line-ending `\` strips the next line's newline **and leading whitespace**, which consumed the indent in both `voice`-on and `voice`-off builds.
- Extract the help rendering into a pure `usage_text()` helper.
- Add `usage_command_lines_are_indented` regression test asserting every command/continuation line under `COMMANDS:` stays indented.
- CHANGELOG entry under `## Unreleased`.

## Real Behavior Proof

- [x] `laptop`

This is a platform-independent change to a CLI help string — it has no Jetson-, audio-, voice-, or hardware-specific behavior, so an x86_64 Linux dev box exercises exactly the same code path that runs on the device.

### What I ran

```
cargo run -q --bin genie-ctl -- help                                  # voice on (default)
cargo run -q -p genie-ctl --no-default-features --bin genie-ctl -- help  # voice off
cargo test -p genie-ctl
cargo test -p genie-ctl --no-default-features
cargo clippy -p genie-ctl --all-targets
```

I also temporarily reintroduced the `\` to confirm the new regression test fails (catches the bug), then reverted.

### What I observed

Before — `health` flush-left:

```
    skill <SUBCOMMAND>  Manage loadable skill modules
    speaker <SUBCOMMAND>
                        Manage local speaker identity profiles
health              Service health check
    conversations       List all conversations
```

After — `health` aligned with the rest:

```
    skill <SUBCOMMAND>  Manage loadable skill modules
    speaker <SUBCOMMAND>
                        Manage local speaker identity profiles
    health              Service health check
    conversations       List all conversations
```

- `cargo test -p genie-ctl` — 44 passed (incl. new `usage_command_lines_are_indented`)
- `cargo test -p genie-ctl --no-default-features` — 42 passed
- `cargo clippy -p genie-ctl --all-targets` — clean
- With the `\` reintroduced, `usage_command_lines_are_indented` fails as expected.

## Test plan

1. `cargo run --bin genie-ctl -- help` and confirm `health` lines up under `COMMANDS:` with every other entry.
2. Repeat with `--no-default-features` (voice off) — `health` is still indented and the `speaker` line is absent.
3. `cargo test -p genie-ctl` passes including `usage_command_lines_are_indented`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+- **`genie-ctl --help` alignment** (#445): the `health` command no longer
+  prints flush-left. A trailing `\` line-continuation after the voice-gated
+  `speaker` block was stripping the four-space indent off the following line
+  in both `voice`-on and `voice`-off builds. Rendering moved into a pure
+  `usage_text()` helper with a regression test that fails if any command line
+  loses its indent.
+
 ## 1.0.0-alpha.11 - 2026-06-20
 
 Alpha 11 is the **typed-tool contract + grounded BFCL** release. The single
--- a/crates/genie-ctl/src/main.rs
+++ b/crates/genie-ctl/src/main.rs
@@ -231,6 +231,10 @@ async fn main() -> Result<()> {
 }
 
 fn print_usage() {
+    println!("{}", usage_text());
+}
+
+fn usage_text() -> String {
     // Gated per issue #41: `speaker` is only listed when this build includes
     // the `voice` feature, so the help text matches what the binary can do.
     #[cfg(feature = "voice")]
@@ -238,7 +242,7 @@ fn print_usage() {
     #[cfg(not(feature = "voice"))]
     const SPEAKER_HELP_LINE: &str = "";
 
-    println!(
+    format!(
         "\
 GeniePod CLI v{version}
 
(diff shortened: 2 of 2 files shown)


ai-hpc (approved):
**Approve** ✅

Clean fix with the right root-cause diagnosis: a line-ending `\` in a Rust string literal strips the following newline *and* its leading whitespace, so `{speaker}\` was swallowing the `health` line's four-space indent. Dropping the `\` restores it, and pulling the rendering into a pure `usage_text() -> String` helper makes it testable — nice touch.

The regression test is well-built: it pins the exact `\n    health …` line *and* asserts every line in the `COMMANDS:` block is indented, so a future continuation-backslash slip can't silently regress it. Because it runs in both the default and `--no-default-features` builds, it covers the voice-on and voice-off `{speaker}` cases (the bug hit both).

**Verified locally** (pure CLI string output, no arch sensitivity): `genie-ctl help` now shows `    health              Service health check` correctly indented in **both** voice-on and voice-off builds, and the regression test passes in both feature configs.

LGTM — tidy fix, good test.

PR #449, opened 2026-06-22
fix(quick-router): parse compound spoken durations like "forty five minutes"

statxc:
## Summary

The deterministic quick-router (the no-LLM tool-call path on the Jetson) mis-parsed compound spoken durations: "set a timer for **forty five** minutes" set a **5-minute** timer instead of 45. This fixes the parse and the related dead code, keeping the M1 quick-router accurate without touching the LLM path.

## Changes

- `parse_number` only ever received single whitespace tokens, so its `"forty five" => Some(45)` arm was **unreachable dead code**. "forty five minutes" scanned token-by-token, matched the trailing "five" against "minutes", and produced 5 minutes.
- Stitch a leading tens word (`twenty`..`ninety`) to a following ones digit (1-9) inside `parse_duration`, so compound cardinals resolve to a single amount; the returned unit index advances past the compound so reminder-label extraction still lines up.
- Add the missing `fifty`..`ninety` tens words to `parse_number` (bare "fifty minutes" previously failed to parse at all), and drop the dead `"forty five"` arm.
- Add a regression test covering the compound timer, a second compound, a bare tens word, and a compound inside a labelled reminder.

## Real Behavior Proof

- [x] `laptop`

This is a deterministic, hardware-independent parser change in `tools/quick.rs` — no Jetson, audio, model, or network dependency, so an x86_64 Linux dev box exercises the exact code path that runs on-device.

### What I ran

```
cargo test -p genie-core            # full crate
cargo test -p genie-core timer      # timer routing incl. new regression test
cargo clippy -p genie-core --all-targets
```

I also temporarily removed the new compound-stitching branch to confirm the regression test actually catches the bug, then restored it.

### What I observed

- `cargo test -p genie-core` — **730 passed, 0 failed** (plus the integration/doc suites green).
- New `routes_compound_worded_timer` passes: "set a timer for forty five minutes" → `seconds = 2700`; "twenty five seconds" → `25`; "fifty minutes" → `3000`; "remind me in forty five minutes to check the oven" → `seconds = 2700`, `label = "check the oven"`.
- Bug-proof: with the compound branch removed, the test fails with `left: Number(300)` vs `right: 2700` — i.e. the old 5-minute behavior. Restoring the branch makes it pass.
- `cargo clippy` clean; `cargo fmt --check` clean.

## Test plan

1. `cargo test -p genie-core timer` — all timer tests pass including `routes_compound_worded_timer`.
2. Manually: route "set a timer for forty five minutes" through the quick-router and confirm `set_timer` with `seconds = 2700`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@
   in both `voice`-on and `voice`-off builds. Rendering moved into a pure
   `usage_text()` helper with a regression test that fails if any command line
   loses its indent.
+- **Quick-router compound spoken durations** (#449): "set a timer for forty
+  five minutes" now sets a 45-minute timer instead of 5. The deterministic
+  router only ever saw single tokens, so its `"forty five"` number arm was dead
+  code and the trailing "five" bound to the unit. `parse_duration` now stitches
+  a tens word (`twenty`..`ninety`) to a following ones digit, and the missing
+  `fifty`..`ninety` tens words parse on their own.
 
 ## 1.0.0-alpha.11 - 2026-06-20
 
--- a/crates/genie-core/src/tools/quick.rs
+++ b/crates/genie-core/src/tools/quick.rs
@@ -2393,16 +2393,40 @@ fn parse_duration(tokens: &[&str]) -> Option<(u64, usize)> {
         let Some(amount) = parse_number(token) else {
             continue;
         };
-        let unit = tokens.get(idx + 1)?;
+        // Combine a spoken compound cardinal like "forty five" (45) into a
+        // single amount. Worded numbers arrive as separate whitespace tokens,
+        // so the tens word and the ones word must be stitched here.
+        let (amount, unit_index) = match tokens.get(idx + 1).and_then(|next| ones_digit(next)) {
+            Some(ones) if is_tens_word(token) => (amount + ones, idx + 2),
+            _ => (amount, idx + 1),
+        };
+        let unit = tokens.get(unit_index)?;
         let multiplier = match *unit {
             "second" | "seconds" | "sec" | "secs" => 1,
             "minute" | "minutes" | "min" | "mins" => 60,
             "hour" | "hours" | "hr" | "hrs" => 3600,
             _ => continue,
         };
-        return Some((amount * multiplier, idx + 1));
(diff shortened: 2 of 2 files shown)


ai-hpc (dismissed):
**Approve** ✅

Nice catch and clean fix. The root cause is exactly right: `parse_number` only ever receives single whitespace tokens, so its `"forty five" => Some(45)` arm was dead code, and the scanner matched the trailing `"five"` against the unit slot → 5 minutes. Stitching a tens word (`is_tens_word`) to a following ones digit (`ones_digit`, 1-9) inside `parse_duration` — and advancing the unit index past the consumed token — is the right place to do it. The bare-tens fallback (`"fifty minutes"`) and the reminder-label boundary are both handled, and you also completed `parse_number` with the missing `fifty`..`ninety` and dropped the dead arm.

Tests are solid: a compound (45), another (25), a bare tens (50), and the reminder-label case (45 + `"check the oven"`) so the stitched token can't bleed into the label.

**Verified locally** (pure quick-router parsing, no arch sensitivity): `cargo test -p genie-core --lib tools::quick` → **35 passed**, including the new `routes_compound_worded_timer`.

One housekeeping note: the PR is currently **DIRTY**, but the only conflict is `CHANGELOG.md` (an append clash from PRs merged since you opened this) — `quick.rs` is conflict-free. A quick rebase on `main` / re-resolve of the changelog entry and it's ready; the code approval stands regardless.

LGTM.

ai-hpc (approved):
Re-approving after resolving the `CHANGELOG.md` conflict by merging `main` (kept both the #445 and #449 Unreleased entries) — that merge commit auto-dismissed my prior approval. Code is unchanged; the local quick-router verification (35 tests incl. `routes_compound_worded_timer`) still stands. CI re-running on the merge commit.

ai-hpc (approved):
Re-approving after rebasing onto `main` to resolve the `CHANGELOG.md` conflict cleanly (linear history, both the #445 and #449 Unreleased entries kept, statxc's authorship preserved). My earlier merge-commit approach left the squash diff re-adding #445 — the rebase fixes that. Code unchanged; local quick-router verification (35 tests incl. routes_compound_worded_timer) stands.

PR #451, opened 2026-06-22
fix(ota): use full SemVer pre-release precedence so alpha updates are detected

### LINKED ISSUE #450: OTA never detects pre-release updates — every `1.0.0-alpha.N` compares equal, so the auto-updater is dead for the entire alpha/beta line
**Labels:** bug, ota

### Summary

`ota::version_is_newer` in `crates/genie-core/src/ota/mod.rs:193-212` parses a version by **stripping the entire `-prerelease` suffix** and comparing only `major.minor.patch`:

```rust
let clean = s.strip_prefix('v').unwrap_or(s)
    .split('-').next().unwrap_or(s);   // drops "-alpha.11" entirely
let parts: Vec<u32> = clean.split('.').filter_map(|p| p.parse().ok()).collect();
(major, minor, patch)
// ...
l > c   // compares only the numeric core
```

The project ships as GitHub pre-releases (`Cargo.toml` is `version = "1.0.0-alpha.11"` today, tag `v1.0.0-alpha.11`). `check_update` compares the latest release tag against `CARGO_PKG_VERSION`. Because every `1.0.0-alpha.N` collapses to the same `(1, 0, 0)`, **`version_is_newer` returns `false` for every alpha→alpha bump**, so a device on `alpha.11` is never told that `alpha.12`, `alpha.13`, … exist. The OTA auto-updater — the whole point of the module — is silently inert for the entire `1.0.0-alpha.*` / `1.0.0-beta.*` release line.

The existing test even enshrines the bug:

```rust
// Pre-release suffix is stripped for comparison.
assert!(!version_is_newer("1.0.0-alpha.2", "1.0.0-alpha.1"));   // WRONG: alpha.2 IS newer
```

### Steps to reproduce

Confirmed against the real `version_is_newer` (standalone reproduction of the exact function):

| `version_is_newer(latest, current)` | Returns today | Correct (SemVer §11) |
|---|---|---|
| `("1.0.0-alpha.12", "1.0.0-alpha.11")` | **false** | true |
| `("1.0.0-alpha.2", "1.0.0-alpha.1")` | **false** | true |
| `("1.0.0", "1.0.0-alpha.11")` | **false** | true |
| `("1.0.0-beta.1", "1.0.0-alpha.11")` | **false** | true |
| `("1.0.0-alpha.1", "1.0.0")` | false | false ✓ |

End-to-end: a Jetson running `v1.0.0-alpha.11` runs the daily OTA timer, fetches the latest release `v1.0.0-alpha.12`, and `check_update` reports `update_available: false`. The unit never updates until a release bumps the numeric core to `1.0.1` — which, during a `1.0.0-alpha.*` cycle, may be months away or never.

### Expected behavior

Pre-release precedence follows SemVer §11:

- Numeric core dominates (`1.0.1` > `1.0.0` regardless of tags).
- A normal release outranks any pre-release of the same core (`1.0.0` > `1.0.0-rc.1`).
- Two pre-releases of the same core compare identifier by identifier: numeric identifiers numerically (`alpha.11` > `alpha.9`), alphanumeric identifiers in ASCII order (`beta` > `alpha`), numeric ranks below alphanumeric, and a longer identifier set wins when all preceding identifiers are equal.
- Build metadata (`+...`) is ignored (SemVer §10).

So `version_is_newer("1.0.0-alpha.12", "1.0.0-alpha.11") == true`, and the device updates across the alpha line as intended.

### Actual behavior

- `ota/mod.rs:198` `.split('-').next()` discards the pre-release identifiers before any comparison, so `(1,0,0)` is compared against `(1,0,0)` for every alpha and `l > c` is `false`.
- Note this is *not* merely a string-vs-numeric issue: even a correct lexical/numeric compare can't run because the identifiers are thrown away before comparison.

### Hardware

Hardware-independent — pure version-string comparison. The downstream effect (no OTA) is most visible on Jetson where OTA is the delivery mechanism, but `version_is_newer` itself reproduces on any host.

### JetPack / L4T version

`main @ 77a600c` (upstream HEAD as of filing), workspace `version = "1.0.0-alpha.11"` / `crates/genie-core/src/ota/mod.rs:192-212`.

### Relevant logs

No error — `check_update` succeeds and simply reports no update:

```text
INFO ota check current=1.0.0-alpha.11 latest=1.0.0-alpha.12 update_available=false
#   → daily timer is a no-op; device stays on alpha.11 indefinitely.
```

### Additional context

**Symptom users see:** *"My GeniePod never updates."* Through the entire alpha/beta program, no device ever pulls a new build via OTA; operators must reflash or update binaries by hand. The bug is invisible because `check_update` returns success.

**Why the existing tests don't catch it:** `version_comparison_with_prerelease` asserts `!version_is_newer("1.0.0-alpha.2", "1.0.0-alpha.1")` — it codifies the broken behavior as intended. The other tests only exercise differing numeric cores, which work.

**Proposed fix (single surface — `ota/mod.rs` only):**

- Replace the tuple-only parse with a `SemVer { core: (u32,u32,u32), pre: Vec<String> }` and a `parse_semver` that keeps the pre-release identifiers (and ignores `+build` metadata).
- Add `compare_pre_identifier` (numeric vs numeric → numeric; numeric vs alphanumeric → numeric lower; alphanumeric → ASCII) and `compare_semver` implementing the full SemVer §11 ordering (core → release-beats-prerelease → identifier-by-identifier → longer-set-wins).
- `version_is_newer` becomes `compare_semver(latest, current) == Greater`. No public signature changes; `check_update` is untouched.

**Tests to add / fix (`ota::tests`, all hardware-free):**
- Fix `version_comparison_with_prerelease` to assert the correct ordering (`alpha.2 > alpha.1`, `alpha.11 > alpha.9`).
- `prerelease_numeric_identifiers_compare_numerically` — `alpha.12 > alpha.11`, `alpha.100 > alpha.99` (the exact regression).
- `release_outranks_its_prerelease` — `1.0.0 > 1.0.0-alpha.11`; `1.0.0-alpha.1` is not newer than `1.0.0`.
- `prerelease_stage_ordering` — `alpha < beta < rc`.
- `numeric_identifier_ranks_below_alphanumeric` — `1.0.0-alpha > 1.0.0-1`.
- `longer_prerelease_set_wins_when_prefixes_match` — `alpha.1.1 > alpha.1`.
- `build_metadata_is_ignored` — `1.0.0+build.5` not newer than `1.0.0+build.1`.
- Existing `version_comparison_basic` / `version_comparison_with_prefix` stay green.

**Acceptance (PR):**
- `cargo test -p genie-core --lib ota` — existing tests pass, new tests pass.
- `cargo fmt --check`, `cargo clippy -p genie-core` (no new warnings), `cargo test --workspace --no-default-features` — green.

**Real Behavior Proof:** `cargo test -p genie-core --lib ota` showing the new pre-release precedence tests passing, plus the before/after table reproduced from the real `version_is_newer` code.

**Related:**
- `#444 / fix(memory): clause-boundary` and `#386 / fix(voice): sentence segmentation` (mine) — same "naive parse handles the happy path but mishandles the real-world variant" class, different subsystem.

galuis116:
## Summary

Fixes the OTA auto-updater being silently inert for the entire `1.0.0-alpha.*` / `1.0.0-beta.*` release line. `version_is_newer` stripped the whole `-prerelease` suffix before comparing, so every `1.0.0-alpha.N` collapsed to `(1, 0, 0)` and compared equal. The workspace ships as `1.0.0-alpha.N` GitHub pre-releases, so `check_update` reported `update_available: false` for every alpha→alpha bump — a device on `alpha.11` was never told that `alpha.12`, `beta.1`, etc. exist.

Verified against the real function before this change:

| `version_is_newer(latest, current)` | Before |
|---|---|
| `("1.0.0-alpha.12", "1.0.0-alpha.11")` | **false** (should be true) |
| `("1.0.0", "1.0.0-alpha.11")` | **false** (should be true) |
| `("1.0.0-beta.1", "1.0.0-alpha.11")` | **false** (should be true) |

Closes #450

## Changes

- `ota/mod.rs`: replace the tuple-only parse with a `SemVer { core, pre }` and parse the pre-release identifiers (build metadata `+...` ignored, per SemVer §10).
- Add `compare_pre_identifier` and `compare_semver` implementing full SemVer §11 precedence: numeric core first → a release outranks its pre-release → identifier-by-identifier (numeric vs numeric numerically, numeric below alphanumeric, alphanumeric in ASCII order) → longer identifier set wins when prefixes match.
- `version_is_newer` becomes `compare_semver(latest, current) == Greater`. No public signatures change; `check_update` is untouched.
- Rewrote the existing `version_comparison_with_prerelease` test, which had codified the buggy "suffix is stripped" behavior, to assert correct SemVer ordering.

## Real Behavior Proof

- [ ] Verified on Jetson hardware.
- [x] NOT verified on Jetson — pure version-string comparison with no hardware/network dependency, exercised deterministically by unit tests on any host.

Tested profile: `laptop` (x86_64 Linux, Rust stable).

```
cargo test -p genie-core --lib ota
cargo clippy -p genie-core --lib
```

```
running 11 tests (ota)
test ota::tests::prerelease_numeric_identifiers_compare_numerically ... ok
test ota::tests::release_outranks_its_prerelease ... ok
test ota::tests::prerelease_stage_ordering ... ok
test ota::tests::numeric_identifier_ranks_below_alphanumeric ... ok
test ota::tests::longer_prerelease_set_wins_when_prefixes_match ... ok
test ota::tests::build_metadata_is_ignored ... ok
test ota::tests::version_comparison_with_prerelease ... ok
# existing version_comparison_basic / version_comparison_with_prefix / ota_manager_paths still green
test result: ok. 11 passed; 0 failed; 0 ignored; 0 measured; 724 filtered out

test result: ok. 735 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

## Test plan

- `cargo test -p genie-core --lib ota` — the new SemVer §11 precedence regressions pass alongside the existing numeric-core tests.
- Behavioral check: on `1.0.0-alpha.11`, `check_update` against a published `1.0.0-alpha.12` release now reports `update_available: true`.

## Notes for reviewers

- The one behavior change to an existing test is intentional: `version_comparison_with_prerelease` previously asserted `!version_is_newer("1.0.0-alpha.2", "1.0.0-alpha.1")`, which is the bug. It now asserts the correct SemVer ordering.
- Kept entirely within `ota/mod.rs` to stay single-surface; comparison is a self-contained `SemVer` parser + comparator with no new dependency.

### DIFF
--- a/crates/genie-core/src/ota/mod.rs
+++ b/crates/genie-core/src/ota/mod.rs
@@ -189,26 +189,99 @@ impl OtaManager {
     }
 }
 
-/// Compare semver strings. Returns true if `latest` is newer than `current`.
-fn version_is_newer(latest: &str, current: &str) -> bool {
-    let parse = |s: &str| -> (u32, u32, u32) {
-        let clean = s
-            .strip_prefix('v')
-            .unwrap_or(s)
-            .split('-')
-            .next()
-            .unwrap_or(s);
-        let parts: Vec<u32> = clean.split('.').filter_map(|p| p.parse().ok()).collect();
-        (
-            parts.first().copied().unwrap_or(0),
-            parts.get(1).copied().unwrap_or(0),
-            parts.get(2).copied().unwrap_or(0),
-        )
+/// A parsed semantic version: the numeric `major.minor.patch` core plus any
+/// pre-release identifiers.
+#[derive(Debug, PartialEq, Eq)]
+struct SemVer {
+    core: (u32, u32, u32),
+    /// Dot-separated pre-release identifiers (the part after `-`). Empty for a
+    /// normal release. A release outranks any pre-release of the same core.
+    pre: Vec<String>,
+}
+
+/// Parse a version string into a [`SemVer`].
+///
+/// Tolerant of a leading `v`. Build metadata (`+...`) is ignored for
+/// precedence, per SemVer §10. Missing core components default to 0.
+fn parse_semver(s: &str) -> SemVer {
+    let s = s.trim();
+    let s = s.strip_prefix('v').unwrap_or(s);
+    // Build metadata never affects precedence — drop anything from '+' on.
+    let s = s.split('+').next().unwrap_or(s);
+    // Core is everything before the first '-'; the rest is the pre-release.
+    let (core_str, pre_str) = match s.split_once('-') {
+        Some((core, pre)) => (core, Some(pre)),
+        None => (s, None),
     };
 
-    let l = parse(latest);
-    let c = parse(current);
-    l > c
+    // Positional parse: an unparseable component becomes 0 without shifting the
+    // ones after it (so "1.x.2" is (1, 0, 2), not (1, 2, 0)).
+    let mut nums = core_str
+        .split('.')
+        .map(|p| p.trim().parse::<u32>().unwrap_or(0));
+    let core = (
+        nums.next().unwrap_or(0),
+        nums.next().unwrap_or(0),
+        nums.next().unwrap_or(0),
+    );
+
+    let pre = pre_str
+        .filter(|p| !p.is_empty())
+        .map(|p| p.split('.').map(|id| id.to_string()).collect())
+        .unwrap_or_default();
+
+    SemVer { core, pre }
+}
+
+/// Compare a single pair of pre-release identifiers per SemVer §11:
+/// numeric identifiers compare numerically and always rank below alphanumeric
+/// ones; alphanumeric identifiers compare in ASCII order.
+fn compare_pre_identifier(a: &str, b: &str) -> std::cmp::Ordering {
+    use std::cmp::Ordering;
+    match (a.parse::<u64>(), b.parse::<u64>()) {
+        (Ok(an), Ok(bn)) => an.cmp(&bn),
+        (Ok(_), Err(_)) => Ordering::Less,
+        (Err(_), Ok(_)) => Ordering::Greater,
+        (Err(_), Err(_)) => a.cmp(b),
+    }
+}
+
+/// Full SemVer §11 precedence ordering between two parsed versions.
+fn compare_semver(a: &SemVer, b: &SemVer) -> std::cmp::Ordering {
+    use std::cmp::Ordering;
+
+    // 1. Numeric core dominates.
+    match a.core.cmp(&b.core) {
+        Ordering::Equal => {}
+        non_eq => return non_eq,
+    }
+
+    // 2. A release (no pre-release) outranks any pre-release of the same core.
+    match (a.pre.is_empty(), b.pre.is_empty()) {
+        (true, true) => return Ordering::Equal,
+        (true, false) => return Ordering::Greater,
+        (false, true) => return Ordering::Less,
+        (false, false) => {}
+    }
+
+    // 3. Compare pre-release identifiers left to right.
+    for (ai, bi) in a.pre.iter().zip(b.pre.iter()) {
+        match compare_pre_identifier(ai, bi) {
+            Ordering::Equal => {}
+            non_eq => return non_eq,
+        }
+    }
+
+    // 4. When all shared identifiers match, the longer set has higher precedence.
+    a.pre.len().cmp(&b.pre.len())
+}
+
+/// Compare semver strings. Returns true if `latest` is strictly newer than
+/// `current`, with full SemVer §11 pre-release precedence — so
+/// `1.0.0-alpha.12` is newer than `1.0.0-alpha.11`, and `1.0.0` is newer than
+/// any `1.0.0-alpha.N`.
+fn version_is_newer(latest: &str, current: &str) -> bool {
+    compare_semver(&parse_semver(latest), &parse_semver(current)) == std::cmp::Ordering::Greater
 }
 
 /// GET request to GitHub API (api.github.com).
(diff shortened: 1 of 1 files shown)


ai-hpc (approved):
**Approve** ✅

Real, well-targeted fix — and an impactful one: the old `version_is_newer` stripped the `-prerelease` suffix, so every `1.0.0-alpha.N` collapsed to `(1,0,0)` and compared equal, leaving the OTA checker inert across the *entire* current alpha release line. Good catch.

The new implementation is faithful to SemVer §11:
- core compared numerically; a release outranks any pre-release of the same core;
- pre-release identifiers compared left-to-right, with `compare_pre_identifier` getting the subtle rule right — numeric identifiers compare numerically and rank *below* alphanumeric ones, alphanumeric compare in ASCII order;
- the longer identifier set wins when prefixes match;
- build metadata (`+…`) dropped per §10.

Two nice touches: the positional core parse (`"1.x.2" → (1,0,2)`, not `(1,2,0)`) and factoring it into a testable `SemVer` / `compare_semver` instead of one opaque function.

Test coverage hits exactly the cases that bite: `alpha.11 > alpha.9` (the numeric-vs-string regression), `alpha < beta < rc`, `numeric < alphanumeric`, longer-set-wins, and build-metadata-ignored.

**Verified locally** (pure version-string logic, no arch sensitivity): `cargo test -p genie-core --lib ota::` → **11 passed**.

LGTM — nice fix.

PR #453, opened 2026-06-22
fix(security): make injection scanner robust to punctuation separators

### LINKED ISSUE #452: Prompt-injection scanner is defeated by punctuation — `ignore, previous, instructions` and `ignore-previous-instructions` slip past every word pattern
**Labels:** bug, security

### Summary

`security::injection::scan` in `crates/genie-core/src/security/injection.rs` normalizes input with `normalize` (lowercase + collapse ASCII whitespace) and then does an exact **contiguous-substring** match against each pattern:

```rust
fn normalize(text: &str) -> String {
    text.to_lowercase().split_whitespace().collect::<Vec<_>>().join(" ")
}

pub fn scan(text: &str) -> InjectionCheck {
    let normalized = normalize(text);
    for pattern in PATTERNS {
        if normalized.contains(pattern.text) { /* Suspicious */ }
    }
    InjectionCheck::Clean
}
```

The module's own doc claims it normalizes "to prevent case-based and whitespace-based evasion" — but it does **not** handle punctuation. Any separator between the words of a phrase — a comma, hyphen, dot, slash, parenthesis — breaks the contiguous substring, so the classic override phrasing sails straight through. This is the same evasion class as the IV-2 case-sensitivity fix this file already adopted, one step further (punctuation, not just case/whitespace).

### Steps to reproduce

Confirmed against the real `scan` logic (standalone reproduction), pattern `"ignore previous instructions"`:

| Input | `scan` today |
|---|---|
| `ignore, previous, instructions` | **Clean** (missed) |
| `ignore-previous-instructions` | **Clean** (missed) |
| `ignore...previous...instructions` | **Clean** (missed) |
| `ignore/previous/instructions` | **Clean** (missed) |
| `please IGNORE previous instructions` | Suspicious ✓ (case-only control) |

So an attacker (or an injected web/HA payload routed through `scan_and_warn`) only has to sprinkle punctuation between the keywords to suppress the telemetry entirely. A secondary gap: trivially reworded variants with a filler word — `ignore all previous instructions`, `ignore the previous instructions` — are also missed, since none of them is the exact substring `ignore previous instructions`.

### Expected behavior

Word-boundary separators do not defeat detection. `ignore, previous, instructions`, `ignore-previous-instructions`, and `ignore . previous . instructions` all flag, exactly as `ignore previous instructions` does. Symbol patterns whose punctuation is meaningful (`rm -rf`, `curl | sh`, `eval(`) keep matching, and benign prose (`please evaluate the options`) is **not** newly flagged.

### Actual behavior

- `injection.rs` `normalize` only folds case and ASCII whitespace; punctuation between words is preserved, so `contains("ignore previous instructions")` fails on any punctuated variant.
- The pattern table stores only the bare phrasing, so common filler-word rewrites are not covered.

### Hardware

Hardware-independent — pure string scanning. Reproduces on any host via a direct `scan(...)` call. Affects every input surface wired through `scan_and_warn` (`api-chat`, `api-chat-stream`, `voice`, `voice-followup`, `repl`, `openai-bridge`).

### JetPack / L4T version

`main @ 77a600c` (upstream HEAD as of filing) / `crates/genie-core/src/security/injection.rs`.

### Relevant logs

`scan_and_warn` is an observability control — the symptom is the **absence** of a warning that should have fired:

```text
# Benign-looking case still fires:
WARN source=api-chat reason="override: matched 'ignore previous instructions'" prompt injection pattern detected
# Punctuated evasion fires NOTHING — the telemetry simply goes silent:
#   (no warn for "ignore, previous, instructions")
```

### Additional context

**Why this matters:** `scan_and_warn` is the project's injection-telemetry surface across every entry point that reaches the LLM (issue #196 wired the `source` tags for exactly this). A detector that any commenter can bypass with a comma gives false confidence — the dashboards look quiet precisely when a real injection attempt is in flight.

**Why the existing tests don't catch it:** every positive test uses clean spacing (`"ignore   previous   instructions"`, `"IGNORE PREVIOUS INSTRUCTIONS"`) — the case/whitespace cases that already work. None inserts punctuation between the words.

**Proposed fix (single surface — `injection.rs` only):**

- Split normalization into two views and pick per pattern:
  - `normalize_words` — lowercase and fold **every** run of non-alphanumeric characters to a single space (`ignore-previous-instructions` → `ignore previous instructions`). Used for natural-language phrase patterns.
  - `normalize_raw` — the existing lowercase + whitespace-collapse, which **preserves** symbols. Used for shell/operator patterns (`rm -rf`, `curl | sh`, `eval(`) so folding punctuation doesn't both break them and false-flag benign words like "evaluate".
- Tag each `Pattern` with a `MatchMode` (`Words` / `Raw`) and match against the corresponding view. Add `const fn` constructors so the table stays readable.
- Expand the pattern table with the common filler-word override rewrites (`ignore all/the/any/prior/above previous instructions`, `disregard …`, `forget …`) and a few more extraction/exfiltration phrasings — each is one more literal, no logic change.

**Tests to add (`injection::tests`, all hardware-free):**
- `punctuation_separators_do_not_evade` — comma / hyphen / dot / slash / parenthesis variants all flag.
- `filler_word_override_variants_are_detected` — `ignore all/the/any previous instructions`, `disregard the previous instructions`, `forget all previous instructions`.
- `benign_words_are_not_false_flagged` — `"please evaluate the options"`, ordinary commands stay `Clean` (guards the `eval(` Raw pattern against punctuation-folding).
- `raw_shell_patterns_still_match_with_symbols` — `rm -rf …`, `eval(payload)` still flag.
- Existing `detects_*` / `whitespace_normalization_prevents_evasion` / `clean_input` stay green.

**Acceptance (PR):**
- `cargo test -p genie-core --lib security::injection` — existing tests pass, new tests pass.
- `cargo fmt --check`, `cargo clippy -p genie-core` (no new warnings), `cargo test --workspace --no-default-features` — green.

**Real Behavior Proof:** `cargo test -p genie-core --lib security::injection` showing the new evasion/false-positive tests passing, plus the before/after table reproduced from the real `scan` code.

**Related:**
- IV-2 (case-sensitivity normalization, already in this file) — this is the punctuation-evasion successor in the same hardening line.
- `#451 / fix(ota): SemVer precedence`, `#444 / fix(memory): clause-boundary` (mine) — same "naive matcher handles the happy path but misses the trivially-varied real input" class.

galuis116:
## Summary

Fixes a punctuation-evasion gap in the prompt-injection scanner. `scan` normalized only case + ASCII whitespace, then matched each pattern as an exact **contiguous substring**, so any separator between the words of a phrase defeated detection:

| Input | `scan` before |
|---|---|
| `ignore, previous, instructions` | **Clean** (missed) |
| `ignore-previous-instructions` | **Clean** (missed) |
| `ignore/previous/instructions` | **Clean** (missed) |
| `please IGNORE previous instructions` | Suspicious ✓ (case-only control) |

`scan_and_warn` is the injection-telemetry surface wired across every input entry point (`api-chat`, `api-chat-stream`, `voice`, `voice-followup`, `repl`, `openai-bridge`), so a detector any commenter can bypass with a comma gives false confidence. This extends the file's existing IV-2 case/whitespace normalization to punctuation.

Closes #452

## Changes

- `security/injection.rs`: split normalization into two views, picked per pattern:
  - `normalize_words` — lowercase and fold every run of non-alphanumeric characters to a single space (`ignore-previous-instructions` → `ignore previous instructions`). Used for natural-language phrase patterns.
  - `normalize_raw` — the existing lowercase + whitespace-collapse, which **preserves** symbols. Used for shell/operator patterns (`rm -rf`, `curl | sh`, `eval(`) so folding punctuation doesn't break them or false-flag benign words like "evaluate".
- Tag each `Pattern` with a `MatchMode` (`Words` / `Raw`) and match against the corresponding view; add `const fn` constructors so the table stays readable.
- Expand the pattern table with the common filler-word override rewrites (`ignore all/the/any/prior/above previous instructions`, `disregard …`, `forget …`) and a few more extraction/exfiltration phrasings — each is one more literal, no logic change.
- No public signatures change; `scan` / `scan_and_warn` keep their shapes.

## Real Behavior Proof

- [ ] Verified on Jetson hardware.
- [x] NOT verified on Jetson — pure string scanning with no hardware/network dependency, exercised deterministically by unit tests on any host.

Tested profile: `laptop` (x86_64 Linux, Rust stable).

```
cargo test -p genie-core --lib security::injection
cargo clippy -p genie-core --lib
```

```
running 13 tests (security::injection)
test security::injection::tests::punctuation_separators_do_not_evade ... ok
test security::injection::tests::filler_word_override_variants_are_detected ... ok
test security::injection::tests::benign_words_are_not_false_flagged ... ok
test security::injection::tests::raw_shell_patterns_still_match_with_symbols ... ok
# existing detects_* / whitespace_normalization_prevents_evasion / clean_input still green
test result: ok. 13 passed; 0 failed; 0 ignored; 0 measured; 720 filtered out

test result: ok. 733 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

## Test plan

- `cargo test -p genie-core --lib security::injection` — new evasion/false-positive regressions pass alongside the existing case/whitespace tests.
- Behavioral check: `scan("ignore, previous, instructions")` now returns `Suspicious`, while `scan("please evaluate the options")` stays `Clean`.

## Notes for reviewers

- The scanner remains observability-only (warns, does not block) — this PR widens detection coverage without changing the control's behavior contract.
- `Words` mode only tightens matching around real phrases (punctuation-robust); it is not loosened to arbitrary word gaps, so it doesn't introduce gap-matching false positives. The `benign_words_are_not_false_flagged` test guards the `eval(` Raw pattern specifically.
- Kept entirely within `security/injection.rs` to stay single-surface.

### DIFF
--- a/crates/genie-core/src/security/injection.rs
+++ b/crates/genie-core/src/security/injection.rs
@@ -17,13 +17,28 @@ pub enum InjectionCheck {
 
 /// Scan text for prompt injection patterns.
 ///
-/// Normalizes input (lowercase, collapse whitespace) before matching
-/// to prevent case-based and whitespace-based evasion.
+/// Two normalized views of the input are matched against, picked per pattern:
+///
+/// - **Word patterns** (natural-language phrases) match against a view that
+///   folds *every* non-alphanumeric run — punctuation, hyphens, dots, slashes,
+///   zero-width separators — down to a single space. This closes the
+///   separator-evasion gap where `ignore, previous, instructions` or
+///   `ignore-previous-instructions` slipped past the old contiguous-substring
+///   match. (Extends the IV-2 case/whitespace normalization to punctuation.)
+/// - **Raw patterns** (shell/operator fragments such as `rm -rf` or `eval(`)
+///   match against a lowercase, whitespace-collapsed view that *preserves*
+///   symbols, so the punctuation that makes them meaningful is not folded away
+///   and benign words like "evaluate" are not flagged.
 pub fn scan(text: &str) -> InjectionCheck {
-    let normalized = normalize(text);
+    let words = normalize_words(text);
+    let raw = normalize_raw(text);
 
     for pattern in PATTERNS {
-        if normalized.contains(pattern.text) {
+        let haystack = match pattern.mode {
+            MatchMode::Words => &words,
+            MatchMode::Raw => &raw,
+        };
+        if haystack.contains(pattern.text) {
             return InjectionCheck::Suspicious(format!(
                 "{}: matched '{}'",
                 pattern.category, pattern.text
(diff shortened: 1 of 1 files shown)


ai-hpc (approved):
**Approve** ✅

Strong security fix. The gap was real and the detector is wired across every input path, so a comma-bypassable scanner gave false confidence. The two-view design is the right call:

- `normalize_words` folds every non-alphanumeric run to a single space, so `ignore, previous, instructions` / `ignore-previous-instructions` / `ignore...previous...instructions` all collapse to the canonical phrase — closing the separator evasion.
- `normalize_raw` *preserves* symbols for shell/operator patterns, so `eval(` still matches while `evaluate` does not — avoiding the obvious false-positive class.

I like that the filler-word variants (`ignore the/all/any/above previous instructions`) are **enumerated explicitly** rather than matched with loose gaps — that keeps it a contiguous-substring match (high precision) instead of opening an over-matching hole. And since `scan_and_warn` is warn-only telemetry, leaning toward recall is the right tradeoff: the residual cost of a false flag is a log line, not a blocked request.

Tests are exactly right: the six separator-evasion forms, the filler-word variants, and a benign set that pins the no-false-positive cases (notably `evaluate` ≠ `eval(`).

**Verified locally** (pure string-scanner logic, no arch sensitivity): `cargo test -p genie-core --lib security::injection` → **13 passed**.

One small, non-blocking note: folding punctuation means a cross-sentence phrase like `"… ignore. Previous instructions were unclear …"` would now flag — but it's warn-only telemetry and that phrasing is injection-adjacent anyway, so the tradeoff is fine. Worth a mental note if this detector ever becomes a hard block.

LGTM — nice hardening.

PR #459, opened 2026-06-22
fix(eval): keep BFCL canon_action from collapsing activate into turn_on

### LINKED ISSUE #458: [bug] BFCL grounded metric credits turn_on for an expected activate (canon_action collapses activate into turn_on)
## Summary

The BFCL grounded metric credits a wrong actuation: predicting `turn_on` for an expected `home_control` `activate` scores as a grounded match, because `eval::bfcl::canon_action` folds `activate` into `turn_on` while the runtime dispatcher keeps them distinct.

## Steps to reproduce

`canon_action` (crates/genie-core/src/eval/bfcl.rs) canonicalizes action args for the grounded metric and maps `activate` → `turn_on`:

```rust
"activate" | "enable" | "switch_on" | "power_on" | "turn_on" => "turn_on".to_string(),
```

But the runtime `tools::dispatch::canon_home_control_action` leaves `activate` as its own action (it is a separate value in the `home_control` action enum and is covered by a `("activate", Some("activate")) // distinct action, must not remap` test).

Reproduce with the CLI:

```
# cases.jsonl
{"id":"scene-activate","category":"home","prompt":"Activate movie night scene","expected_tool_calls":[{"tool":"home_control","arguments":{"entity":"kitchen lights","action":"activate"}}]}
# preds.jsonl  (model predicted turn_on instead of activate)
{"id":"scene-activate","response":"{\"tool\":\"home_control\",\"arguments\":{\"entity\":\"kitchen lights\",\"action\":\"turn_on\"}}"}

genie-ctl bfcl-score --cases cases.jsonl --predictions preds.jsonl --json
```

## Expected behavior

`grounded_argument_matches` / `grounded_strict_matches` = `0`. Predicting `turn_on` for an expected `activate` is a different actuation (a no-op `turn_on` does not activate the scene) and must not be credited. The doc comment on `canon_action` states it "mirrors the canonicalization the runtime tool dispatch should apply" — the runtime keeps `activate` distinct.

## Actual behavior

```
"grounded_argument_matches": 1,
"grounded_strict_matches": 1,
```

The wrong actuation is scored as correct, which inflates the grounded accuracy the M1 harness gates on. The fixtures already contain an `activate` case (`multi-homework-timer`), so this affects real scoring.

## Hardware

## GenieClaw version / commit

main @ a2cf324 (v1.0.0-alpha.11 workspace)

## Additional context

Fix is to drop `activate` from the `turn_on` arm of `canon_action` so it passes through unchanged, mirroring `canon_home_control_action`. PR with a regression test and before/after `bfcl-score` proof to follow.

carlos4s:
## Summary

The BFCL grounded metric scored a wrong actuation as correct: predicting `home_control` `turn_on` for an expected `activate` counted as a grounded match. `eval::bfcl::canon_action` folded `activate` into `turn_on`, but the runtime dispatcher keeps `activate` a distinct action. This PR aligns the scorer with the runtime. Fixes #458.

## Changes

- `canon_action` no longer maps `activate` into `turn_on`; it passes through normalized, mirroring `tools::dispatch::canon_home_control_action` (which leaves `activate` as-is and is covered by a "must not remap" test).
- Updated the doc comment to state why `activate` stays distinct.
- Added regression test `grounded_metric_keeps_activate_distinct_from_turn_on` asserting the grounded metric does not credit `turn_on` for an expected `activate`.

## Real Behavior Proof

- [x] `laptop`

This change is to the BFCL scoring metric (host-side eval), not a runtime/hardware path, so a Jetson run is not applicable — the scorer is deterministic and produces identical results on any host. Verified on x86_64 Linux.

### What I ran

Built `genie-ctl` and scored a one-case fixture where the expected action is `activate` and the prediction emits `turn_on`:

```
# cases.jsonl
{"id":"scene-activate","category":"home","prompt":"Activate movie night scene","expected_tool_calls":[{"tool":"home_control","arguments":{"entity":"kitchen lights","action":"activate"}}]}
# preds.jsonl
{"id":"scene-activate","response":"{\"tool\":\"home_control\",\"arguments\":{\"entity\":\"kitchen lights\",\"action\":\"turn_on\"}}"}

genie-ctl bfcl-score --cases cases.jsonl --predictions preds.jsonl --json
```

Also ran the BFCL unit suite: `cargo test -p genie-core --lib eval::bfcl`.

### What I observed

`bfcl-score` before vs after:

| metric | before (main) | after |
|---|---|---|
| `grounded_argument_matches` | 1 | 0 |
| `grounded_strict_matches` | 1 | 0 |

Before the fix the wrong actuation was credited; after, it is correctly rejected with diagnostic `tool[0] argument mismatch: $.action expected "activate", got "turn_on"`. The non-grounded `argument_matches` was already `0` in both cases.

All 15 `eval::bfcl` tests pass, including the existing `grounded_metric_credits_action_synonym` (deactivate→turn_off still credited) and `loads_jsonl_fixture_and_scores_report` (21-case fixture unchanged).

## Test plan

1. `cargo test -p genie-core --lib eval::bfcl` — all green, new test included.
2. Reproduce the before/after with the fixture above via `genie-ctl bfcl-score`.

### DIFF
--- a/crates/genie-core/src/eval/bfcl.rs
+++ b/crates/genie-core/src/eval/bfcl.rs
@@ -1079,6 +1079,27 @@ mod tests {
     }
 
     #[test]
+    fn grounded_metric_credits_matching_activate() {
+        // Positive complement to grounded_metric_keeps_activate_distinct_from_turn_on:
+        // a correctly-predicted `activate` must still be credited — the fix passes
+        // `activate` through, it does not drop it.
+        let case = case(vec![expected(
+            "home_control",
+            serde_json::json!({"action": "activate", "entity": "kitchen lights"}),
+        )]);
+        let response = r#"{"tool":"home_control","arguments":{"action":"activate","entity":"kitchen lights"}}"#;
+        let score = score_response(&case, response);
+        assert!(
+            score.argument_match,
+            "exact activate match must be credited"
+        );
+        assert!(
+            score.grounded_argument_match,
+            "a matching activate must be credited by the grounded metric"
+        );
+    }
+
+    #[test]
     fn loads_jsonl_fixture_and_scores_report() {
         let root = Path::new(env!("CARGO_MANIFEST_DIR")).join("../..");
         let cases = load_cases_jsonl(root.join("tests/bfcl/home_tool_cases.jsonl")).unwrap();
--- a/crates/genie-core/src/eval/bfcl.rs
+++ b/crates/genie-core/src/eval/bfcl.rs
@@ -599,14 +599,21 @@ fn canonicalize_entity_value(value: &Value, graph: &HomeGraph) -> Value {
 /// Canonicalize an action verb to a small set of stable forms, so the grounded
 /// metric credits synonyms the agent would execute identically. Unknown verbs
 /// pass through normalized (lowercase, separators -> `_`) so exact matches still
-/// hold. Mirrors the canonicalization the runtime tool dispatch should apply.
+/// hold. Mirrors the canonicalization the runtime tool dispatch applies in
+/// `tools::dispatch::canon_home_control_action`.
+///
+/// `activate` is intentionally *not* folded into `turn_on`: the runtime keeps it
+/// as a distinct `home_control` action for scenes/scripts (it appears separately
+/// in the action enum and `canon_home_control_action` leaves it as-is), so
+/// crediting `turn_on` for an expected `activate` would score a wrong actuation
+/// as correct and inflate the grounded metric.
 fn canon_action(text: &str) -> String {
     let normalized = text.trim().to_lowercase().replace([' ', '-'], "_");
     match normalized.as_str() {
         "deactivate" | "disable" | "switch_off" | "power_off" | "shut_off" | "turn_off" => {
             "turn_off".to_string()
         }
-        "activate" | "enable" | "switch_on" | "power_on" | "turn_on" => "turn_on".to_string(),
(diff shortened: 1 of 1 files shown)


ai-hpc (approved):
**Approve** ✅

Correct, honest fix — and a good contrast to the over-crediting direction. The scorer's `canon_action` was folding `activate → turn_on`, so predicting `turn_on` for an expected `activate` scored as a grounded match and inflated the metric. Aligning it with the runtime `canon_home_control_action` (which keeps `activate` distinct) is exactly right: the grounded metric should credit a prediction only when the runtime would actuate the same thing.

I pushed two test-only commits on top:

- **`canon_action_contract`** (`53d3418`) — exhaustive lock of the mapping: turn_on/turn_off synonyms fold, `activate` stays distinct, every other `home_control` action passes through, and normalization (case + space/hyphen) holds. So a future edit can't silently re-introduce the `activate → turn_on` collapse.
- **`grounded_metric_credits_matching_activate`** (`57618a8`) — the positive complement to your negative test: a correctly-predicted `activate` is still credited (exact + grounded), confirming the passthrough doesn't *under*-credit legitimate activates.

Both verified locally: `cargo test -p genie-core --lib eval::bfcl` → **17 passed**.

One non-blocking idea: the highest-leverage test against this whole class of bug would be a drift guard asserting the scorer's `canon_action` agrees with the runtime's `canon_home_control_action` for every action + synonym — but that needs exposing `canon_home_control_action` (it's module-private), so I left it as a suggestion rather than bundling a visibility change into your PR.

LGTM — good catch, and it makes the metric more trustworthy.

PR #464, opened 2026-06-22
perf(memory): skip open-time derived-table rebuild when derivation version is unchanged

jakearmstrong59:
## Summary

`Memory::open` re-derives every derived table (household profiles, aliases, rules, notes, secret references, media items, calendar, shopping, inventory, access permissions, task logs, schedule, event logs, embeddings) and the FTS index from the canonical `memories` rows on **every open**. #442 made that pass cheaper by batching its writes into one transaction; this goes further and **skips it entirely when nothing relevant changed**. Those derived tables are already maintained live on every mutation (`store_with_metadata` upserts each derived row, deletes cascade, the `memories_a*` triggers keep FTS current), so the full re-derivation is only needed when the derivation logic itself changes. A `DERIVATION_VERSION` stamp (in a new `memory_meta` table) is written after a rebuild; while the stored stamp matches the code's version, the rebuild is skipped — a single integer comparison instead of a full re-scan and re-embed of every memory. Measured **~2.0× faster `Memory::open` (median 1.166 s → 0.577 s with 2000 memories)** on the Jetson Orin Nano, on top of #442. Contributes under #402 (performance bucket).

## Changes

- `crates/genie-core/src/memory/mod.rs`:
  - New `memory_meta(key, value)` table.
  - `const DERIVATION_VERSION` (bump when any `rebuild_*` logic or the embedding model changes, or a derived table is added).
  - `Memory::open` reads the stored `derivation_version`; only when it differs from `DERIVATION_VERSION` does it run the (transaction-batched) re-derivation of all 15 derived tables + FTS, then stamps the version. Otherwise the rebuild is skipped.
  - Test `recall_works_after_reopen_skips_rebuild` proving recall is correct after a reopen that skips the rebuild (complements the existing `semantic_embeddings_rebuild_on_reopen`).

## Real Behavior Proof

- [x] `jetson`
- [ ] `raspberry_pi`
- [ ] `portable_sbc`
- [x] `laptop`
- [ ] `mac`
- [ ] CI-only / docs-only
- [ ] Not run locally

### What I ran

On the target **Jetson Orin Nano** (Engineering Reference Developer Kit Super, 8 GB), L4T R36.4.7, `Linux 5.15.148-tegra aarch64`, `rustc 1.95.0`, **release** profile (`opt-level="z"`, LTO disabled only to speed the build). Cloned current `main` (base `56b1b04`, which already includes #442) into a tmpfs target dir; the `memory_open_bench` benchmark (added in #442) seeds 2000 memories once, then times `Memory::open` over 10 opens; the database is on the SD card. Ran the bench on `main`, then applied this change and reran (3× each):

```
```

Also on an x86_64 dev box (`rustc 1.95.0`): full `memory::` suite (169 tests) green, `cargo fmt -p genie-core -- --check` and `cargo clippy -p genie-core --tests` clean.

### What I observed

`Memory::open`, 2000 memories, 10 opens per run:

| run | before (main + #442 batched rebuild) | after (skip when version matches) |
| --- | --- | --- |
| 1 | 1.189 s/open | 0.577 s/open |
| 2 | 1.052 s/open | 0.590 s/open |
| 3 | 1.166 s/open | 0.517 s/open |
| **median** | **1.166 s** | **0.577 s** |

- **~2.0× faster** open at the median (the skipped re-derivation accounts for ~590 ms). On top of #442 (~2.57×), this is ~6× versus the original per-open full rebuild.
- Behavior unchanged: the derived tables and FTS are maintained live, so a reopen that skips the rebuild still serves correct results. The 169-test `memory::` suite passes on x86_64 and aarch64, including `recall_works_after_reopen_skips_rebuild` and `semantic_embeddings_rebuild_on_reopen`.

### Test plan

- `cargo test -p genie-core --lib memory::`
- `cargo test -p genie-core --release --test memory_open_bench -- --ignored --nocapture` on `main` vs this branch to reproduce the table.

### Notes for reviewers

- Builds on #442 (already merged): when re-derivation does run (fresh DB, or after a `DERIVATION_VERSION` bump), it uses #442's single transaction.
- Trade-off: a database mutated out-of-band (a recovery tool writing rows without going through this code) is not re-derived until `DERIVATION_VERSION` is bumped. Normal store/update/delete keep the derived tables in sync, so this only affects external edits; bump the constant whenever derivation logic or the embedding model changes.
- The remaining ~0.58 s/open is the schema-ensure + index-migration pass that still runs every open (`CREATE TABLE/INDEX IF NOT EXISTS`, `backfill_policy_columns`), out of scope here — a candidate for a follow-up.

### DIFF
--- a/crates/genie-core/src/memory/mod.rs
+++ b/crates/genie-core/src/memory/mod.rs
@@ -16100,6 +16100,30 @@ mod tests {
     }
 
     #[test]
+    fn derived_table_survives_reopen_skip() {
+        // The open-time rebuild is skipped when DERIVATION_VERSION matches, so a
+        // non-FTS derived table must be kept live on store (not by the rebuild).
+        // Store a relationship (populates household_profiles), reopen — which
+        // skips the rebuild since the version matches — and confirm the derived
+        // row is still there.
+        let path = temp_memory_path("derived-reopen-skip");
+        {
+            let mem = Memory::open(&path).unwrap();
+            mem.store("relationship", "Jared is the dad").unwrap();
+            assert_eq!(mem.household_profiles_by_role("father").unwrap().len(), 1);
+        }
+        let reopened = Memory::open(&path).unwrap();
+        let profiles = reopened.household_profiles_by_role("father").unwrap();
+        assert_eq!(
+            profiles.len(),
+            1,
+            "household_profiles must survive a rebuild-skipping reopen"
+        );
+        assert_eq!(profiles[0].name, "Jared");
+        assert_eq!(profiles[0].role, "dad");
+    }
+
+    #[test]
     fn semantic_embeddings_rebuild_on_reopen() {
         let path = temp_memory_path("semantic-reopen");
         {
--- a/crates/genie-core/src/memory/mod.rs
+++ b/crates/genie-core/src/memory/mod.rs
@@ -16,6 +16,8 @@ use std::time::Duration;
 
 const MAX_QUERY_HASHES: usize = 16;
 
+const DERIVATION_VERSION: i64 = 1;
+
 /// Persistent conversational memory with dreaming-inspired consolidation.
 ///
 /// Architecture (inspired by OpenClaw's memory-core, clean-room Rust):
@@ -561,6 +563,11 @@ impl Memory {
             CREATE INDEX IF NOT EXISTS idx_embedded_memories_type
                 ON embedded_memories(memory_type, updated_ms DESC);
 
+            CREATE TABLE IF NOT EXISTS memory_meta (
+                key   TEXT PRIMARY KEY,
+                value INTEGER NOT NULL
+            );
+
             CREATE VIRTUAL TABLE IF NOT EXISTS memories_fts USING fts5(
                 content,
                 content='memories',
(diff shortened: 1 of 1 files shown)


matedev01 (approved):
**Approve** ✅

Good optimization on top of #442 — and the skip is safe for normal operation. I checked the load-bearing assumption: `store_with_metadata` fans out to all 15 `upsert_*_from_memory` calls (one per derived table, mirroring the `rebuild_*` set) and the `memories_a*` triggers keep FTS current, so the derived tables genuinely stay consistent on every store/delete without the open-time rebuild. The version gate handles the edges: a missing version (old or fresh DB) or a mismatch forces a full rebuild and records the new version, all inside the existing transaction — so a failed rebuild rolls back and re-runs next open.

I pushed two test-only commits (your test covers FTS/search; these cover the rest):

- `derived_table_survives_reopen_skip` — a non-FTS derived table (`household_profiles`) is still correct after a reopen that *skips* the rebuild, proving live-maintenance (not the rebuild) keeps it.
- `version_mismatch_reopen_rebuilds_derived_tables` — wiping a derived table + resetting the stored version forces the rebuild to re-derive it on reopen (the other half of the gate).

Both pass locally: `cargo test -p genie-core --lib memory::tests` → **111 passed**.

Two non-blocking caveats, since this removes the open-rebuild safety net:

1. **Version-bump discipline.** Any change to a `rebuild_*` function or a derived-table schema now *must* bump `DERIVATION_VERSION`, or existing DBs silently keep stale derivations. There's no enforcement — a short comment by the `rebuild_*` block pointing at `DERIVATION_VERSION` (and ideally a test that fails if a derived-table schema changes without a bump) would protect future edits.
2. **External-edit self-healing is gone.** The comment you removed noted the open-rebuild healed DBs "edited by a recovery tool" or predating the FTS trigger. Old DBs are still handled (no version → one rebuild), but a mid-life external edit that bypasses the upserts/triggers now persists stale, since the version still matches. Reasonable trade for the perf win — just a behavior change to flag.

LGTM.

PR #466, opened 2026-06-22
perf(memory): store embeddings as packed f32 BLOB instead of JSON text

jaso0n0818:
## Summary

The `embedded_memories.embedding` column was `TEXT`, decoded with `serde_json::from_str::<Vec<f32>>()` on every row scanned during `semantic_search`. Replace with a little-endian packed BLOB: 4 bytes per dimension, decoded by a single `chunks_exact(4)` pass — no JSON tokenizer, no intermediate string allocation.

Because `rebuild_embedded_memories` clears and rewrites the entire table on every `Memory::open`, no migration is needed.

## Changes

- `embedded_memories.embedding` schema: `TEXT` → `BLOB`
- `upsert_embedded_memory_from_memory`: pack `Vec<f32>` to `Vec<u8>` via `to_le_bytes()` instead of `serde_json::to_string`
- `semantic_search`: read column as `Vec<u8>` instead of `String`
- `parse_embedding`: decode bytes with `chunks_exact(4)` instead of JSON parse

## Tests

- `parse_embedding_blob_roundtrip_is_bit_identical` — embed → pack → unpack must produce the same `Vec<f32>`
- `parse_embedding_rejects_wrong_length` — truncated or dimension-mismatched byte slices return `None`
- All 748 existing unit tests pass

### DIFF
--- a/crates/genie-core/src/memory/mod.rs
+++ b/crates/genie-core/src/memory/mod.rs
@@ -554,7 +554,7 @@ impl Memory {
                 memory_type      TEXT NOT NULL,
                 embedding_model  TEXT NOT NULL,
                 dimensions       INTEGER NOT NULL,
-                embedding        TEXT NOT NULL,
+                embedding        BLOB NOT NULL,
                 updated_ms       INTEGER NOT NULL
             );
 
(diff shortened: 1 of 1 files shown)


matedev01:
Closing this for the reason the **PR body checklist** is already failing: there's no **Real Behavior Proof**. Per the contribution scope (#402), a performance PR needs a **measurable Jetson before→after** — this has unit tests (good, and the roundtrip/length tests are the right ones), but no measured win, so the actual improvement on the target can't be confirmed.

To be clear, the change itself is sound: packing embeddings as a little-endian f32 BLOB and decoding with `chunks_exact(4)` is genuinely cheaper than `serde_json::from_str::<Vec<f32>>` per row in `semantic_search`, and since `rebuild_embedded_memories` rewrites the table on open, no migration is needed. Good idea.

Because it's worth landing, I'll open a follow-up PR with this optimization **plus a measured Jetson benchmark** (JSON vs packed-BLOB decode, before→after), crediting this PR. If you'd rather carry it yourself, reopen with the before→after numbers and we'll take it.

Closing per #402 (no Real Behavior Proof).

PR #475, opened 2026-06-23
perf(memory): gate open-time schema-ensure + migrations on PRAGMA user_version

jakearmstrong59:
## Summary

`Memory::open` re-ran the **entire schema-ensure pass on every open** — the big `CREATE TABLE/INDEX/TRIGGER IF NOT EXISTS` batch, the 14 `ALTER`/`CREATE INDEX` `run_open_migration` calls, and the policy-column backfill — even when the on-disk schema was already current. With the derived-table rebuild already gated by #464, that ensure pass is the **entire remaining steady-state reopen cost** (~99.8% of it on Jetson). This change gates it on SQLite's built-in `PRAGMA user_version`: run the ensure pass only when the stored version is below `SCHEMA_VERSION`, then stamp the version; skip it otherwise. The result is **~575× faster `Memory::open`** on steady-state restarts on the Jetson Orin Nano (483 ms → 0.84 ms). Contributes under #402 (performance bucket).

## Changes

- `crates/genie-core/src/memory/mod.rs` (`open_with_half_life`):
  - Split the **per-connection PRAGMAs** (`journal_mode`/`synchronous`/`temp_store`/`foreign_keys`) out of the big schema batch into their own `execute_batch` that still runs on **every** open — they are connection state, not persisted schema.
  - Gate the schema-ensure pass (the `CREATE … IF NOT EXISTS` batch + the 14 `run_open_migration` calls + `backfill_policy_columns`) behind `if schema_version < SCHEMA_VERSION { … }`, reading `PRAGMA user_version` (defaulting to `0`).
  - Stamp `PRAGMA user_version = SCHEMA_VERSION` after the pass, **only when `!migration_degraded`** — so a transient migration failure is not locked in and self-heals on the next open.
  - New `const SCHEMA_VERSION: i64 = 1;` next to `DERIVATION_VERSION`.
  - The separate #464 `derivation_version` rebuild gate is unchanged and still runs afterward.
- `crates/genie-core/src/memory/mod.rs` (tests): `stale_schema_version_reopen_reensures_and_restamps` — opens a DB, asserts a successful open stamps `user_version`, then drops a migration-created index and resets `user_version`, and asserts the reopen re-runs the ensure pass (recreates the index) and re-stamps the version.

The line count of the diff is inflated by re-indentation: the existing schema-ensure statements are now nested in the version-gate `if`, so `rustfmt` shifts them. `git diff --ignore-all-space` shows the real change is the gate + the PRAGMA split + the stamp guard.

## Real Behavior Proof

- [x] `jetson`
- [ ] `raspberry_pi`
- [ ] `portable_sbc`
- [x] `laptop`
- [ ] `mac`
- [ ] CI-only / docs-only
- [ ] Not run locally

### What I ran

On the target **Jetson Orin Nano** (Engineering Reference Developer Kit Super, 8 GB), L4T R36.4.7, `Linux 5.15.148-tegra aarch64`, `rustc 1.95.0`, **release** profile (`opt-level="z"`, LTO disabled only to speed the build), DB on the SD card. Cloned current `main` into a tmpfs target dir, ran the existing `memory_open_bench` (added in #464 — seeds 2000 memories, then times 10 reopens of the same DB), then applied this change and reran (3× each):

```
```

Also on an x86_64 dev box (`rustc 1.95.0`): the same bench (median 209 ms → 1.35 ms/open, ~155×); `cargo test -p genie-core --lib memory::` → **174 tests green** (incl. the new regression test); `cargo fmt -p genie-core -- --check`, `cargo clippy -p genie-core --tests`, and `cargo clippy -p genie-core --no-default-features --tests` all clean.

### What I observed

`Memory::open` steady-state reopen, 2000 memories, 10 opens/run (the rebuild is already skipped by #464, so this isolates the schema-ensure cost):

| run | before (schema-ensure every open) | after (gated on user_version) |
| --- | --- | --- |
| 1 | 468.95 ms/open | 0.842 ms/open |
| 2 | 483.26 ms/open | 0.840 ms/open |
| 3 | 493.45 ms/open | 0.837 ms/open |
| **median** | **483 ms** | **0.84 ms** |

- **~575× faster** `Memory::open` on the Jetson at the median; ~155× on x86_64. The schema-ensure pass was ~99.8% of steady-state reopen cost on the Jetson.
- Behavior unchanged on the paths that matter: fresh DBs (`user_version` defaults to `0`) and legacy/upgraded DBs run the full idempotent ensure exactly as before, then skip on subsequent opens. Verified by the existing suite — `recall_works_after_reopen_skips_rebuild`, `version_mismatch_reopen_rebuilds_derived_tables`, `open_backfills_policy_columns_for_existing_rows`, `open_marks_migration_degraded_when_fts_rebuild_fails` — plus the new `stale_schema_version_reopen_reensures_and_restamps`.

### Test plan

- `cargo test -p genie-core --lib memory::` (open/reopen/migration/backfill behavior unchanged; 174 green)
- `cargo test -p genie-core --release --test memory_open_bench -- --ignored --nocapture` on `main` vs this branch to reproduce the table.

## Notes for reviewers

- **Self-healing preserved.** `run_open_migration` swallows non-duplicate failures (disk full, lock, corrupt page) into the in-memory `migration_degraded` flag. The version is stamped **only when `!migration_degraded`**, so a degraded open leaves `user_version` below `SCHEMA_VERSION` and the next open re-runs the idempotent ensure — the same self-healing the per-open ensure gave, without re-running it in the steady state.
- **Monotonic gate.** The check is `schema_version < SCHEMA_VERSION` (not `!=`), so an older binary opening a DB written by a newer binary will not downgrade the stamp.
- **Maintenance contract:** bump `SCHEMA_VERSION` whenever the schema DDL or any `run_open_migration` call changes, otherwise already-stamped DBs will skip the new schema. This mirrors the existing `DERIVATION_VERSION` discipline from #464.
- Per-connection PRAGMAs (`foreign_keys`, `synchronous`, `temp_store`, `journal_mode`) still run on every open — they were moved out of the gated batch, not gated.
- Out of scope (hardening, needs out-of-band schema tampering to hit): the FTS table/triggers are `CREATE … IF NOT EXISTS` inside the gated pass, so they are no longer re-asserted on every open; `health().fts_consistent` still surfaces row-count drift.
- No new behavior, config, or data migration; no schema content change.

### DIFF
--- a/crates/genie-core/src/memory/mod.rs
+++ b/crates/genie-core/src/memory/mod.rs
@@ -17,6 +17,7 @@ use std::time::Duration;
 const MAX_QUERY_HASHES: usize = 16;
 
 const DERIVATION_VERSION: i64 = 1;
+const SCHEMA_VERSION: i64 = 1;
 
 /// Persistent conversational memory with dreaming-inspired consolidation.
 ///
@@ -302,12 +303,19 @@ impl Memory {
         )?;
         conn.busy_timeout(Duration::from_secs(5))?;
         conn.execute_batch(
-            "
-            PRAGMA journal_mode = WAL;
-            PRAGMA synchronous = NORMAL;
-            PRAGMA temp_store = MEMORY;
-            PRAGMA foreign_keys = ON;
+            "PRAGMA journal_mode = WAL;
+             PRAGMA synchronous = NORMAL;
+             PRAGMA temp_store = MEMORY;
+             PRAGMA foreign_keys = ON;",
+        )?;
 
+        let mut migration_degraded = false;
+        let schema_version: i64 = conn
+            .query_row("PRAGMA user_version", [], |row| row.get(0))
+            .unwrap_or(0);
+        if schema_version < SCHEMA_VERSION {
+            conn.execute_batch(
+                "
             CREATE TABLE IF NOT EXISTS memories (
                 id            INTEGER PRIMARY KEY,
                 kind          TEXT NOT NULL,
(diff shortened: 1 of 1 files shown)


matedev01:
Merged — nice work. One small follow-up worth keeping in mind (not blocking, since it's already in):

The version gate means a migration now only reaches existing DBs if `SCHEMA_VERSION` is bumped past their stamped value. Right now `const SCHEMA_VERSION: i64 = 1;` carries no note about that, so it'd be easy for a future change to add a 15th `run_open_migration(...)` step and forget the bump — existing DBs at `user_version = 1` would silently skip it.

A one-line doc comment on the constant would lock the invariant, e.g.:

```rust
/// Bump whenever a schema-ensure / `run_open_migration` step is added, so existing
/// DBs re-run the ensure pass. Stamped into `PRAGMA user_version` after a clean open.
const SCHEMA_VERSION: i64 = 1;
```

Happy to do it in a tiny follow-up if you'd rather not.

PR #473, opened 2026-06-23
fix(dispatch): action_history shows undo_restore hints for value-changing actions (#471)

kiannidev:
Fixes #471

## Summary

`action_history` now mirrors `last_undoable`: value-changing home actions (`set_brightness`, `set_temperature`, `toggle`) show an `undo:` hint derived from `undo_restore` instead of **not undoable** when `inverse_action` is absent. Closes the UX gap left after #434 / #435 where `home_undo` worked in-session but history misled the model and dashboard users.

## Changes

- Add `RecordedAction::action_history_undo_hint()` in `actuation.rs` — prefers `inverse_action`, then formats `undo_restore` (e.g. `undo: set_brightness 100`, `undo: turn_off`).
- `exec_action_history` delegates to that helper instead of only checking `inverse_action`.
- Unit tests for hint formatting plus `action_history_shows_undo_restore_hint_after_dim` integration test (`RecordingHomeProvider` stub).

## Real Behavior Proof

- [x] I have built and run the affected code locally (or noted why I could not).
- [ ] I have verified the change end-to-end on Jetson hardware.
- [x] I have NOT verified on Jetson hardware, and I explain the equivalent verification path or validation gap below.

Tested profile / hardware (check all that apply):

- [ ] `jetson`
- [ ] `raspberry_pi`
- [ ] `portable_sbc`
- [x] `laptop`
- [ ] `mac`
- [ ] CI-only / docs-only
- [ ] Not run locally

### What I ran

Linux x86_64 dev host (`main` base + this branch), no live Home Assistant:

```bash
cd genie-claw
cargo test -p genie-core action_history -- --nocapture
cargo clippy -p genie-core -- -D warnings
```

### What I observed

All `action_history`-related tests pass. The new integration test exercises the full dispatch path: `home_control` turn_on → `set_brightness` 30 → `action_history`. Output includes `set_brightness kitchen light` with `undo: set_brightness 100` (prior brightness 255 → 100%) and does **not** contain `not undoable` for that entry. Existing `turn_on` hydrate test still shows `undo: turn_off`. Clippy clean.

**Validation gap:** Not exercised on Jetson or against a live HA instance. This change is display-only in `exec_action_history`; undo execution logic is unchanged and already covered by `home_undo_restores_prior_brightness_after_dim`. Real HA would surface the same ledger fields; restart/hydrate undo hints remain #472.

## Test plan

- `cargo test -p genie-core action_history`
- `cargo clippy -p genie-core -- -D warnings`
[...]

PR #460, opened 2026-06-22
fix(home): restore off state on home_undo after set_temperature (#455)

kiannidev:
Fixes #455

## Summary

Mirror the `set_brightness` off-state handling in `undo_restore_from_prior` for `set_temperature`. When the HVAC was `off` before a setpoint change, `home_undo` now records `turn_off` as the restore action instead of leaving `undo_restore` empty (which skipped the thermostat change and reversed an unrelated earlier action).

## Changes

- `undo_restore_from_prior("set_temperature")`: prior `state == "off"` → `UndoRestore { action: "turn_off" }`.
- Unit test: off vs heating prior states.
- Dispatch test: off thermostat → `set_temperature(72)` → `home_undo` → `turn_off` (climate back off).
- Extend `RecordingHomeProvider` stub with climate domain for the integration test.

## Real Behavior Proof

- [x] I have built and run the affected code locally (or noted why I could not).
- [ ] I have verified the change end-to-end on Jetson hardware.
- [x] I have NOT verified on Jetson hardware, and I explain the equivalent verification path or validation gap below.

Tested profile / hardware (check all that apply):

- [ ] `jetson`
- [ ] `raspberry_pi`
- [ ] `portable_sbc`
- [x] `laptop`
- [ ] `mac`
- [ ] CI-only / docs-only
- [ ] Not run locally

### What I ran

```bash
cd genie-claw
git checkout fix/issue-455-home-undo-set-temperature-off

cargo fmt
cargo test -p genie-core undo_restore_from_prior_set_temperature
cargo test -p genie-core home_undo_restores_off_state_after_set_temperature
cargo test -p genie-core home_undo
cargo clippy -p genie-core -- -D warnings
```

Environment: x86_64 Linux dev host (`laptop` profile). Stub provider test + unit tests; optional live HA repro below.

Optional live HA repro:

```bash
docker compose -f deploy/homeassistant/docker-compose.yml up -d
# Ensure climate entity is off with no active setpoint exposed
genie-ctl chat "set the thermostat to 72"
genie-ctl chat "undo the last home action"
# Expect: HVAC returns to off, not an unrelated earlier action
```

### What I observed

**Before (`main`):**

- `undo_restore_from_prior("set_temperature", prior_off)` → `None`.
- `home_undo` after heating an off thermostat could skip the temperature change.

**After (this PR):**

```text
$ cargo test -p genie-core undo_restore_from_prior_set_temperature
[...]

PR #435, opened 2026-06-20
fix(home): restore prior state on home_undo for set_brightness and value actions (#434)

kiannidev:
Fixes #434

`home_undo` now targets the most recent home actuation, including `set_brightness`, `set_temperature`, and `toggle`. Value-changing actions capture a pre-actuation HA state snapshot and restore it on undo instead of skipping to an older binary flip.

## Summary

- Add `UndoRestore` metadata to `RecordedAction` so the action ledger can invert value-changing actions, not just binary flips.
- Before executing `set_brightness`, `set_temperature`, or `toggle`, `exec_home_control_inner` reads the target's prior HA state and stores the restore call in the ledger.
- `home_undo` builds `home_control` args via `ActionLedger::undo_home_control_args`, including `value` when restoring brightness or temperature.
- `last_undoable()` considers entries with `undo_restore` as well as `inverse_action`.

## Real Behavior Proof

- [x] I have built and run the affected code locally (or noted why I could not).
- [ ] I have verified the change end-to-end on Jetson hardware.
- [x] I have NOT verified on Jetson hardware, and I explain the equivalent verification path or validation gap below.

Tested profile / hardware (check all that apply):

- [ ] `jetson`
- [ ] `raspberry_pi`
- [ ] `portable_sbc`
- [x] `laptop`
- [ ] `mac`
- [ ] CI-only / docs-only
- [ ] Not run locally

### What I ran

```bash
cd genie-claw
git checkout fix/issue-434-home-undo-value-actions

cargo fmt
cargo test -p genie-core home_undo
cargo test -p genie-core undo_restore
cargo test -p genie-core action_ledger_undo
cargo clippy -p genie-core -- -D warnings
```

Environment: x86_64 Linux dev host (`laptop` profile). Jetson not available; live HA repro below uses `deploy/homeassistant/`.

Optional live HA repro (issue #434 steps):

```bash
docker compose -f deploy/homeassistant/docker-compose.yml up -d
# wire HA token + [services.homeassistant] per deploy/homeassistant/README.md

curl -s -H "Authorization: Bearer $HA_TOKEN" \
  http://127.0.0.1:8123/api/states/light.kitchen_lights | jq '{state, brightness: .attributes.brightness}'

genie-ctl chat "turn on the kitchen lights"
genie-ctl chat "set the kitchen lights brightness to 30 percent"

curl -s -H "Authorization: Bearer $HA_TOKEN" \
[...]

PR #481, opened 2026-06-23
fix(home): require unanimous group undo_restore snapshot (#480)

### LINKED ISSUE #480: [bug] home_undo group targets snapshot only first entity prior state (#402)
### Summary

`undo_restore_from_prior` captures pre-action state from `prior.entities.first()` only. For Home Assistant **group/area** targets (`HomeTargetKind::Group`), `get_state` returns all member entities (order = HA `/api/states` scan order). When group members have mixed power or brightness, `home_undo` after `set_brightness` or `toggle` can restore the wrong action for the whole group — e.g. `turn_off` because the first member happened to be off, while another member was on at 80%.

### Steps to reproduce

1. From repo root on `main` (`4f3f9ca` or later):

   ```bash
   cd genie-claw
   cargo test -p genie-core undo_restore_from_prior_captures_brightness_percent -- --nocapture
   ```

   Existing tests cover **single-entity** prior state only.

2. **Unit repro (add failing test):**

   ```rust
   use genie_core::tools::actuation::undo_restore_from_prior;
   use genie_core::ha::{Entity, HomeState};

   let prior = HomeState {
       entities: vec![
           Entity { entity_id: "light.a".into(), state: "off".into(), attributes: json!({}), .. },
           Entity { entity_id: "light.b".into(), state: "on".into(),
                    attributes: json!({"brightness": 204}), .. }, // ~80%
       ],
       ..Default::default()
   };

   let restore = undo_restore_from_prior("set_brightness", &prior).unwrap();
   // Bug: restore.action == "turn_off" (from light.a, the first entity)
   // Expected: per-entity restore, aggregate semantics, or explicit failure — not whole-group turn_off
   ```

3. **Dispatch sequence repro (extend `RecordingHomeProvider` to return 2 entities):**

   ```text
   home_control set_brightness 30 on "living room lights"  # group with mixed member states
   home_undo
   Bug: issues turn_off for entire group instead of restoring light.b to ~80%
   ```

4. **Optional live HA** (`deploy/homeassistant/` sample config):

   ```bash
   # Confirm group member states differ before dimming an area group
   curl -s -H "Authorization: Bearer $HA_TOKEN" \
     http://127.0.0.1:8123/api/states/light.<member_a> | jq .state
   curl -s -H "Authorization: Bearer $HA_TOKEN" \
     http://127.0.0.1:8123/api/states/light.<member_b> | jq .state

   genie-ctl chat "dim the living room lights to 30 percent"
   genie-ctl chat "undo the last home action"
   # Observe: wrong member states after undo when members were heterogeneous
   ```

5. **Code paths:**

   - `undo_restore_from_prior` (`actuation.rs` ~341–342): `let entity = prior.entities.first()?`
   - `exec_home_control_inner` (`dispatch.rs` ~1136–1140): calls `undo_restore_from_prior(action, &prior)` on full `HomeState` from `ha.get_state(&target)`
   - `get_live_entities` (`ha/provider.rs` ~588–595): returns all group members, order not guaranteed stable

### Expected behavior

- Group `home_undo` after value-changing actions restores pre-action state correctly for **all** affected members, or fails explicitly when members disagree.
- `undo_restore` metadata must not depend on arbitrary first-entity ordering from HA state scans.

### Actual behavior

| Prior group state | Action | `undo_restore` captured | `home_undo` effect |
|-------------------|--------|-------------------------|-------------------|
| `light.a` off, `light.b` on @ 80% | `set_brightness` 30 | `turn_off` (from first entity) | Whole group off — **light.b not restored to 80%** |
| `light.a` on, `light.b` off | `toggle` | `turn_on` (from first entity) | Whole group on — **light.b should stay off** |

Single-entity targets work; group/area targets with heterogeneous members are wrong.

### Hardware

Non-Jetson (x86_64 dev / cross-build host). Reproducible with `RecordingHomeProvider` stub extended to 2 entities; real HA optional for Real Behavior Proof in PR.

### JetPack / L4T version

_No response_

`main` @ `4f3f9ca`

### Relevant logs

_No response_

### Additional context

**Contribution fit ([#402](https://github.com/GeniePod/genie-claw/issues/402) / M1):** Real-HA tool-dispatch correctness — area groups are first-class `home_control` targets via `resolve_group_target`, but undo metadata is single-entity. Distinct from closed #434 (single-entity value restore) and from audit-hydrate persistence (separate concern).

**Suggested fix directions:**

1. Store per-`entity_id` undo snapshots for `HomeTargetKind::Group`.
2. Refuse undo when member states are heterogeneous (conservative).
3. Capture aggregate semantics (e.g. count of members on) instead of `entities.first()`.

Include unit test with 2-entity stub and optional live-HA before→after on a real area group.

kiannidev:
Fixes #480

## Summary

`undo_restore_from_prior` no longer derives group undo metadata from `prior.entities.first()` alone. For multi-entity group targets, every member must produce the same restore snapshot; heterogeneous member states skip `undo_restore` capture so `home_undo` cannot issue a wrong group-wide `turn_off` (or other action) based on arbitrary HA state ordering.

## Changes

- Extract per-entity restore logic into `undo_restore_from_entity`.
- `undo_restore_from_prior` requires unanimous restore across all group members; returns `None` when members disagree.
- Add `PartialEq` to `UndoRestore` for snapshot comparison.
- Unit tests: mixed group → `None`; unanimous group → shared `set_brightness` restore.
- Integration test: heterogeneous group dim is recorded as not undoable; `home_undo` fails safely instead of executing a bogus restore.

## Real Behavior Proof

- [ ] `jetson`
- [ ] `raspberry_pi`
- [ ] `portable_sbc`
- [x] `laptop`
- [ ] `mac`
- [ ] CI-only / docs-only
- [ ] Not run locally

### What I ran

Linux x86_64 dev host, no live Home Assistant (`MixedGroupHomeProvider` stub with `HomeTargetKind::Group`):

```bash
cd genie-claw
cargo test -p genie-core --lib undo_restore_from_prior
cargo test -p genie-core --lib home_undo_skips_heterogeneous
cargo clippy -p genie-core -- -D warnings
```

### What I observed

- `undo_restore_from_prior_mixed_group_brightness_returns_none` — mixed off/on members → no `undo_restore` for `set_brightness` or `toggle`.
- `undo_restore_from_prior_unanimous_group_brightness_succeeds` — both members on @ 204 → `set_brightness ~80%` restore.
- `home_undo_skips_heterogeneous_group_dim_without_wrong_turn_off` — dim on heterogeneous group records `not undoable` in `action_history`; `home_undo` returns "No recent reversible" with only one `SetBrightness` executed (no erroneous `TurnOff`).
- Clippy clean.

**Validation gap:** Not run on Jetson or live HA area group. Stub exercises the same `get_state` → `undo_restore_from_prior` → ledger → `home_undo` path as production group targets.

## Test plan

- `cargo test -p genie-core --lib undo_restore_from_prior`
- `cargo test -p genie-core --lib home_undo_skips_heterogeneous`
- `cargo test -p genie-core --lib home_undo_restores_prior_brightness_after_dim` (regression: single entity unchanged)
- `cargo clippy -p genie-core -- -D warnings`

## Notes for reviewers

- Conservative fix: heterogeneous groups are not undoable via single `home_control` restore (same limitation as before, but no longer **wrong**). Per-member undo would need a follow-up schema change.
- Single-entity behavior unchanged.

### DIFF
--- a/crates/genie-core/src/tools/actuation.rs
+++ b/crates/genie-core/src/tools/actuation.rs
@@ -63,7 +63,7 @@ pub struct PendingConfirmation {
     pub expires_ms: u64,
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub struct UndoRestore {
     pub action: String,
     #[serde(skip_serializing_if = "Option::is_none")]
@@ -337,9 +337,8 @@ fn inverse_action(action: &str) -> Option<&'static str> {
     }
 }
 
-/// Capture the `home_control` call that restores `prior` before `action` ran.
-pub fn undo_restore_from_prior(action: &str, prior: &HomeState) -> Option<UndoRestore> {
-    let entity = prior.entities.first()?;
+/// Capture the `home_control` call that restores one entity's pre-action state.
+fn undo_restore_from_entity(action: &str, entity: &crate::ha::Entity) -> Option<UndoRestore> {
     match action {
         "set_brightness" => {
             if entity.state == "off" {
@@ -380,6 +379,27 @@ pub fn undo_restore_from_prior(action: &str, prior: &HomeState) -> Option<UndoRe
         "toggle" => undo_restore_for_power_state(&entity.state),
         _ => None,
     }
+}
+
+/// Capture the `home_control` call that restores `prior` before `action` ran.
+///
+/// For group targets with multiple member entities, every member must agree on
+/// the same restore snapshot; heterogeneous member states return `None` so
+/// `home_undo` does not apply a single group-wide action derived from an
+/// arbitrary first entity (#480).
+pub fn undo_restore_from_prior(action: &str, prior: &HomeState) -> Option<UndoRestore> {
+    let first = prior.entities.first()?;
+    let baseline = undo_restore_from_entity(action, first)?;
+    if prior.entities.len() == 1 {
+        return Some(baseline);
+    }
+
+    prior
+        .entities
+        .iter()
+        .skip(1)
+        .all(|entity| undo_restore_from_entity(action, entity).as_ref() == Some(&baseline))
+        .then_some(baseline)
 }
 
 fn undo_restore_for_power_state(state: &str) -> Option<UndoRestore> {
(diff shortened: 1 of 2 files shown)


matedev01:
Thanks @kiannidev — the instinct here is sound: a group target whose members are in different states shouldn't get a single group-wide undo derived from whichever entity happens to sort first, and the unanimity check (`undo_restore_from_prior` returning `None` when members disagree) is a clean way to express that. The fix itself is correct.

I'm going to close this one, though, for a few reasons:

1. **Scope / diminishing returns.** This is the fifth home_undo refinement in the series (#435, #460, #473, #474, now #480/#481), each targeting a narrower corner. A *group* target with *heterogeneous* member states being undone is an extreme edge — it needs a group action, mixed member states, and an undo all at once. Against current M1 priorities, the value-to-complexity is low.

2. **Diff weight.** Most of the 219 lines are test scaffolding — including a brand-new full `HomeProvider` stub stood up for a single integration test, the same heavy-fixture pattern we trimmed on #460. The actual behavior change is ~20 lines.

3. **Marginal safety delta.** The current first-entity snapshot yields an imperfect (not dangerous) group undo; skipping capture on disagreement is a reasonable refinement, but not urgent enough to carry the added surface area right now.

None of this is a knock on the reasoning — it's a real corner and the fix is correct. If group-undo correctness becomes a priority, this is worth revisiting, ideally with just the focused `undo_restore_from_prior` unit test you already added (no full provider stub). Closing for now — thanks again for the careful eye on the undo path.

PR #483, opened 2026-06-23
perf(voice): single-pass TTS text cleanup in for_voice

jakearmstrong59:
## Summary

`for_voice` cleans every spoken assistant reply before TTS, and it was a multi-pass, allocation-heavy pipeline. The biggest offender, `clean_for_tts`, chained **twelve** whole-buffer `String::replace` calls (one allocation each); `strip_raw_urls` built a `Vec<&str>` then `join`'d it; and `truncate_for_voice` collected a `Vec<String>` of per-sentence clones then `join`'d. This collapses the **order-independent** work into single passes while preserving the **order-sensitive** rewrites exactly, so the output is byte-for-byte identical. Measured **~1.14–1.27× faster** on the Jetson Orin Nano. Contributes under #402 (performance bucket).

## Changes

- `crates/genie-core/src/voice/format.rs`:
  - `clean_for_tts`: keep the four ordered multi-char replaces (`...`, ` - `, ` — `, ` – ` — their overlapping-space behavior is order-dependent) and the trailing possessive `'s`→`s`, but fold the seven single-char rules (`(`/`)`→`, `, and dropping `[`/`]`/`{`/`}`/`"`) into **one** char-scan pass. 12 allocations → 6.
  - `strip_raw_urls`: stream the kept tokens straight into the result `String` instead of building a `Vec<&str>` and `join`-ing it.
  - `truncate_for_voice`: accumulate the kept sentences into a single `String` with a running count instead of a `Vec<String>` + per-sentence `to_string()` clones + `join`.
- Tests: the 16 existing `for_voice` tests only exercise the public API, so they move to `crates/genie-core/tests/voice_format_test.rs` (integration). `crates/genie-core/tests/voice_format_bench.rs` adds an ignored microbench. Both are gated `#![cfg(feature = "voice")]` so the `--no-default-features` build stays green.

## Real Behavior Proof

- [x] `jetson`
- [ ] `raspberry_pi`
- [ ] `portable_sbc`
- [x] `laptop`
- [ ] `mac`
- [ ] CI-only / docs-only
- [ ] Not run locally

### What I ran

On the target **Jetson Orin Nano** (Engineering Reference Developer Kit Super, 8 GB), L4T R36.4.7, `Linux 5.15.148-tegra aarch64`, `rustc 1.95.0`, **release** profile (`opt-level="z"`, LTO disabled only to speed the build). Cloned current `main` into a tmpfs target dir, added the new ignored microbench (`voice_format_bench.rs`, two representative replies — a short one-liner and a markdown/list/link/URL/ellipsis-heavy long reply), ran it on `main`, then applied this change and reran (3× each):

```
cargo test -p genie-core --release --test voice_format_bench -- --ignored --nocapture
```

**Correctness was proven separately and exhaustively:** a differential fuzz harness ran the new `for_voice` against the original over **400,000 randomized inputs** (drawn from a charset of markdown markers, dashes, brackets, quotes, CJK terminators, URLs, digits and whitespace) plus a fixed markdown/punctuation corpus, asserting **byte-identical** output. It caught three real edge cases during development (cross-rule `'s`/`"` adjacency, hyphen/em-dash shared-space priority, and `**`/`__` removal ordering) — which is why the order-sensitive replaces are kept intact. The 16 relocated tests pass, and `cargo fmt --check` + `cargo clippy --tests` + `cargo clippy --no-default-features --tests` are clean.

### What I observed

`for_voice`, 300,000 calls/run, 3 runs each (medians shown):

| corpus | before (multi-pass) | after (single-pass) | speedup |
| --- | --- | --- | --- |
| short (117 B) | 14.71 µs/call | 11.62 µs/call | ~1.27× |
| long (439 B) | 31.16 µs/call | 27.42 µs/call | ~1.14× |

- Stable across runs (before short 14.64–14.71 µs; after short 11.61–11.62 µs). The gain is the eliminated allocations: `clean_for_tts` 12 whole-buffer passes → 6, plus the dropped `Vec` in `strip_raw_urls` and the dropped `Vec<String>` + per-sentence clones in `truncate_for_voice`.
- Output is unchanged (differential fuzz + the 16 tests). On an x86_64 dev box the same bench shows ~1.21× (short) / ~1.14× (long).

### Test plan

- `cargo test -p genie-core --test voice_format_test` (behavior unchanged; 16 tests)
- `cargo test -p genie-core --release --test voice_format_bench -- --ignored --nocapture` on `main` vs this branch to reproduce the table.

## Notes for reviewers

- **Scope is honest:** `for_voice` runs once per reply, not per audio frame, so this is a steady efficiency win on the reply path (fewer allocations, lower fragmentation), not a headline end-to-end latency change. It removes real redundant work with a reproducible on-device delta.
- **Behavior-preserving by construction:** the order-sensitive rewrites (the four multi-char replaces and possessive `'s`) are kept as-is precisely because fusing them changes output in edge cases; only the order-independent single-char rules and the `Vec`/`join` assembly were rewritten. The differential fuzz is the guarantee.
- **Test move:** `for_voice`'s tests are black-box (public API only), so they belong as an integration test in `tests/`; this also lets the microbench live beside them. No coverage is lost.
- No new behavior, config, or schema.

### DIFF
--- a/crates/genie-core/src/voice/format.rs
+++ b/crates/genie-core/src/voice/format.rs
@@ -132,20 +132,26 @@ fn strip_links(text: &str) -> String {
 }
 
 fn strip_raw_urls(text: &str) -> String {
-    text.split_whitespace()
-        .filter(|token| {
-            let trimmed = token.trim_matches(|c: char| {
-                matches!(
-                    c,
-                    '(' | ')' | '[' | ']' | '{' | '}' | ',' | ';' | ':' | '"' | '\''
-                )
-            });
-            !(trimmed.starts_with("http://")
-                || trimmed.starts_with("https://")
-                || trimmed.starts_with("www."))
-        })
-        .collect::<Vec<_>>()
-        .join(" ")
+    let mut result = String::with_capacity(text.len());
+    for token in text.split_whitespace() {
+        let trimmed = token.trim_matches(|c: char| {
+            matches!(
+                c,
+                '(' | ')' | '[' | ']' | '{' | '}' | ',' | ';' | ':' | '"' | '\''
+            )
+        });
+        if trimmed.starts_with("http://")
+            || trimmed.starts_with("https://")
+            || trimmed.starts_with("www.")
+        {
+            continue;
+        }
+        if !result.is_empty() {
+            result.push(' ');
+        }
+        result.push_str(token);
+    }
+    result
 }
 
 /// Normalize whitespace: collapse multiple spaces, trim.
@@ -200,7 +206,8 @@ fn truncate_for_voice(text: &str, max_sentences: usize) -> String {
     }
 
     let chars: Vec<char> = text.chars().collect();
-    let mut sentences = Vec::new();
+    let mut result = String::with_capacity(text.len());
+    let mut sentence_count = 0;
     let mut current = String::new();
 
     for (i, &ch) in chars.iter().enumerate() {
@@ -227,21 +234,28 @@ fn truncate_for_voice(text: &str, max_sentences: usize) -> String {
         };
 
         if ends_sentence {
-            sentences.push(current.trim().to_string());
+            if !result.is_empty() {
+                result.push(' ');
+            }
+            result.push_str(current.trim());
+            sentence_count += 1;
             current.clear();
-            if sentences.len() >= max_sentences {
+            if sentence_count >= max_sentences {
                 break;
             }
         }
     }
 
     // Include the trailing fragment if we still have room.
-    let trailing = current.trim().to_string();
-    if !trailing.is_empty() && sentences.len() < max_sentences {
-        sentences.push(trailing);
+    let trailing = current.trim();
+    if !trailing.is_empty() && sentence_count < max_sentences {
+        if !result.is_empty() {
+            result.push(' ');
+        }
+        result.push_str(trailing);
     }
 
-    sentences.join(" ")
+    result
 }
 
 fn is_ascii_terminator(ch: char) -> bool {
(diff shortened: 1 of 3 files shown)


matedev01 (approved):
To be sure rather than just reason about it, I ran a **differential test** (old 12-replace impl vs the new `clean_for_tts`) over an edge-case battery — parens, brackets, quotes, all three dashes (`-`/`—`/`–`), possessives, adjacent/nested specials, unicode, empty: **0 byte-for-byte divergences**.

matedev01:
**Approve** ✅

Behavior-preserving, and I confirmed it byte-for-byte. All three collapses are sound:

- `strip_raw_urls` / `truncate_for_voice`: same split/filter and sentence-collection logic, just building the `String` directly instead of via `Vec` + `join` — equivalent by inspection.
- `clean_for_tts`: the 12 replaces are restructured as four sequential multi-char replaces (`pre`), then a single char-scan for the single-char ones, then `'s` last — preserving the original relative order. The single-char replaces are mutually order-independent (no replacement's `", "` output or char-removal forms another's pattern), so the single-pass scan equals the sequential chain.

**Verified on Jetson Orin Nano**

matedev01 (approved):
**Approve** ✅

Behavior-preserving, and I confirmed it byte-for-byte. All three collapses are sound:

- `strip_raw_urls` / `truncate_for_voice`: same split/filter and sentence-collection logic, just building the `String` directly instead of via `Vec` + `join` — equivalent by inspection.
- `clean_for_tts`: the 12 replaces are restructured as four sequential multi-char replaces (`pre`), then a single char-scan for the single-char ones, then `'s` last — preserving the original relative order. The single-char replaces are mutually order-independent (no replacement's `", "` output or char-removal forms another's pattern), so the single-pass scan equals the sequential chain.

To be sure rather than just reason about it, I ran a **differential test** (old 12-replace impl vs the new `clean_for_tts`) over an edge-case battery — parens, brackets, quotes, all three dashes (`-`/`—`/`–`), possessives, adjacent/nested specials, unicode, empty: **0 byte-for-byte divergences**.

**Verified locally** (pure string logic, no arch sensitivity): `voice_format_test` → **16 passed**; the `voice::format` lib tests and the differential check pass; `cargo fmt` / clippy clean. Only the three functions changed — the markdown stripping is untouched.

The ~1.14–1.27× is modest but real (12 `String::replace` allocations → 4 + one single-pass scan; `Vec`+`join` → direct `String`), measured on the Jetson with a reusable bench. Clean perf work.

PR #484, opened 2026-06-23
perf(memory): const relationship patterns + early-out in extract_facts

jakearmstrong59:
## Summary

`extract_facts` runs on every user utterance (Tier-1 auto-capture). Its `extract_relationships` helper rebuilt **57 `format!` pattern strings on every call** — 3 phrasings × 19 relations, each a heap allocation — and then scanned the text for each, *even when the utterance contained no relationship phrase at all*. This replaces the per-call `format!` loop with a `const` table of `&'static str` patterns (byte-identical to the old `format!` output) and adds an early-out: the text is already lowercased at the call site and every pattern begins with `"my "`, so when the utterance contains no `"my "` the whole 19-relation scan is skipped. Measured **~6.3× faster on the common (no-relationship) utterance** and **~1.65× when `"my "` is present**, on the Jetson Orin Nano. Output is unchanged. Contributes under #402 (performance bucket).

## Changes

- `crates/genie-core/src/memory/extract.rs` (`extract_relationships`):
  - Hoist the 57 per-call `format!("my {relation} is named ", …)` patterns into a `const RELATIONSHIP_PATTERNS: &[(&str, [&str; 3])]` table of fixed `&'static str`s — same relations, same phrasing order, so matched substrings and the pushed `(relation, name)` pairs are identical. Eliminates 57 heap allocations per call.
  - Early-out `if !text.contains("my ") { return … }`. Safe and behavior-preserving: the caller passes the already-lowercased text and every pattern starts with `"my "`, so no match is possible without it.
- Tests: the 24 existing tests only exercise the public `extract_facts`/`extract_and_store` API, so they move to `crates/genie-core/tests/extract_test.rs` (integration). `crates/genie-core/tests/extract_bench.rs` adds an ignored microbench.

## Real Behavior Proof

- [x] `jetson`
- [ ] `raspberry_pi`
- [ ] `portable_sbc`
- [x] `laptop`
- [ ] `mac`
- [ ] CI-only / docs-only
- [ ] Not run locally

### What I ran

On the target **Jetson Orin Nano** (Engineering Reference Developer Kit Super, 8 GB), L4T R36.4.7, `Linux 5.15.148-tegra aarch64`, `rustc 1.95.0`, **release** profile (`opt-level="z"`, LTO disabled only to speed the build). Cloned current `main` into a tmpfs target dir, added the new ignored microbench (`extract_bench.rs`, three inputs — a no-`"my "` utterance, a `"my "`-but-no-match utterance, and a real relationship phrase), ran it on `main`, then applied this change and reran (3× each):

```
cargo test -p genie-core --release --test extract_bench -- --ignored --nocapture
```

**Correctness was verified directly:** the 57 const patterns are exactly the old `format!` output — a check confirmed all 76 string literals in the table (19 relation tags + 57 patterns) equal the `format!` templates in order — so matched substrings and returned pairs are byte-identical. The early-out cannot change results (all patterns start with `"my "`, input is pre-lowercased). The 24 relocated tests pass, and `cargo fmt --check` + `cargo clippy --tests` + `cargo clippy --no-default-features --tests` are clean.

### What I observed

`extract_facts`, 300,000 calls/run, 3 runs each (medians):

| input | before (`format!` loop) | after (const table + early-out) | speedup |
| --- | --- | --- | --- |
| no `"my "` (common) | 13.82 µs/call | 2.20 µs/call | ~6.3× |
| `"my "`, no match | 13.65 µs/call | 8.25 µs/call | ~1.65× |
| real relationship | 13.61 µs/call | 8.33 µs/call | ~1.63× |

- The common case (most utterances have no relationship phrase) is **~6.3×** faster — the early-out skips the 19-relation scan and all 57 allocations. When `"my "` is present the const table still removes the 57 per-call allocations for **~1.65×**.
- Output unchanged (24 tests + the literal-equivalence check). On an x86_64 dev box the same bench shows ~5.7× / ~1.67×.

### Test plan

- `cargo test -p genie-core --test extract_test` (behavior unchanged; 24 tests)
- `cargo test -p genie-core --release --test extract_bench -- --ignored --nocapture` on `main` vs this branch to reproduce the table.

## Notes for reviewers

- **Scope is honest:** `extract_facts` runs once per utterance (warm), not per audio frame, so this saves microseconds per spoken sentence — a real allocation/scan elimination with a reproducible on-device delta, not a user-perceptible latency change.
- **Byte-identical by construction:** the const strings equal the old `format!` output (verified), and the early-out is provably safe given the pre-lowercased input and the `"my "` prefix on every pattern.
- **Test move:** the tests are black-box (public API only), so they belong as an integration test in `tests/`; this also lets the microbench live beside them. No coverage lost.
- I intentionally left `to_lowercase` (required for case-insensitive matching) and the existing favorite-pattern lookup unchanged. No new behavior, config, or schema.


matedev01:
**Approve** ✅

Behavior-preserving, and I confirmed it byte-for-byte. The `const RELATIONSHIP_PATTERNS` table reproduces the old `format!` output exactly — 19 relations × 3 phrasings = 57 patterns, every one `"my {relation} (is named|name is|is called) "` — and the inner find/extract logic is unchanged. The early-out is safe: every pattern begins with `"my "` and the text is lowercased at the call site, so `!text.contains("my ")` can only skip utterances where no pattern could possibly match.

To be sure rather than just eyeball the table, I ran a **differential test** (old `format!` loop vs the new const-table impl) over an edge battery — multiple relations in one utterance, possessives, mid-string matches, the `"myself"` early-out edge, empty / empty-name, hyphen / apostrophe / unicode names: **0 divergences**.

PR #490, opened 2026-06-24
fix(quick-router): parse the full spoken-number range for durations

statxc:
## Summary

The deterministic quick-router's spoken-duration parser only recognized a sparse, hand-picked set of number words, so common phrasings silently failed to set a timer. "set a timer for **thirteen** minutes" did nothing, because `thirteen`, `fourteen`, `sixteen`, `seventeen`, `eighteen`, and `nineteen` were missing from the word list — and there was no support for `hundred`/`thousand` at all. This replaces the sparse lookup with a real English cardinal parser.

## Changes

- The old `parse_number` mapped a sparse word set (it lacked the teens above twelve except fifteen), and an ad-hoc `is_tens_word`/`ones_digit` special case stitched only tens+ones compounds — nothing past ninety-nine.
- New `parse_spoken_number(tokens, start)` consumes a maximal run of number words and folds units, teens, tens, `hundred`, `thousand`, and the connector `and` into a single value, returning the value and the index just past the run. `cardinal_word` is the single classifier.
- `parse_duration` now drives that parser and then checks a time unit follows.
- Digit forms (`"10 minutes"`) and every previously handled phrasing — including the `"forty five"` compound — are preserved; new coverage adds the missing teens and arbitrary hundreds/thousands (e.g. `"one hundred and twenty seconds"`).
- All arithmetic is saturating, so adversarial input can't overflow.

## Real Behavior Proof

- [x] `laptop`

This is a deterministic, hardware-independent change to the quick-router's duration parsing. No Jetson, audio, model, or network dependency — an x86_64 Linux dev box exercises the exact code path that runs on-device, and the parser is fully covered by unit tests.

### What I ran

```
cargo test -p genie-core
cargo test -p genie-core --lib tools::quick::tests::routes
```

I added the regression tests first and confirmed they **fail on the current code** before applying the fix (e.g. `'thirteen minutes' should route` panicked).

### What I observed

- Before: `route("set a timer for thirteen minutes")` returned `None` (no timer); hundreds/thousands likewise unhandled.
- After:
  - `thirteen / fourteen / sixteen / seventeen / eighteen / nineteen minutes` → correct `n * 60` seconds.
  - `"one hundred seconds"` → 100; `"two hundred thirty seconds"` → 230; `"one hundred and twenty seconds"` → 120; `"one hundred twenty minutes"` → 7200; `"one thousand seconds"` → 1000; `"ninety nine seconds"` → 99.
  - Existing phrasings unchanged: `"10 minutes"` → 600, `"forty five minutes"` → 2700, `"twenty five seconds"` → 25, `"fifty minutes"` → 3000, and the labelled-reminder boundary still extracts `"check the oven"`.
- `cargo test -p genie-core` — **787 passed, 0 failed** (1 pre-existing ignored), so no regression.

## Test plan

1. `cargo test -p genie-core --lib tools::quick::tests::routes` — all routing tests pass, including `routes_teen_worded_durations` and `routes_hundreds_and_thousands_worded_durations`.
2. Manually route "set a timer for thirteen minutes" and confirm `set_timer { seconds: 780 }`.

## Notes for reviewers

`parse_number`, `is_tens_word`, and `ones_digit` were used only by `parse_duration`, so this is fully contained to duration parsing — no other routing path changes. The parser handles "and" only as an in-number connector (when a number word follows), so a trailing conjunction like "set a timer for ten minutes and turn on the lights" is unaffected (the unit is matched before "and" is reached).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## Unreleased
 
+- **Full spoken-number durations** (#490): the quick-router only mapped a sparse
+  set of number words, so "set a timer for thirteen minutes" set no timer
+  (thirteen, fourteen, sixteen–nineteen were missing, and hundreds/thousands
+  were unsupported). A real English cardinal parser now folds units, teens,
+  tens, `hundred`, `thousand`, and the `and` connector into one value, so
+  "one hundred and twenty seconds" and the like resolve correctly. Digit forms
+  and the existing "forty five" compound are unchanged.
+
 ## 1.0.0-rc.1 - 2026-06-24
 
 First release candidate — and the first **installable** build. GenieClaw can now
--- a/crates/genie-core/src/tools/quick.rs
+++ b/crates/genie-core/src/tools/quick.rs
@@ -2389,75 +2389,105 @@ fn parse_decimal_token(token: &str) -> Option<f64> {
 }
 
 fn parse_duration(tokens: &[&str]) -> Option<(u64, usize)> {
-    for (idx, token) in tokens.iter().enumerate() {
-        let Some(amount) = parse_number(token) else {
+    let mut start = 0;
+    while start < tokens.len() {
+        let Some((amount, unit_index)) = parse_spoken_number(tokens, start) else {
+            start += 1;
             continue;
         };
-        // Combine a spoken compound cardinal like "forty five" (45) into a
-        // single amount. Worded numbers arrive as separate whitespace tokens,
-        // so the tens word and the ones word must be stitched here.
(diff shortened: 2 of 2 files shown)


matedev01:
**LGTM!** 

Great fix for the silent failures in the quick-router. Moving away from a brittle, hardcoded lookup to a proper English cardinal parser is a massive improvement for robustness. 

Catching the missing teens (13-19) and adding support for multipliers like `hundred` and `thousand` ensures we actually parse natural speech correctly instead of just dropping the command. Excellent work!

